PR #10, opened 2023-06-07
Snap scrollable to bottom on message sent / received

### LINKED ISSUE #3: Scroll to bottom when new messages arrives
Currently the scrollable isn't scrolling to bottom when new messages arrives.  
This means its hard to know if anything happened because the scrollable isn't moving.

tarkah:
Resolves #3

I don't think we can effectively manage this with a custom widget since it'll have to be aware of what "new messages" are and we'll eventually need more complex logic, like if the user manually scrolls up, we don't auto-snap down but show an indicator they can click.

Therefore I think it's better to manage the lifecycle of this within the app and simply return a `snap_to` operation when we send / receive new messages.

I've added a new `message::Source` that is checked when receiving messages. Based on this, it'll query the correct underlying buffer for it's scrollable id to return in a `snap_to` operation.

Within the buffer, I've also added `snap_to` when sending messages. `update` plumbing has been updated to support these.


casperstorm (approved):
I think this is the way, really good pluming as well.

PR #12, opened 2023-06-07
Fix/multi server

tarkah:
Fixes a couple bugs I noticed when testing a multiple server config:

- Server / channel names would jump around due to hashmap iteration not being order stable. BTreeMap fixes this.
- Stream bug that mapped all incoming messages against the first server setup.

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -1,4 +1,5 @@
-use std::{collections::HashMap, fmt};
+use std::collections::BTreeMap;
+use std::fmt;
 
 use irc::client::Client;
 use irc::proto::Command;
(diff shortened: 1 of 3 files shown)


data/src/client.rs
```diff
@@ -60,7 +61,7 @@ impl Connection {
 }
 
 #[derive(Debug, Default)]
-pub struct Map(HashMap<Server, State>);
+pub struct Map(BTreeMap<Server, State>);
```
casperstorm, line 64: Nice spotted, this make sense.

casperstorm (approved):

PR #14, opened 2023-06-07
Unfocus on single ESC press

tarkah:
This fixes the issue where it takes 2 presses of ESC to unfocus a buffer since text input captures the first ESC press to unfocus itself. It feels better to unfocus the buffer on a single press (for now).


casperstorm:
Good catch, and i also think this is the best behavior for now.

casperstorm (approved):

Issue #15, opened 2023-06-08
Colored nicknames


casperstorm:
It's common that each nickname has it's own color, which helps readability.
For picking the colors we should use `accent` color converted to HSL and then pick a random hue value, while not touching saturation and lightning.

Stretch goals:
 - Caching the color to a file with `nick@host` and invalidated on theme change.
 - Enable/disable colored nicks in channel settings (#6)

tarkah:
I think we can implement without any caching. If we hash nick@host, we can use that as a seed for rng for the hue value and it should be deterministic. Maybe we cache at runtime for performance reasons, but we shouldn't need to persist it to fs

casperstorm:
Thats a great idea!

PR #18, opened 2023-06-09
Random user colors based on fixed seed

casperstorm:
Fixes https://github.com/squidowl/halloy/issues/15.

For now i just added a toggle in title header, but this will be removed once we get https://github.com/squidowl/halloy/issues/6.

### DIFF
--- a/data/src/message.rs
+++ b/data/src/message.rs
@@ -1,5 +1,7 @@
 use irc::proto::{Command, Response};
 
+use crate::User;
+
 pub enum Source {
     Server,
     Channel(String),
@@ -47,13 +49,13 @@ impl Message {
         }
     }
 
-    pub fn nickname(&self) -> Option<&str> {
+    pub fn user(&self) -> Option<User> {
         match &self {
-            Message::Sent { nickname, .. } => Some(nickname.as_str()),
-            Message::Received(message) => message.prefix.as_ref().and_then(|prefix| match prefix {
-                irc::proto::Prefix::ServerName(_) => None,
-                irc::proto::Prefix::Nickname(nickname, _, _) => Some(nickname.as_str()),
-            }),
+            Message::Sent { nickname, .. } => Some(User::new(nickname)),
+            Message::Received(message) => message
+                .prefix
+                .as_ref()
+                .and_then(|prefix| User::try_from(prefix).ok()),
         }
     }
 }
--- a/data/src/user.rs
+++ b/data/src/user.rs
@@ -2,12 +2,41 @@
 pub struct User(irc::client::data::User);
 
 impl User {
+    pub fn new(nickname: &str) -> Self {
+        User(irc::client::data::User::new(nickname))
+    }
+
+    pub fn color_seed(&self) -> &str {
+        self.hostname().unwrap_or_else(|| self.nickname())
+    }
+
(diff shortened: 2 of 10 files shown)


data/src/message.rs
```diff
@@ -47,13 +49,10 @@ impl Message {
...
 
-    pub fn nickname(&self) -> Option<&str> {
+    pub fn user(&self) -> Option<User> {
         match &self {
-            Message::Sent { nickname, .. } => Some(nickname.as_str()),
-            Message::Received(message) => message.prefix.as_ref().and_then(|prefix| match prefix {
-                irc::proto::Prefix::ServerName(_) => None,
-                irc::proto::Prefix::Nickname(nickname, _, _) => Some(nickname.as_str()),
-            }),
+            Message::Sent { nickname, .. } => Some(User::new(nickname)),
```
casperstorm, line 54: I think we, down the line, need to do a `WHOIS` on ourself because we don't know much about yourself here. And in my case where im connected through ZNC my nickname is actually different from the config:

Screenshot 2023-06-09 at 14 32 43

data/src/user.rs
```diff
@@ -2,15 +2,44 @@
...
+        match prefix {
+            Prefix::ServerName(_) => Err(()),
+            Prefix::Nickname(nickname, username, host) => {
+                let user = format!("{}!{}@{}", nickname, username, host);
+                Ok(User(data::User::new(&user)))
+            }
+        }
+    }
+}
+
```
tarkah, line 42: Nice!

src/buffer/channel.rs
```diff
@@ -24,13 +24,19 @@ pub fn view<'a>(
         .get_channel_messages(&state.server, &state.channel)
         .into_iter()
         .filter_map(|message| {
-            let nickname = message.nickname().unwrap_or_default();
+            let Some(user) = message.user() else {
+                return None
+            };
```
tarkah, line 29: You can just `message.user()?` throw since we are returning an Option

src/buffer/channel.rs
```diff
@@ -24,13 +24,19 @@ pub fn view<'a>(
...
             let message = message.text()?;
 
             Some(
                 container(
                     row![
-                        text(format!("<{nickname}>")).style(theme::Text::Accent),
+                        text(format!("<{}>", user.nickname())).style(theme::Text::Nickname(
+                            state
+                                .unique_user_colors
+                                .then_some(user.color_seed().to_string())
```
tarkah, line 38: Use `then(|| )` so we don't call `color_seed` eagerly even when `false`

src/screen/dashboard.rs
```diff
@@ -250,6 +251,17 @@ impl Dashboard {
                     self.panes.maximize(&pane);
                 }
             }
+            Message::UniqueUserColors => {
+                if let Some(pane) = self.focus {
+                    if let Some(pane) = self.panes.get_mut(&pane) {
```
tarkah, line 256: Maybe we can add a helper here: `get_focused_mut`?

tarkah (approved):
LGTM! Makes a big difference. I've left a few nits

src/buffer/channel.rs
casperstorm: Ah doh, of cause.

PR #21, opened 2023-06-10
Save to settings

casperstorm:
This adds a `config::Channel` struct where we can add different settings which is channel specific. Editing these values will save the config so its persisted.

I moved `user_color` to be a global setting, and removed it from module header. 
Does not make sense to set it on a per buffer basis imo.

### DIFF
--- a/data/src/channel.rs
+++ b/data/src/channel.rs
@@ -7,8 +7,8 @@ pub struct Config {
 
 #[derive(Debug, Clone, Copy, Default, Deserialize, Serialize)]
 pub enum Position {
-    #[default]
     Left,
+    #[default]
     Right,
 }
 
@@ -23,7 +23,7 @@ impl Default for Users {
     fn default() -> Self {
         Self {
             visible: true,
-            position: Position::Left,
+            position: Position::default(),
         }
     }
 }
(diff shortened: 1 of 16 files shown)


data/src/channel.rs
```diff
@@ -0,0 +1,6 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Default, Deserialize, Serialize)]
+pub struct Config {
+    pub show_user_list: bool,
```
casperstorm, line 5: Do we have a better name for this? 🤔

data/src/config.rs
```diff
@@ -7,13 +7,46 @@ use serde::{Deserialize, Serialize};
...
 
 #[derive(Debug, Clone, Default, Deserialize, Serialize)]
 pub struct Config {
     #[serde(default)]
     pub palette: Palette,
     pub servers: BTreeMap<String, server::Config>,
+    #[serde(default)]
+    pub channels: BTreeMap<String, BTreeMap<String, channel::Config>>,
+    #[serde(default)]
+    pub user_color: UserColor,
```
casperstorm, line 20: Suggestion for a better name?

src/buffer/channel.rs
```diff
@@ -64,7 +63,7 @@ pub fn view<'a>(
     });
 
     // TODO: Maybe we should show it to the right instead of left.
```
casperstorm, line 65: Not using position in `User` position user sidebar. Will add that later. We have the information now, which was the goal.

tarkah (approved):
LGTM!

PR #22, opened 2023-06-11
A few different UI tweaks

casperstorm:
1. Updated UI on side menu to make it a little easier to distinguish servers from channels.
    1. Sorted channels and servers alphabetical 
2. Added ability to change side on channel users. Also changed default to right to give the UI a better weight ([menu, channel, user]). Changing side is done through config, don't want to add clutter for that right now.
3. Updated styling on commands. Hardcoded a small y-offset which could turn into a config later. 

Screenshot 2023-06-11 at 23 36 15


tarkah:
Looks good!

tarkah (approved):

PR #27, opened 2023-06-13
Feat/optimizations

tarkah:
Optimizes two different things

Adds message batching to reduce number of Messages app has to `update`

Refactors the buffer scrollable to a new `scroll_view` module which now handles the following:

- A new Limit to limit how many messages we render
- Scrollable alignment settings to align to bottom by default, but then handle switching alignment for desired scrolling behavior when "lazy loading" older messages
- Bumping limit / displaying more messages when top of scrollable is hit
- If user scrolls away from bottom, the viewport will stay where it's at when new messages come in so the user isn't pulled away from the messages they were looking at
- Scroll to bottom when user submits a new message.

You can test this scroll_view behavior by issuing a `/list` command and scrolling up while looking at scrollabar, you'll notice the new messages "loading" and displaying seamlessly

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1017,7 +1017,7 @@ dependencies = [
 [[package]]
 name = "iced"
 version = "0.9.0"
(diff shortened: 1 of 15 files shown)


casperstorm:
LGTM,works super nice.

casperstorm (approved):

PR #29, opened 2023-06-14
Add history

tarkah:
This adds a new `history` module for loading / saving message history.

`history::manager` is where all messages are loaded, added to & flushed to disk.

- Any open pane is "tracked" meaning we load it's full history from disk. When flushed, we overwrite the entire file since we have all message.
- Messages for servers / channels / users not in an open pane are buffered and flushed to disk (appended) to reduce memory consumption. They get loaded in full when added to a buffer.


casperstorm (approved):
This is really an amazing addition

PR #30, opened 2023-06-16
feat/timestamp

casperstorm:
this is the pr for adding initial timestamps
@tarkah could you take a look at this when you have time, not entirely sure why we don't always get timestamps along the tags.


data/src/message.rs
```diff
@@ -52,6 +52,18 @@ impl Message {
...
 
+        // Check to see if we have timestamp on a tag.
+        // TODO: not sure why it isn't always working.
+        let timestamp = proto
+            .tags
+            .as_ref()
+            .and_then(|tags| tags.iter().find(|tag| tag.0 == "time"))
+            .and_then(|tag| tag.1.clone())
+            .and_then(time::Posix::parse_rfc3339)
+            .unwrap_or_else(time::Posix::now);
```
casperstorm, line 63: i would assume we get timestamp on all messages after enabling the capability but that seems not to be the case.

data/src/time.rs
```diff
@@ -14,4 +14,13 @@ impl Posix {
 
         Self(nanos_since_epoch)
     }
+
+    pub fn parse_rfc3339(date: impl AsRef<str>) -> Option<Self> {
+        let datetime = chrono::DateTime::parse_from_rfc3339(date.as_ref()).ok()?;
```
tarkah, line 19: I say we update Message to hold a DateTime<Utc> instead of posix. Since we will most likely need to format this into human readable anyways, it makes sense.

We can always make the serde implementation go into / from u64 timestamp for compactness.

src/buffer/channel.rs
```diff
@@ -35,11 +35,17 @@ pub fn view<'a>(
             |message| match &message.source {
                 data::message::Source::Channel(_, kind) => match kind {
                     data::message::ChannelSender::User(user) => {
+                        let datetime = selectable_text(
+                            &message
+                                .timestamp
+                                .with_timezone(&chrono::Local)
+                                .format("[%H:%M] "),
```
casperstorm, line 42: should we make a helper method so we save `DateTime<Utc>` but return `DateTime<Local>`?

data/src/message.rs
```diff
@@ -77,7 +77,7 @@ impl fmt::Display for Content {
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Message {
-    pub timestamp: time::Posix,
+    pub timestamp: DateTime<Utc>,
```
casperstorm, line 80: Rename this to `datetime`?

data/src/config.rs
```diff
@@ -17,26 +17,80 @@ pub struct Config {
...
-    }
-}
+// impl NicknameColor {
+//     pub fn unique_colors(&self) -> bool {
+//         match self {
+//             Nickname::Solid => false,
+//             Nickname::Unique => true,
+//         }
+//     }
+// }
```
tarkah, line 93: Don't forget to delete

data/src/message.rs
```diff
@@ -88,6 +88,10 @@ impl Message {
         matches!(self.source, Source::Server)
     }
 
+    pub fn datetime(&self, fmt: &str) -> String {
+        self.datetime.with_timezone(&Local).format(fmt).to_string()
+    }
+
```
tarkah, line 94: We should either un-pub the field or name this differently from the field since one returns a DateTime and this one a string, but they're named the same.

data/src/message.rs
```diff
@@ -226,6 +232,17 @@ fn source(message: irc::proto::Message, our_nick: &Nick) -> Option<Source> {
...
 
+fn datetime(message: &irc::proto::Message) -> DateTime<Utc> {
+    message
+        .tags
+        .as_ref()
+        .and_then(|tags| tags.iter().find(|tag| tag.0 == "time"))
+        .and_then(|tag| tag.1.clone())
+        .and_then(|rfc3339| DateTime::parse_from_rfc3339(&rfc3339).ok())
+        .map(|dt| dt.with_timezone(&Utc))
+        .unwrap_or(Utc::now())
```
tarkah, line 243:
```suggestion
        .unwrap_or_else(Utc::now)
```

tarkah (approved):
Looks great!! Just a few nits.

PR #32, opened 2023-06-19
Feat/context menu

tarkah:
Adds a context menu widget and implements a basic menu on the side menu.

There is a known bug where swapping panes can cause one to disappear. Need to investigate further.

@casperstorm Please update styling / add new theme variants to style context menu.

https://github.com/squidowl/halloy/assets/10239377/26b3838e-ae3c-4dc2-a757-2ec89718c406


casperstorm:
this is super sleek

PR #33, opened 2023-06-20
better messages

casperstorm:
this PR adds some better filtering of where messages goes.

Screenshot 2023-06-20 at 21 45 18

### DIFF
--- a/data/src/message.rs
+++ b/data/src/message.rs
@@ -10,23 +10,23 @@ pub type Channel = String;
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
 pub enum Source {
     Server,
-    Channel(Channel, Kind),
+    Channel(Channel, ChannelSender),
     Query(User),
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
-pub enum Kind {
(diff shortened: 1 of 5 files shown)


data/src/message.rs
```diff
@@ -5,14 +5,32 @@ use serde::{Deserialize, Serialize};
...
+
+impl Kind {
+    pub fn user(&self) -> Option<&User> {
+        match self {
+            Kind::User(user) => Some(user),
+            Kind::Server => None,
+        }
+    }
+}
+
```
tarkah, line 33: `Kind` is a bit generic since it only applies to `Channel`. Maybe we call it `ChannelSender`?

data/src/message.rs
```diff
@@ -51,33 +69,92 @@ impl Message {
...
-                    if target.is_channel_name() {
-                        Source::Channel(target, user)
-                    } else {
-                        Source::Query(user)
+                match (target.is_channel_name(), &user) {
+                    (true, Some(user)) => Source::Channel(target.clone(), Kind::User(user.clone())),
+                    (false, Some(user)) => Source::Query(user.clone()),
+                    _ => {
+                        return None;
                     }
```
tarkah, line 95: Nit: Remove reference from &user and no need for `user.clone()`

tarkah (approved):
LGTM! Just a couple minor nits

PR #34, opened 2023-06-20
Fix query messages

tarkah:
Fixes query messages in the following ways:

- Only ties nick to the query source 
- Adds a user field so we know who sent the message in the query buffer (us vs them)
- Ensures the privmsg target matches our nick
- Update our nickname when REPL_WELCOME or NICK messages come over the wire

I've also hooked into `send_command` to dispatch the `send_user_message` when `/privmsg` is used and return a message we can record to history. This will ensure the buffer is viewable from the sidebar when initially sending a query message to someone.

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -3,7 +3,9 @@ use std::fmt;
 
 use irc::client::Client;
 use irc::proto;
+use irc::proto::ChannelExt;
 
+use crate::user::Nick;
 use crate::{message, time, Command, Message, Server, User};
 
 #[derive(Debug)]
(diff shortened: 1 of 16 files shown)


data/src/client.rs
```diff
@@ -96,18 +101,28 @@ impl Connection {
...
+
+    pub fn handle_message(&mut self, message: &irc::proto::Message) {
+        use irc::proto::{Command, Response};
+
+        match &message.command {
+            Command::NICK(nick) => self.resolved_nick = Some(nick.to_string()),
+            Command::Response(response, args) => match response {
+                Response::RPL_WELCOME => {
+                    if let Some(nick) = args.first() {
+                        self.resolved_nick = Some(nick.to_string());
```
casperstorm, line 120: Good stuff.

data/src/history/manager.rs
```diff
@@ -133,11 +134,19 @@ impl Manager {
     pub fn add_raw_messages(
         &mut self,
         messages: Vec<(Server, irc::proto::Message)>,
+        clients: &mut client::Map,
     ) -> HashSet<(Server, message::Source)> {
         messages
             .into_iter()
             .filter_map(|(server, message)| {
-                let message = crate::Message::received(message)?;
+                log::debug!("Message received => {message:?}");
```
casperstorm: Might wanna remove this.

casperstorm (approved):

casperstorm:
super smooth query messages now

PR #37, opened 2023-06-20
Parse & display topic message

### LINKED ISSUE #8: Channel topic
We need to fetch channel topic initially when the channel buffer is created: https://docs.rs/irc/0.15.0/irc/client/struct.Client.html#method.send_topic.

And we also need to update this topic when a TOPIC message is being sent.

Topic should be displayed in the pane title.

tarkah:
Resolves #8 

@casperstorm can you look at topic formatting? Right now it just adds it to titlebar.

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -43,7 +43,7 @@ impl Connection {
             timestamp: time::Posix::now(),
             direction: message::Direction::Sent,
             source: message::Source::Channel(channel, message::ChannelSender::User(user)),
-            text,
+            content: message::Content::Text(text.to_string()),
         }
     }
 
@@ -62,7 +62,7 @@ impl Connection {
             timestamp: time::Posix::now(),
             direction: message::Direction::Sent,
             source: message::Source::Query(nick.clone(), User::new(self.nickname(), None, None)),
-            text,
+            content: message::Content::Text(text.to_string()),
         }
     }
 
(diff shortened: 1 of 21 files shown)


data/src/client.rs
```diff
@@ -43,7 +43,7 @@ impl Connection {
             timestamp: time::Posix::now(),
             direction: message::Direction::Sent,
             source: message::Source::Channel(channel, message::ChannelSender::User(user)),
-            text,
+            content: message::Content::Text(text.to_string()),
```
casperstorm, line 46: nit: no need to `to_string()` here as we do it on line 32 as well.

data/src/client.rs
```diff
@@ -62,7 +62,7 @@ impl Connection {
             timestamp: time::Posix::now(),
             direction: message::Direction::Sent,
             source: message::Source::Query(nick.clone(), User::new(self.nickname(), None, None)),
-            text,
+            content: message::Content::Text(text.to_string()),
```
casperstorm, line 65: nit: no need to to_string() here as we do it on line 51 as well.

casperstorm (approved):
Awesome - i like this approach.
One issue i found we could address now or wait is capping the width of the topic - the issue is if the length is greater than the width the entire `text()` gets hidden when focused. Again, i dont see this as a blocker right now just wanted to highlight it.

topic

tarkah:
As discussed, I've reverted showing topic anywhere in the UI since we already display the messages in the buffer. There's no great way to display what can be a large amount of topic text in the titlebar. We can revisit this in the future.

PR #39, opened 2023-06-22
Add support for /me

tarkah:
Adds support for issuing `/me <action>` commands and receiving them as well.

This is a "synthetic" type command where it just encode the action into PRIVMSG and we need to decode from that.

image

@casperstorm do we maybe want to use a different color for this from normal server message? If so, I'd add a variant to `Sender::Action` and you can use that for theming different

### DIFF
--- a/data/src/buffer.rs
+++ b/data/src/buffer.rs
@@ -7,3 +7,13 @@ pub enum Buffer {
     Channel(Server, String),
     Query(Server, Nick),
 }
+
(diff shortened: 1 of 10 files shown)


casperstorm:
Awesome!!
I just updated with `Sender::Action` and styled it differently than `Sender::Server`.

casperstorm (approved):

tarkah:
Looks great!

PR #41, opened 2023-06-22
Feat/input refactor

tarkah:
This refactors how we handle receiving encoding messages, recording parsed messags into history, parsing user input into both encoded messages and parsed messages that get recorded into history.

There is now a much better separation between everything. `client / connection` isn't aware of history. `history` isn't aware of `client / connection`. `client / connection` is responsible over sending encoded messages. `history` is responsible over storing / recording parsed messages.

`input` now parses into the new `Input` on send. This struct defines how we go from input to both encoded messages to send via `client / connection` and parsed messages that get recorded in `history`.

A new `input_view` module is added to de-duplicate the input flow of send -> input -> send encoded + record parsed message from each buffer.

### DIFF
--- a/data/src/input.rs
+++ b/data/src/input.rs
@@ -1,0 +1,89 @@
+use chrono::Utc;
+use irc::proto::ChannelExt;
+
+use crate::user::Nick;
+use crate::{command, message, Buffer, Command, Message, Server, User};
+
+pub fn parse(buffer: Buffer, input: &str) -> Result<Input, command::Error> {
+    let content = match command::parse(input, &buffer) {
+        Ok(command) => Content::Command(command),
+        Err(command::Error::MissingSlash) => Content::Text(input.to_string()),
+        Err(error) => return Err(error),
+    };
+
+    Ok(Input { buffer, content })
+}
+
+#[derive(Debug, Clone)]
+pub struct Input {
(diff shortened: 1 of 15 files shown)


src/buffer/channel.rs
```diff
@@ -5,15 +5,14 @@ use data::server::Server;
...
 use crate::theme;
-use crate::widget::{input, selectable_text, Collection, Column, Element};
+use crate::widget::{selectable_text, Collection, Column, Element};
 
 #[derive(Debug, Clone)]
 pub enum Message {
-    Send(input::Content),
-    CompletionSelected,
     ScrollView(scroll_view::Message),
+    InputView(input_view::Message),
```
casperstorm, line 15: Based

casperstorm (approved):

data/src/input.rs
```diff
@@ -0,0 +1,89 @@
...
+
+        match command {
+            Command::Msg(target, text) => Some(Message {
+                datetime: Utc::now(),
+                direction: message::Direction::Sent,
+                source: source(
+                    target,
+                    message::Sender::User(User::new(our_nick.clone(), None, None)),
+                )?,
+                text: text.clone(),
```
casperstorm, line 49: No need for `.clone()` here :)

PR #42, opened 2023-06-22
ability to build macos

casperstorm:
This PR makes it possible publish an release to drafts.
Currently only supporting macOS.

```
sh scripts/build-macos.sh dmg
```


Makefile
```diff
@@ -0,0 +1,65 @@
+TARGET = halloy
```
tarkah, line 1: Do we plan on having linux commands in the Makefile ?

casperstorm: I think it make sense to have it all there - think Mac should be the most complicated because .App and .Dmg?

tarkah: Do you plan on using the makefile for anything outside of CI? I would never use it normally for just building / testing the app. Maybe we just make this a script `build-macos` and throw it in `scripts` folder?

tarkah: Just feel like if ppl see `Makefile` in the root of our repo, they'll expect it to work on Linux too and I don't really want to use `Makefile` for that

casperstorm:
@tarkah take a look at the updated script in `assets`.

tarkah:
> @tarkah take a look at the updated script in `assets`.

Looks great!

tarkah (approved):

PR #43, opened 2023-06-22
Rework config

tarkah:
This improves the config by making it read only and refactors all the options into `Settings` related structures.

The config now defines a `new_pane` option which are the pane settings that will be applied to new panes by default. Toggling the user list will update the settings for that single pane for the duration of the program.

When we eventually add layout persistence, these pane settings will get persisted.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -10,7 +10,7 @@ mod screen;
 mod theme;
 mod widget;
 
-use data::config::Config;
+use data::config::{self, Config};
 use data::stream;
 use iced::widget::container;
 use iced::{executor, Application, Command, Length, Subscription};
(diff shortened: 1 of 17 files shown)


src/main.rs
```diff
@@ -75,18 +75,8 @@ struct Halloy {
...
-        Config {
-            palette: self.config.palette,
-            servers: self.config.servers.clone(),
-            channels: self.config.channels.clone(),
-            buffer: self.config.buffer.clone(),
-            error: self.config.error.clone(),
-        }
-    }
+    // TODO: Make this a different screen?
+    load_config_error: Option<config::Error>,
```
casperstorm, line 79: Yes!

casperstorm (approved):
This is the right approach with read only config and then settings. I can make the error screen later to remove the config error we pass through the entire thing. Looks great, good clean up.

src/main.rs
tarkah: Thanks!

PR #45, opened 2023-06-23
Flatpak support

tarkah:
This adds support for flatpak by making the following changes:


scripts/flatpak.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -xe
```
tarkah, line 2: I referenced this and the json manifest file from `wezterm`. Huge shoutout there, it was a big help!

src/logger.rs
```diff
@@ -22,6 +22,7 @@ pub fn setup(is_debug: bool) -> Result<(), log::Error> {
         })
         .level(log::LevelFilter::Off)
         .level_for("panic", log::LevelFilter::Error)
+        .level_for("iced_wgpu", log::LevelFilter::Info)
```
tarkah, line 25: Included so we can see if `wgpu` is initialized in the logs. I've confirmed flatpak + wgpu works fine! The tiny-skia override also works via ICED_BACKEND var

Cargo.toml
```diff
@@ -27,4 +27,4 @@ features = ["v4"]
 members = ["data"]
 
 [patch.crates-io]
-iced = { git = "https://github.com/tarkah/iced", rev = "bcd8fda9a6e9c5fb1e1c6d1fa38d807c931bfebc" }
+iced = { git = "https://github.com/tarkah/iced", rev = "c6fa0d185c8f44425445d3c4d85406a1967287d7" }
```
tarkah, line 30: Forked until we can get upstream fixes merged

casperstorm:
@tarkah take a look at #42 - i've included icon there in `assets/`.

PR #56, opened 2023-06-26
Add persistent layout (dashboard)

### LINKED ISSUE #48: Dashboard persistence
We should persist user dashboard so it's faster for the user to get going after launching Halloy.
If they have lost acces (eg. been banned) from a channel and we can't load it we could just show a empty buffer instead, but keeping the layout as it was.

tarkah:
Resolves #48

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -35,44 +35,42 @@ servers:
     channels:
       - "##rust"
       
-# Settings applied by default to new panes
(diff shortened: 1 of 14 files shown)


casperstorm (approved):

casperstorm:
Clean AF code <3

PR #57, opened 2023-06-26
add windows icon

casperstorm:
The good thing; its working.
The bad thing; its a little too small:

image

Will see if i can improve this.


tarkah (approved):
LGTM! We can improve in the future w/ MSI install for an easier upgrade path

PR #58, opened 2023-06-26
Allow custom font

### LINKED ISSUE #26: Support for Custom Fonts in the App
As a user, I want the ability to use custom fonts in the app. I should be able to place font files in the config directory and specify them in the config file. The application should automatically detect new font files and utilize them instead of the default fonts.

Acceptance Criteria:

1. Have a designated font directory where font files can be placed.
2. The config file should have an option to specify the custom font files to be used.
3. Upon launching the app, it should detect and load the custom font files specified in the config file.
4. The app should handle any errors or missing font files gracefully, falling back to the default fonts if necessary.

tarkah:
Resolves #26

### DIFF
--- a/data/src/config.rs
+++ b/data/src/config.rs
@@ -25,6 +25,7 @@ pub struct Config {
 #[derive(Debug, Clone, Default, Deserialize)]
 pub struct Font {
     pub family: Option<String>,
+    pub size: Option<u8>,
     // TODO: Do we make size, etc configurable and pass to Theme?
 }
 
--- a/data/src/config.rs
+++ b/data/src/config.rs
@@ -26,7 +26,6 @@ pub struct Config {
 pub struct Font {
     pub family: Option<String>,
     pub size: Option<u8>,
-    // TODO: Do we make size, etc configurable and pass to Theme?
 }
 
 impl Config {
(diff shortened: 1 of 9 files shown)


src/main.rs
```diff
@@ -47,7 +47,13 @@ pub fn main() -> iced::Result {
         data::environment::formatted_version()
     );
 
-    if let Err(error) = Halloy::run(settings()) {
+    let config_load = Config::load();
+
+    // DANGER ZONE - font must be set using config
+    // before we do any iced related stuff w/ it
+    font::set(config_load.as_ref().ok());
```
casperstorm, line 54: make sense!

data/src/config.rs
```diff
@@ -15,11 +15,20 @@ pub struct Config {
...
     /// Default settings when creating a new buffer
     #[serde(default)]
     pub new_buffer: buffer::Settings,
 }
 
+#[derive(Debug, Clone, Default, Deserialize)]
+pub struct Font {
+    pub family: Option<String>,
+    pub size: Option<u8>,
+    // TODO: Do we make size, etc configurable and pass to Theme?
```
casperstorm, line 29: What do you mean by this?

casperstorm (approved):
Tested on both Windows and macOS. Works really well.

data/src/config.rs
tarkah: Oops old comment

PR #59, opened 2023-06-27
Fix/scroll view

tarkah:
Fixes a few issues related to scrollview:

- Changing message time to DateTime removed nanosecond precision which makes using `Limit::Since` not reliable for filtering since a specific message.
- Macos trackpad w/ multiple on_scroll events between views caused some issues.

Scrolling extremely fast on trackpad either from top of scrollable down or bottom up should be very reliable now.

@casperstorm can you confirm you can't break scrollable from either top or bottom (home / end) then scroll? I usually issue a `/list` in server to ensure scrollable is huge and scroll like crazy.

### DIFF
--- a/data/src/environment.rs
+++ b/data/src/environment.rs
@@ -1,5 +1,5 @@
 use std::env;
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 
 pub const VERSION: &str = env!("VERSION");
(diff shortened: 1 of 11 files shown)


casperstorm (approved):
Code looks good, and I havent been able to break it yet.

PR #61, opened 2023-06-28
Nicklist context

### LINKED ISSUE #49: Simple context menu for user
Add a simple context menu for users in the user sidebar.
The context menu should have a few essential actions for a user.

We can always add more, but as a start i think the following is important:

* Whois
* Private message (Query)
* Give Op (+o) (or Take Op (-o))
* Give Voice (+v) (or Take Voice (-v))
* Ban
* Kick

casperstorm:
This is the first draft of context menu for nicklist.
Currently it supports two things:
- Whois
- Message (Query)

Closes #49

### DIFF
--- a/data/src/input.rs
+++ b/data/src/input.rs
@@ -22,6 +22,13 @@ pub struct Input {
 }
 
 impl Input {
+    pub fn command(buffer: Buffer, command: Command) -> Self {
+        Self {
+            buffer,
+            content: Content::Command(command),
+        }
+    }
+
     pub fn server(&self) -> &Server {
         self.buffer.server()
     }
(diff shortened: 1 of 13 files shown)


tarkah:
> [ ] Align the context menu better. (@tarkah can you help with this?)

Will do :+1:

data/src/input.rs
```diff
@@ -74,7 +74,7 @@ impl Input {
 }
 
 #[derive(Debug, Clone)]
-enum Content {
+pub enum Content {
     Text(String),
```
tarkah: I've locked this module back down 322ec2e and added an appropriate builder method. We don't need to expose these internals.

PR #62, opened 2023-06-29
Feat: unread

### LINKED ISSUE #40: Display unread message count
As a user, I want to see the number of unread messages in unfocused buffers. I also need a small tag next to the buffer name showing the unread count. Additionally, it would be great to have a horizontal line indicating the starting point of unread messages (stretch goal).

Acceptance Criteria:

* Display a small tag next to the buffer name, showing the current unread message count.
* Remove the unread count tag when when focusing the buffer
* (Stretch Goal) Add a horizontal line across the buffer to indicate the start of unread messages.

tarkah:
Resolves #40 

Adds the following:

- Sidebar icons are "filled" when there are "unread" messages. "unread" counts as any received message from a user or action (/me). Server messages don't trigger this. Likewise, the server entries in the sidebar don't show unread notifications. I feel this makes sense since we only care about new _user_ messages.
- Buffers show a horizontal rule between the last and new messages since the buffer was last opened.

image

### DIFF
--- a/src/buffer/scroll_view.rs
+++ b/src/buffer/scroll_view.rs
@@ -65,13 +65,15 @@ pub fn view<'a>(
         .filter_map(format)
         .collect::<Vec<_>>();
 
-    let content = if new.is_empty() && old.is_empty() {
-        column![]
-    } else {
+    let show_divider = !new.is_empty() || matches!(status, Status::Idle(Anchor::Bottom));
+
+    let content = if show_divider {
         let divider = container(horizontal_rule(1).style(theme::Rule::Unread))
             .width(Length::Fill)
             .padding(5);
         column![column(old), divider, column(new)]
+    } else {
+        column![column(old), column(new)]
     };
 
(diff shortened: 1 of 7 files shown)


src/buffer/scroll_view.rs
```diff
@@ -31,42 +32,66 @@ pub fn view<'a>(
...
-        status,
-        viewport,
-    })
-    .id(state.scrollable.clone())
-    .into()
+            .padding(5);
+        column![column(old), divider, column(new)]
+    } else {
+        column![column(old), column(new)]
+    };
```
casperstorm, line 77: Clean way of doing this, nice!

casperstorm (approved):
This turned out awesome. I might wanna play with other "unread" indicators, but for now this is great.

tarkah:
> I took action and changed our icons to only a dot if theres any unread messages. I feel like it became to busy before.

Agreed! This is much nicer :D

PR #74, opened 2023-06-30
Allow configuring sidebar default pane action (new pane vs replace)

neilalexander:
This is somewhat of an attempt to modify the default click behaviour of channels in the sidebar. Today, each channel clicked opens a new pane, which quickly clutters up when dealing with lots of channels. Ideally I'd like to be able to just replace the focused pane instead, allowing me to right-click to open new panes as needed, but to not clutter up the window otherwise.

I figured just changing the default behaviour without configurability is probably a bad thing, so in lieu of that, I've tried to add a new configuration option that lets you choose either `NewPane` (today's behaviour) or `ReplacePane` (my proposed behaviour):
```
dashboard:
  sidebar_default_action: ReplacePane
```

The fundamental new-vs-replace pane part works but I am struggling with the config side of things. I am not really a native Rustacean (please go easy on me) and I am not *really* sure of the best place to put this config item or how to pass it down the hierarchy.

It appears that the `Dashboard` state gets written out to a file, so that means that any config there also gets written into the state file and therefore later changes to the option in `config.yaml` are ignored. I'd appreciate any pointers on what the best approach here should be or if it should go somewhere else.

### DIFF
--- a/src/screen/dashboard.rs
+++ b/src/screen/dashboard.rs
@@ -4,7 +4,7 @@ pub mod side_menu;
 use std::time::{Duration, Instant};
 
 use data::history::manager::Broadcast;
-use data::{history, Config, Server};
+use data::{dashboard, history, Config, Server};
 use iced::widget::pane_grid::{self, PaneGrid};
 use iced::widget::{container, row};
 use iced::{clipboard, window, Command, Length, Subscription};
@@ -330,7 +330,11 @@ impl Dashboard {
         Command::none()
     }
 
-    pub fn view<'a>(&'a self, clients: &'a data::client::Map) -> Element<'a, Message> {
+    pub fn view<'a>(
+        &'a self,
+        clients: &'a data::client::Map,
+        config: dashboard::Config,
+    ) -> Element<'a, Message> {
         let focus = self.focus;
 
         let pane_grid: Element<_> = PaneGrid::new(&self.panes, |id, pane, maximized| {
@@ -351,7 +355,13 @@ impl Dashboard {
 
         let side_menu = self
             .side_menu
-            .view(clients, &self.history, &self.panes, self.focus)
+            .view(
+                clients,
+                &self.history,
+                &self.panes,
+                self.focus,
+                config.sidebar_default_action,
+            )
             .map(Message::SideMenu);
 
         // The height margin varies across different operating systems due to design differences.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Unreleased
 
+Added:
+
+- New configuration option `dashboard.sidebar_default_action` allows controlling the pane behaviour when selecting channels in the sidebar
+
 Changed:
+
 - Default channel in `config.yaml` has been changed to `#halloy` (from `##rust`)
 
 # 2023.1-alpha1 (2023-06-30)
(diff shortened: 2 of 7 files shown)


tarkah:
Thanks for the contribution! This looks great and makes sense. 

Regarding your question around config vs settings. Our Settings structs are the runtime settings that can be affected by user interactions and are persisted between sessions via persisting that dashboard state to disk.

Since this is a purely config option, we can keep it isolated to `config` and only worry about loading it and passing it down during `view`.

I'm on vacation rn but I'll try to add a few comments a little later. I think it's really close!

tarkah:
Thanks for the contribution! This looks great and makes sense. 

Regarding your question around config vs settings. Our Settings structs are the runtime settings that can be affected by user interactions and are persisted between sessions via persisting that dashboard state to disk.

Since this is a purely config option, we can keep it isolated to `config` and only worry about loading it and passing it down during `view`.

You bring up a great point though and we don't have proper separation between our settings and config. We should probably separate these and have proper From implementations and convert them when dashboard / panes are created.

I'm on vacation rn but I'll try to add a few comments a little later. I think it's really close!

neilalexander:
Thanks for your feedback! I've passed down the config into the dashboard after the state restore which seems to avoid the problem of the state capturing the first config and then never allowing it to change. I've also then propagated to the sidebar via `view` as suggested.

I've also renamed the option to `dashboard.sidebar_default_action` as that seems more self-explanatory, although I'm happy to take advice on if you can think of a better name.

Local testing seems to be good, please let me know if there's anything else. 😄

src/screen/dashboard.rs
```diff
@@ -17,6 +17,7 @@ use crate::widget::{selectable_text, Element};
 const SAVE_AFTER: Duration = Duration::from_secs(3);
 
 pub struct Dashboard {
+    config: dashboard::Config,
```
tarkah, line 20: Since we don't plan to ever modify config at runtime, it can be a presentation only item. We can just pass it down during `view` from our root state and reference it when building the view tree.

tarkah (commented):
Change looks good! Just one minor nit, we can avoid duplicating Config data across state since we can just pass it down during `view`.

src/screen/dashboard.rs
neilalexander: I see what you mean, how does it look now?

tarkah (approved):
Looks great! Thanks for this contribution :D I'll follow up next week with a refactor to remove all `Settings` from config.

@casperstorm can you just do a quick test of this before we merge? I'm away from a computer for the next few days.

neilalexander:
Happy to contribute, I'm very much enjoying Halloy. Keep up the great work!

One final thing, do you prefer things to be rebased down into a single commit or shall I leave the history as-is?

tarkah:
> One final thing, do you prefer things to be rebased down into a single commit or shall I leave the history as-is?

Commit history is fine :) The only preference I have is preferring rebasing a feature branch to master vs back-merging when resolving conflicts. But this isn't a hard rule or anything.

casperstorm:
This is great 👍🏻 
Before I merge, please fill out the [CHANGELOG](https://github.com/squidowl/halloy/blob/main/CHANGELOG.md) with your change, and also update `config.yaml` with your new configuration.

neilalexander:
I've rebased on top of your most recent changes and updated the changelog/sample config.

Issue #68, opened 2023-06-30
Sorted user list in channel


lodenrogue:
It would be nice to have the user list sorted in each channel

lodenrogue:
I will submit a PR for this shortly

PR #80, opened 2023-07-02
Added sorting channel nicks

lodenrogue:
This PR sorts the nicknames in the channel and keeps the highest level access nicknames sorted at the top. This addresses #68 

This is my first time working in Rust. Please let me know if I'm doing anything non Rustic(?) or if you'd like any changes to the code.

### DIFF
--- a/data/src/user.rs
+++ b/data/src/user.rs
@@ -23,17 +23,27 @@ impl Hash for User {
 
 impl Ord for User {
     fn cmp(&self, other: &Self) -> std::cmp::Ordering {
-        other
-            .highest_access_level()
-            .cmp(&self.highest_access_level())
+        let access_level_comparison = other.highest_access_level()
+            .cmp(&self.highest_access_level());
+
+        if access_level_comparison != std::cmp::Ordering::Equal {
+            return access_level_comparison;
+        }
+
+        other.nickname().cmp(&self.nickname())
     }
 }
 
 impl PartialOrd for User {
     fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
-        other
-            .highest_access_level()
-            .partial_cmp(&self.highest_access_level())
+        let access_level_comparison = other.highest_access_level()
+            .partial_cmp(&self.highest_access_level());
+
+        if access_level_comparison != Some(std::cmp::Ordering::Equal) {
+            return access_level_comparison;
+        }
+
+        other.nickname().partial_cmp(&self.nickname())
     }
 }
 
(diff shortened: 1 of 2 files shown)


casperstorm (commented):
> This is my first time working in Rust.

Congratulations on your first contributions in Rust!
The feature is working as expected.

Rebase with master, and add your change to CHANGELOG then I think we are good to merge this.

lodenrogue:
> > This is my first time working in Rust.
> 
> Congratulations on your first contributions in Rust! The feature is working as expected.
> 
> Rebase with master, and add your change to CHANGELOG then I think we are good to merge this.

Thank you so much!

Done

data/src/user.rs
```diff
@@ -23,17 +23,27 @@ impl Hash for User {
...
 impl Ord for User {
     fn cmp(&self, other: &Self) -> std::cmp::Ordering {
-        other
-            .highest_access_level()
-            .cmp(&self.highest_access_level())
+        let access_level_comparison = other.highest_access_level()
+            .cmp(&self.highest_access_level());
+
+        if access_level_comparison != std::cmp::Ordering::Equal {
+            return access_level_comparison;
```
tarkah, line 30: You can chain orderings so you don't need to early return. Checkout Ordering::then.

lodenrogue: I was able to change the code for cmp. But couldn't figure it out for partial_cmp. Like I said this is my first time working in Rust. Can you take a look and let me know if I made the change correctly?

Also, can you suggest how to change partial_cmp?

tarkah: Just call access_level.cmp as well then wrap the returned Ordering in Some

lodenrogue: I came up with this but I think it becomes way less ledgible than the original:

```
impl PartialOrd for User {
    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
        other.highest_access_level()
            .partial_cmp(&self.highest_access_level())
            .and_then(|access_cmp| {
                other.nickname().partial_cmp(&self.nickname()).map(|nickname_cmp| {
                    access_cmp.then(nickname_cmp)
                })
            })
            
        // Original
        // let access_level_comparison = other.highest_access_level()
        //     .partial_cmp(&self.highest_access_level());
        //
        // if access_level_comparison != Some(std::cmp::Ordering::Equal) {
        //     return access_level_comparison;
        // }
        //
        // other.nickname().partial_cmp(&self.nickname())
    }
}
```

Thoughts?

tarkah: Both AccessLevel and Nick are Eq so we can just use `cmp` instead of partial_cmp.

The body here can be identical to what we used in the Eq body, just wrapping the return value in Some

lodenrogue: Yes. Spoke with Casper and he clarified. I submitted the change.

tarkah (approved):

PR #88, opened 2023-07-02
added raw command

### LINKED ISSUE #86: add raw command
it would be nice to have a `/raw` command in halloy to send raw irc lines, since some daemons have custom aliases (such as /NS).

example: `/raw NS INFO nick`, `/raw HELP CMODE +`

casperstorm:
Fixes #86.

We now correctly handle, and show a RAW command.

Screenshot 2023-07-03 at 00 01 58

### DIFF
--- a/data/src/command.rs
+++ b/data/src/command.rs
@@ -17,6 +17,7 @@ pub enum Kind {
     Part,
     Topic,
     Kick,
+    Raw,
 }
 
 impl FromStr for Kind {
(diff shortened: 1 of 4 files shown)


lodenrogue (approved):

data/src/command.rs
```diff
@@ -104,6 +107,12 @@ pub fn parse(s: &str, buffer: &Buffer) -> Result<Command, Error> {
             Kind::Kick => validated::<2, 1, true>(args, |[channel, user], [comment]| {
                 Command::Kick(channel, user, comment)
             }),
+            Kind::Raw => validated::<1, 1, true>(args, |[cmd], [args]| {
+                let args = args
+                    .map(|args| args.split_whitespace().map(|s| s.to_string()).collect())
+                    .unwrap_or_default();
+                Command::Raw(cmd, args)
+            }),
```
tarkah, line 115: You can skip using the validated function entirely and just use the cmd / args we parse initially into it

Issue #65, opened 2023-06-30
Create directory for themes


casperstorm:
As a user it would be nice to have a directory where multiple themes could live. These themes can then be previewed and selected from within the app. This would work nice in conjunction with #24.

PR #92, opened 2023-07-03
Create directory for themes

jhff:
This change addresses issue #65 by:

* Adding themes folder to config directory
* Adding default theme that is created when it is not in the config directory yet
* Changing config file to take a theme filename instead of `palette`. If theme is not found, default `Palette `is used.
* Allowing users to add their own theme files in themes directory


lodenrogue (commented):
I'm trying to run the code for this change but it's not creating the theme files in the theme directory. 

Theme directory is created successfully though.

I'm on macOS.

lodenrogue:
> I'm trying to run the code for this change but it's not creating the theme files in the theme directory.
> 
> Theme directory is created successfully though.
> 
> I'm on macOS.

So it seems the issue is if you already have a config and have used the app before it doesn't go into the _ option in the main->load_from_state() function. Which means that Config::create_themes_dir() is never called.

If add it at the start of the load_from_state() function it adds the themes but it doesn't update the app UI until the next time you open the app.

jhff:
> > I'm trying to run the code for this change but it's not creating the theme files in the theme directory.
> > Theme directory is created successfully though.
> > I'm on macOS.
> 
> So it seems the issue is if you already have a config and have used the app before it doesn't go into this option in the main->load_from_state() function.
> 
> ```
>  _ => (
>     Screen::Welcome(screen::Welcome::new()),
>     Config::default(),
>     Command::none(),
> )
> ```
> 
> Which means that Config::create_themes_dir() is never called inside the Welcome::new() method.
> 
> If we add Config::create_themes_dir() at the start of the load_from_state() function it adds the theme files in the themes directory but it doesn't update the app UI until the next time you open the app.

You're right, if you already had a config file, themes files weren't being created.

Moved `Config::create_themes_dir()` to `main()` so it runs at startup. It should only do this the first time you open the app without themes. 

Main issue is that a config file with `palette` instead of `theme` param, will go into default theme. Any user who already used the app might not notice that palette has been deprecated, but I don't think it makes sense to add code specifically for that use case in such an early phase.

casperstorm:
Great PR @jhff.
Personally, I wouldn't want us to create all those themes though 🤔 
I would rather keep it simple and just make Halloy able to read from `/themes/*.yaml` - and then if: `theme: "foobar.yaml"` We use `/themes/foobar.yaml` - if `theme` is None, or invalid, we use the default theme.
Then, in the `config.template.yaml` we can instead point to our [Wiki](https://github.com/squidowl/halloy/wiki/Themes) to encourage people to download and add new themes.

I think my reason behind this is that, as a user i would rather build a collection of themes i want, rather than a bunch of themes i never want to use.

jhff:
> Great PR @jhff. Personally, I wouldn't want us to create all those themes though 🤔 I would rather keep it simple and just make Halloy able to read from `/themes/*.yaml` - and then if: `theme: "foobar.yaml"` We use `/themes/foobar.yaml` - if `theme` is None, or invalid, we use the default theme. Then, in the `config.template.yaml` we can instead point to our [Wiki](https://github.com/squidowl/halloy/wiki/Themes) to encourage people to download and add new themes.
> 
> I think my reason behind this is that, as a user i would rather build a collection of themes i want, rather than a bunch of themes i never want to use.

np, I'll rework this PR to handle that! It was just an initial approach with a set of hardcoded themes 🙂

jhff:
Removed all themes except the default one [ce6da49](https://github.com/squidowl/halloy/pull/92/commits/ce6da492b1f21d6de9168f3c49807b7bf975f189)

Users can add their own theme files. Also expect "name" field in theme files so that we can list them in the future in the app if needed with the proper name.

assets/themes/halloy.yaml
```diff
@@ -0,0 +1,11 @@
+name: "Halloy"
```
tarkah, line 1: Do we need `name` or can the filename just be the key?

src/main.rs
```diff
@@ -49,6 +49,9 @@ pub fn main() -> iced::Result {
         data::environment::formatted_version()
     );
 
+    // Create themes directory
+    Config::create_themes_dir();
+
```
tarkah, line 54: Nit: A module namespace works better here than struct namespace... `config::create...`

assets/themes/halloy.yaml
lodenrogue: Name of the theme as the value is more user friendly in my opinion

jhff: I think the way to go is:
- filename -> config file theme key
- `name` -> theme name to display in the app

Let's say we add the file `solarized_dark.yaml` for the `Solarized (Dark)` theme. `solarized_dark` is used as the key and `Solarized (Dark)` can be used to display in the future command bar of the app, e.g..

jhff: In this PR `name` is not used anywhere, it's just there so users are already educated to add that instead of having to change their files in the future.

tarkah: Agreed that makes sense

src/main.rs
jhff: [fe179a6](https://github.com/squidowl/halloy/pull/92/commits/fe179a67eaa80f8396d9b5edaaf55c3fdf8b924a)

assets/themes/halloy.yaml
jhff: Should wiki be updated with `name` in the example themes?

casperstorm: I think it should yeah, @jhff.

casperstorm:
The last thing I have is that I would like to rename `halloy` (default theme) to `ferra` since thats the name of it.
Otherwise I am good to merge.

casperstorm (approved):
Lets go!

PR #93, opened 2023-07-04
Added username completion in channels

Tangdongle:
This adds nickname auto-completion for channels.

When we instantiate the text_input from the Channel's view with a list of users in that channel. 

From there, we need to differentiate between a /command and a user completion, so we add a CompletionType to each Completion Entry. /commands only auto-complete from the start of the input text, whereas name completion can occur from any word.

Finally, we replace the current word being completed without sending, so the user can continue to type.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ Added:
 - Messages from other users containing your nickname are now highlighted using the `info` colour
 - Previously sent messages can be accessed per buffer in the text input with up / down arrows
 - Themes directory where users can add their own theme files
-- Nickname completions in text input with TAB
+- Nickname completions in text input with <kbd>Tab</kbd>
 
 Changed:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Added:
 - Messages from other users containing your nickname are now highlighted using the `info` colour
 - Previously sent messages can be accessed per buffer in the text input with up / down arrows
 - Themes directory where users can add their own theme files
+- Nickname completions in text input with TAB
 
 Changed:
 
--- a/src/widget/input.rs
+++ b/src/widget/input.rs
@@ -19,7 +19,7 @@ pub type Id = text_input::Id;
 pub fn input<'a, Message>(
     id: Id,
     buffer: Buffer,
-    users: Vec<User>,
+    users: &'a [User],
     on_submit: impl Fn(data::Input) -> Message + 'a,
     on_completion: Message,
 ) -> Element<'a, Message>
@@ -54,7 +54,7 @@ pub enum Event {
 pub struct Input<'a, Message> {
     id: Id,
     buffer: Buffer,
-    users: Vec<User>,
+    users: &'a [User],
     on_submit: Box<dyn Fn(data::Input) -> Message + 'a>,
     on_completion: Message,
 }
@@ -85,7 +85,7 @@ where
 
                 state.input = input;
 
-                state.completion.process(&state.input, &self.users);
+                state.completion.process(&state.input, self.users);
 
                 None
             }
@@ -142,7 +142,7 @@ where
                         .get(state.selected_history.unwrap())
                         .unwrap()
                         .clone();
-                    state.completion.process(&state.input, &self.users);
+                    state.completion.process(&state.input, self.users);
 
                     return Some(self.on_completion.clone());
                 }
@@ -159,7 +159,7 @@ where
                     } else {
                         *index -= 1;
                         state.input = state.history.get(*index).unwrap().clone();
-                        state.completion.process(&state.input, &self.users);
+                        state.completion.process(&state.input, self.users);
                     }
 
                     return Some(self.on_completion.clone());
--- a/src/widget/input.rs
+++ b/src/widget/input.rs
@@ -119,7 +119,16 @@ where
             }
             Event::Tab => {
                 state.completion.tab();
-                None
+                if let Some(entry) = state.completion.select() {
+                    if entry.is_user() {
+                        state.input = entry.complete_input(&state.input);
+                        Some(self.on_completion.clone())
+                    } else {
+                        None
+                    }
+                } else {
+                    None
+                }
             }
             Event::Up => {
                 state.completion.reset();
--- a/src/widget/input.rs
+++ b/src/widget/input.rs
@@ -1,8 +1,8 @@
-use std::collections::VecDeque;
-use data::{user::User, input, Buffer, Command};
+use data::{input, user::User, Buffer, Command};
 use iced::advanced::widget::{self, Operation};
 pub use iced::widget::text_input::{focus, move_cursor_to_end};
 use iced::widget::{component, container, row, text, text_input, Component};
+use std::collections::VecDeque;
 
 use self::completion::Completion;
 use super::{anchored_overlay, key_press, Element, Renderer};
@@ -93,10 +93,8 @@ where
                 // Reset selected history
                 state.selected_history = None;
 
-                if let Some(command) = state.completion.select() {
-                    state.input = Completion::complete_selected_word(&state.input, &command);
-                    // We've completed a word and replace out input, reset our completion options
-                    state.completion.reset();
+                if let Some(entry) = state.completion.select() {
+                    state.input = entry.complete_input(&state.input);
                     Some(self.on_completion.clone())
                 } else if !state.input.is_empty() {
                     state.completion.reset();
--- a/src/widget/input.rs
+++ b/src/widget/input.rs
@@ -1,8 +1,10 @@
-use data::{input, user::User, Buffer, Command};
+use std::collections::VecDeque;
+
+use data::user::User;
+use data::{input, Buffer, Command};
 use iced::advanced::widget::{self, Operation};
 pub use iced::widget::text_input::{focus, move_cursor_to_end};
 use iced::widget::{component, container, row, text, text_input, Component};
-use std::collections::VecDeque;
 
 use self::completion::Completion;
 use super::{anchored_overlay, key_press, Element, Renderer};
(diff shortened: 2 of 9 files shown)


src/widget/input/completion.rs
```diff
@@ -164,18 +185,81 @@ impl Completion {
...
-    pub fn process(&mut self, input: &str) {
-        let Some((head, rest)) = input.split_once('/') else {
+    /// Convert a list of User structs into Completion Entries
+    pub fn with_users(&mut self, users: Vec<User>) -> Self {
+        self.users = users
+            .iter()
+            .map(|u| {
+                // Let's make some eternal strings
+                // TODO! Is this bad?
+                let static_str = Box::leak(u.nickname().to_string().into_boxed_str());
```
tarkah, line 195: We definitely don't want to do this! We're leaking this memory here, and we're doing it for every single user nickname every single time the view tree is built!! Memory will grow forever 😬

Instead of trying to use Entry for this, just create a new struct. The args of entry aren't useful here anyways.

Existing Entry can become Command and we can add a Nickname struct, then we can have an Entry enum that wraps both to use for the filtered entries.

src/widget/input/completion.rs
```diff
@@ -164,18 +185,81 @@ impl Completion {
...
+        };
+        format!("{} {}", original_input, completion)
+    }
+
+    pub fn process(&mut self, input: &str, users: Vec<User>) {
+        self.with_users(users);
+        let maybe_command = input.starts_with('/');
+        let Some((head, rest)) = Self::valid_completion_input(input, maybe_command) else {
             self.reset();
             return;
```
tarkah, line 248: Let's add some better separation between processing for a command vs user completion. If input starts with a slash, we can branch into a `process_command` function, otherwise `process_user`.

Then we only need to add the users info inside `process_user` compared to us always adding it here

src/widget/input.rs
```diff
@@ -63,12 +67,13 @@ where
...
 
     fn update(&mut self, state: &mut Self::State, event: Self::Event) -> Option<Message> {
+        let users = &self.users;
         match event {
             Event::Input(input) => {
                 state.error = None;
                 state.input = input;
 
-                state.completion.process(&state.input);
+                state.completion.process(&state.input, users.clone());
```
tarkah, line 76: Not sure `process` needs to take an owned Vec of users. Try `&[User]` and you can pass `&self.users` here and avoid a potentially large allocation

src/widget/input/completion.rs
```diff
@@ -164,18 +185,81 @@ impl Completion {
...
+                // TODO! Is this bad?
+                let static_str = Box::leak(u.nickname().to_string().into_boxed_str());
+                Entry {
+                    title: static_str,
+                    args: vec![],
+                    completion_type: CompletionType::User,
+                }
+            })
+            .collect();
+        self.clone()
```
tarkah, line 203: Instead of `&mut self` in the signature, you can make it `self` and avoid this `clone` here. Then calling this would be `*self = self.with_users()`

Tangdongle:
Awesome, thanks for the feedback, I'll have a look at these today

src/widget/input/completion.rs
Tangdongle: This has been updated, no more leaky strings.

src/widget/input/completion.rs
Tangdongle: I changed this function to convert and return a slice of users into Entry's, rather than storing them against the Completion's state. Then we just operate on that list and build our filtered_entries list as per usual.
I hope I'm not misunderstanding, but as far as I understand, we're passing in a list of users to the widget on every `update`, so we need to look through the current users in a channel each time, in case any new ones have been added.

tarkah:
@Tangdongle It's looking good! I've pushed a few commits to get it over the finish line, let me know what you think.

- I've cleaned up some of the types to eliminate some impossible states and streamline some things
- User completions are now only triggered if `@` is typed. This is because normal typing becomes extremely annoying when the completions list is constantly popping up for any letter. I'm open to suggestions on better alternatives to trigger user completions, but this feels fairly standard.
- I've limited the completion popup to a sliding window of 5 entries. Since many user entries can show up, the popup sometimes would get way too long. Tabbing shifts this window as expected.

tarkah (approved):

Tangdongle:
@tarkah  Legendary! Thanks for cleaning that up, hope it wasn't too much of a hassle!

Yeah, I see what you've done here, looks much cleaner. I've taken it for a spin and it looks good to me!

tarkah:
Awesome! Not at all, you put it in a really nice spot so it all came together nicely 👌

I'm going to hold off on merging until @casperstorm can test. I'm not sure I love `@` as the trigger in IRC so I want his feedback on the best UX for this.

src/widget/input.rs
```diff
@@ -51,6 +52,7 @@ pub enum Event {
 pub struct Input<'a, Message> {
     id: Id,
     buffer: Buffer,
+    users: Vec<User>,
```
hecrj, line 55: Could this borrow a slice to avoid the `clone` in `view`?

```suggestion
    users: &'a [User],
```

tarkah: Unfortunately not. We're currently depending on the IRC crate for managing the user list per channel and it doesn't expose a way to get a reference to this data.

Ideally we manage channels and users ourselves so we could make this optimization. It'd definitely benefit us in the long run as we have better control over our data.

tarkah: A middle ground solution would be to extract channel and user lists after receiving each batch of IRC messages. This will ensure they're in sync but push those types into our data so we could do this.

hecrj: Yup, just took a look at the `irc` crate. A bit unfortunate it relies heavily on interior mutation :/

tarkah: I've created a tracking issue for this here:

https://github.com/squidowl/halloy/issues/102#issue-1790818159

casperstorm:
> I'm going to hold off on merging until @casperstorm can test. I'm not sure I love `@` as the trigger in IRC so I want his feedback on the best UX for this.

I think we need to rethink the UX slightly, it not that irc idiomatic. The core functionality however is spot on.
Reason its nice to have completions list for <kbd>/</kbd> commands is that it also serves as a simple documentation for users. IRC commands can be a little awkward to remember. However, for highlighting a nick I don't think we need a completions list since you already know what (who) you are looking for, as soon as you start writing.

Because of this, the most common practise is to be able to trigger at all times when pressing <kbd>Tab</kbd> (without any prefix). This also means that we save a few clicks. Currently in this PR if i want to highlight `tarkah` I would do:

1. <kbd>@</kbd>
2. <kbd>t</kbd><kbd>a</kbd>
3. <kbd>Tab</kbd>
4. <kbd>Enter</kbd>

My suggestion would be a system where when pressing <kbd>Tab</kbd> we try to match a nick, and pressing <kbd>Tab</kbd> again cycles, without displaying our completion list. Take the following scenario as an example: We have a irc channel with three users: `[tarkah, tarsnap, casper]`.
Starting with a empty prompt:

1. <kbd>h</kbd><kbd>i</kbd><kbd>Space</kbd>
  a. `prompt: [hi ]`
1. <kbd>t</kbd>
  a. `prompt: [hi t]`
3. <kbd>tab</kbd>
  a. `prompt: [hi tarkah]`
4. <kbd>tab</kbd>
  a. `prompt: [hi tarsnap]`

Tangdongle:
Ok, I got a little something going on, lemme know what you think. Seems to work well enough from my tests (i.e. "it works on my machine").

 did notice that if you have text in the input buffer and then open a new channel, the text carries across to the newly opened channels buffer. Doesn't seem related to the change, so I'll raise a bug for it.

tarkah:
@Tangdongle I ended up pushing a larger refactor since the previous implementation didn't really jive with the new requirements and we were introducing a lot of impossible states. 

The new implementation separates user / command completion logic and should be much simpler to follow. We also can get user completions while showing the command popup for the selected command! Great for /msg and the like :D 

Thanks so much for tackling this, it was a huge help!

src/widget/input.rs
tarkah: Fixed in 3a54355af9121a9a8fc50fee815f394c4bdb537f after rebasing to #104

src/widget/input/completion.rs
```diff
@@ -354,3 +281,193 @@ impl fmt::Display for Arg {
...
+            return;
+        }
+
+        self.selected = None;
+        self.prompt = rest.to_string();
+        self.filtered = users
+            .iter()
+            .filter_map(|user| {
+                let nickname = user.nickname();
+                nickname
```
casperstorm, line 307: A tiny thing could be that we lower-cased both `rest` and `nickname` so we can tab-complete from `t` to `Tarkah`.

Not blocking but just an idea.

src/widget/input/completion.rs
```diff
@@ -354,3 +281,193 @@ impl fmt::Display for Arg {
...
+
+        if let Some(index) = self.selected {
+            self.filtered.get(index).cloned()
+        } else {
+            (!self.prompt.is_empty()).then(|| self.prompt.clone())
+        }
+    }
+}
+
+static COMMAND_LIST: Lazy<Vec<Command>> = Lazy::new(|| {
```
casperstorm, line 336: I like this 👍🏻 
Perhaps some macro magic in the future.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ Added:
 - Messages from other users containing your nickname are now highlighted using the `info` colour
 - Previously sent messages can be accessed per buffer in the text input with up / down arrows
 - Themes directory where users can add their own theme files
+- Nickname completions in text input with TAB
```
casperstorm, line 10:
```suggestion
- Nickname completions in text input with <kbd>Tab</kbd>
```

casperstorm (approved):
This is SUCH a great feature.
Thanks for the work @Tangdongle and @tarkah.

src/widget/input/completion.rs
tarkah: Good idea! 09c9e78

PR #96, opened 2023-07-04
Feat/typing history

### LINKED ISSUE #85: Typing history
It would be nice to use the up/down arrow to recall typing history in a channel

tarkah:
Resolves #85

### DIFF
--- a/src/widget/input.rs
+++ b/src/widget/input.rs
@@ -1,3 +1,5 @@
+use std::collections::VecDeque;
+
 use data::{input, Buffer, Command};
 pub use iced::widget::text_input::{focus, move_cursor_to_end};
 use iced::widget::{component, container, row, text, text_input, Component};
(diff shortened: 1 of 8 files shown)


src/widget/input.rs
```diff
@@ -91,8 +104,8 @@ where
                         }
                     };
 
-                    // Clear message, we parsed it succesfully
-                    state.input = String::new();
+                    // Clear message and add it to history
+                    state.history.push_front(std::mem::take(&mut state.input));
```
neilalexander, line 108: Can this grow unbounded?

tarkah: Good point, we should probably have some limit

casperstorm (approved):
LGTM, great feature.

src/widget/input.rs
tarkah: 64905a2

I've truncated to 100 here. Shouldn't have much memory footprint and is more than enough.

PR #99, opened 2023-07-05
Update casperstorm -> squidowl in README

Daeraxa:
Noticed that the repo had changed to the `squidowl` org, none of the links were broken as it was redirecting but it should probably reflect the current repo.

PR just updates `casperstorm` github references to `squidowl`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,11 @@
 <div align="center">
   
 # Halloy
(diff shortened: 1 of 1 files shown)


tarkah (approved):
Thanks for the catch!

PR #101, opened 2023-07-06
Make image windows only dep

tarkah:
@casperstorm can you test this, it should still work. We don't need image as a non-windows dep for now. Also removed `extern crate` as that's not idiomatic


jhff:
Still there in windows:

halloy_znqvvlazqb

PR #104, opened 2023-07-06
Sync channels & users to connection state

### LINKED ISSUE #102: Managed Channel & User lists
We have some potentially large allocations across the application during `view` when referencing channel user lists.

We should try to manage this data ourselves instead of querying it from the IRC crate client. This will allow us to pass references to them and avoid allocation.

A naive implementation would be to simply query & store this data in our `Connection` state after processing each batch of IRC messages.

tarkah:
Resolves #102. 

Will also resolve an issue we're having w/ broadcasting QUIT / PART messages to buffers because the underlying lists from `irc` crate are already updated to exclude those users. Now we can reference the lists from our state before we "sync" them so the users will still be there.

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -1,6 +1,7 @@
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashMap};
 
 use irc::client::Client;
+use itertools::Itertools;
 
 use crate::user::NickRef;
 use crate::{message, Message, Server, User};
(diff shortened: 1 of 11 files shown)


data/src/client.rs
```diff
@@ -28,13 +29,17 @@ pub enum State {
 pub struct Connection {
     client: Client,
     resolved_nick: Option<String>,
+    channels: Vec<String>,
+    users: HashMap<String, Vec<User>>,
```
casperstorm, line 33: Power to us!

casperstorm (approved):

PR #106, opened 2023-07-06
Persist input history outside widget tree

### LINKED ISSUE #98: Typing history cycling lost after switching channels
I really like the new typing history feature but it seems the history cycling is lost if you switch to another channel and come back. 

Can we add a feature where the typing history is preserved per channel (or globally)?

tarkah:
Resolves #98 

This stores typing history in app state vs widget tree so it'll persist past the buffers being re-arranged or closed


tarkah:
No changelog required as it fixes a new feature on the unreleased log

casperstorm (approved):

PR #107, opened 2023-07-06
Fix/text input persistence

### LINKED ISSUE #105: Content of message input box is lost when message input box closes
Halloy version: built from #dacb740
OS: Linux Mint 21.1

- With an open channel, type something in the text input box "Send Message..." but do not press return
- Move window - chat box closes
- Click back in window - chat box reopens but contents have been lost

This also happens when connection to the server is lost, whatever you are typing is instantly deleted as box closes on disconnect then reopens once connected.

tarkah:
Resolves #105

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Changed:
 Fixed:
 
 - The last word of a message sometimes dissapeared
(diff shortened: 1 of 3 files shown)


casperstorm (approved):
👍🏻

Issue #73, opened 2023-06-30
Parts not shown if we kill the application


casperstorm:
If a user kills the application, or connection, we don't see the `{user} has left the channel`

ineeee:
this makes sense because the client never sends a `PART` or `QUIT` to the server

tarkah:
What's the expected behavior here? Does IRC mandate a client sends QUIT before exiting? I'd imagine the server will broadcast part / quit when the client loses connection to the server.

I'd imagine we only want to implement this if we add a configuration where user can specify a QUIT / PART message when exiting the application / leaving channels.

ineeee:
the client doesn't/can't determine the quit message because the process is already dead by that point, so its up to the irc daemon.

tarkah:
> the client doesn't/can't determine the quit message because the process is already dead by that point, so its up to the irc daemon.

We actually capture the request to close the application and do some housecleaning... (flush history to disk, etc). So in theory we could send a QUIT from our client w/ a configurable message, if desired. This obv doesn't work if someone kills the running process vs graceful close.

tarkah:
I'd imagine though that we should never send PART and only send QUIT as that is effectively a superset of PART. QUIT inherently means they have parted from all channels. 

So I think the scope of this issue should be:

- Send QUIT when user closes the application. Allow for configuring the QUIT message otherwise it sends an empty reason.

casperstorm:
Fixed by #109.
Opened https://github.com/squidowl/halloy/issues/110 which continues the above discussion.

PR #109, opened 2023-07-06
brodcast quit and nick changes

### LINKED ISSUE #83: Change nick is not announced in channels
Doing /nick newnick changes your nick but does not announce it in the channels. May be we're not processing that IRC event.

casperstorm:
This PR adds a the ability to broadcast two types of Messages:
* Quit (when a user quit, it will broadcast it to all channels, and queries user was on)
* Nick (when a user change nickname, it will broadcast it to all channels, and queries user is on)

Fixes #83, #73

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ Added:
 - Previously sent messages can be accessed per buffer in the text input with up / down arrows
 - Themes directory where users can add their own theme files
 - Nickname completions in text input with <kbd>Tab</kbd>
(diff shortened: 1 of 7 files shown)


casperstorm:
Fixed by #109.
Opened https://github.com/squidowl/halloy/issues/110 which continues the above discussion.

tarkah (approved):
LGTM and tested! I added named / typed fields where applicable and fixed a small bug where we sync'd all connections & not just the server for which the batch of messages were received.

Issue #79, opened 2023-07-01
changing nick to someone's nick kills the connection


ineeee:
changing nick to someone who is online will kill the connection with ` ∙ connection to server lost`.

when testing on hexchat, the irc daemon returns `433 mynick mytarget :Nickname is already in use.`. i don't know how to check the raw tcp stream when using halloy with tls.

### steps to reproduce
1. join any channel
2. pick a random user in the channel
3. change nick to that user
4. connection will be closed

### my environment
windows 10, halloy 23.1-alpha1, connected to rizon via znc

PR #113, opened 2023-07-06
handle notice in server

### LINKED ISSUE #76: notify on tls error
currently a connection fails silently when connecting to a server with an invalid tls certificate. its possible to read the error in the `config/halloy.log` but its not user-friendly.

it would be an improvement to show an error in the window mentioning why it failed to connect.

environment: windows 10, halloy 23.1-alpha1, own irc server with self-signed cert.

### LINKED ISSUE #112: Feature request: Notify on no available username
When starting Halloy with a config.yaml from the template the application connected and disconnected without telling me why.

Found out that the problem was no available nickname in the `~/.var/app/org.squidowl.halloy/data/halloy/halloy.log` file.

Related to https://github.com/squidowl/halloy/issues/76 in a "general user error/warning" way, this could also possibly be printed directly in the application for faster user feedback?

casperstorm:
This handles all uncaught NOTICE and sends them to server.

Closes #112 
Closes #76

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,14 +2,15 @@
 
 Added:
 
-- New configuration option `dashboard.sidebar_default_action` allows controlling the pane behaviour when selecting channels in the sidebar
-- Support for RAW command
-- Messages from other users containing your nickname are now highlighted using the `info` colour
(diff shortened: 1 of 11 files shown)


tarkah:
Also resolves #79, we print the disconnect error to the buffer now

PR #115, opened 2023-07-06
Route whois messages to focused buffer

tarkah:
This routes whois replies to the currently focused buffer for the server. This makes the WHOIS from user context menu a lot more usable as it'll print it into the buffer you triggered it from.


casperstorm (approved):
Works really great. Good job.

PR #120, opened 2023-07-07
leave text is now dependent on buffer

### LINKED ISSUE #117: possible to "leave" an user
when opening the context menu on the channel list (via right click), its possible to leave a nickname.

in other words, the menu option to PART a channel also shows up for people, but i can't actually PART someone.

casperstorm:
Closes #117 

"Leave" text has been changed to be more in line with buffer context:
```
query   -> "Close query"
channel -> "Leave channel"
server  -> "Leave server"
```

Leaving a server, when you only have 1 server will leave you in a quite broken state. I'll look into that in a separate PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Changed:
 - `palette` field has been deprecated and replaced by `theme` in `config.yaml`
 - Sorting channel nicknames
 - Title headers has been changed to also display user count for channels
+- Copy change: "Leave" has been changed to "Close query" for queries
 
 Fixed:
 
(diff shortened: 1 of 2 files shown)


tarkah:
Do we want to call it Query externally to users? How common is this naming? Or should we call it something more conventional like dm?

casperstorm:
> Do we want to call it Query externally to users? How common is this naming? Or should we call it something more conventional like dm? 

I would say it's quite common. Personally, I think we should stick to query, it's more common than DM. And shorter than "Close private message".

tarkah (approved):

PR #125, opened 2023-07-07
added text_input visibility option

casperstorm:
This adds `new_buffer.input_visibility` to control input visibility.
If you have multiple buffers it can be annoying that the text jumps within the buffer when input appears, and disappears.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,13 @@
 # Unreleased
 
+Added:
+
+- Configuration option `new_buffer.input_visibility` to control input field visibility: always shown or following the focused buffer.
+
 Fixed:
 
 - Changes done in the config file are now properly applied to the old buffers
+
 
 # 2023.2 (2023-07-07)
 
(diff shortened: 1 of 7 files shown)


casperstorm:
Rebased on top of #127.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Unreleased
 
+Added:
+
+- Configuration option `new_buffer.input_visibility` to control input field visibility: always shown or following the focused buffer.
```
tarkah, line 5: I feel like we should rename this config option to just `buffer` at this point. Maybe with serde alias to `new_buffer` for backwards compatibility.

Now that we don't have the settings vs config confusion, I'm not sure `new_buffer` here makes sense as the name.

tarkah (approved):
LGTM! Just one comment about config field renaming, lmk if you agree

CHANGELOG.md
casperstorm: Agree!

PR #126, opened 2023-07-07
Allow overlay menus to layout as shrink

tarkah:
This adds a new `double_pass` widget which enables us to use the layout size from the first widget for the second widget. By doing this, we can set the first widget to `Shrink` and then safely set the second widget to `Fill` and it'll only fill within the bounds produced by the `Shrink` layout.

This is needed for our completion popup & context menu widgets because they are overlay's and using "fill" causes it to fill the entire viewport. Now we don't need to hardcode widths for these and they'll always shrink around the largest item.


casperstorm (approved):
LGTM! Great to have this dynamic instead of our hardcoded width.

PR #127, opened 2023-07-07
Better separation between config & settings

tarkah:
This refactors our settings structs to move most of the fields over to config structs. Currently when we persist Dashboard, all the settings are saved. However, most of these aren't runtime settings and things that should only be controlled via config. Since we used the same structs for both, this caused "config" to get persisted in a buffer and not updated when users updated their config.

- `config` module now contains all top-level config data structures
- `Settings` structs have been updated to only include the runtime configured settings
- We can convert from Config -> Settings where the config determines the default settings values
- For the above, the fields in config that derive a settings field are marked `pub(crate)` so they can't accidentally be referenced from the config and MUST be referenced from the settings. This ensure the application is always pulling from the right location.

@neilalexander This is the refactor I mentioned last week. Hopefully this makes more sense and provides a better separation of concerns

### DIFF
--- a/data/src/config/buffer.rs
+++ b/data/src/config/buffer.rs
@@ -1,13 +1,17 @@
-use serde::{Deserialize, Serialize};
+use chrono::{DateTime, Local, Utc};
+use serde::Deserialize;
 
-use crate::Message;
+use super::Channel;
+use crate::buffer::{Color, Nickname, Timestamp};
 
-#[derive(Debug, Clone, Deserialize, Serialize)]
+#[derive(Debug, Clone, Deserialize)]
 pub struct Buffer {
     #[serde(default)]
     pub timestamp: Option<Timestamp>,
     #[serde(default)]
     pub nickname: Nickname,
+    #[serde(default)]
+    pub channel: Channel,
 }
 
 impl Default for Buffer {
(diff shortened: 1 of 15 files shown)


data/src/config/buffer.rs
```diff
@@ -1,13 +1,17 @@
-use serde::{Deserialize, Serialize};
```
tarkah, line 1: Lol `config::buffer` and `config::channel` files already existed but weren't previously modules. This must have been from a past refactor where we forgot to delete them :grimacing:

data/src/config/channel.rs
```diff
@@ -1,20 +1,14 @@
...
-    Left,
-    #[default]
-    Right,
-}
-
-#[derive(Debug, Clone, Copy, Deserialize, Serialize)]
+#[derive(Debug, Clone, Copy, Deserialize)]
 pub struct Users {
-    pub visible: bool,
+    pub(crate) visible: bool,
```
tarkah, line 11: Since this drives the default value of a Setting, we don't allow the app to access it since it should be accessed from the settings

casperstorm (approved):
Works as expected; this is a good clean change.

casperstorm (commented):
Works as expected; this is a good clean change.
Before merge, I say we add a section to CHANGELOG where we fixed a bug with config applying new settings to old buffer or smt.

PR #140, opened 2023-07-09
Extension support

tarkah:
This PR adds initial support for the following extensions:

- `batch` 
- `echo-message`
- `labeled-response`

It appears `echo-message` is required by some bouncers (soju) for seeing sent messages across multiple connected clients. `labeled-response` has been added so we can properly track our sent messages to know if we need to apply the echo or not. These capabilities are only sent if the server supports BOTH, as `echo-message` without `labeled-response` will cause duplicate messages in halloy.

`batch` is added since `labeled-response` has a hard requirement on it per the docs.

I've not been able to fully test this yet since `ZNC` doesn't use `labeled-response`. I'll need to setup a soju instance to further test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,15 @@ cargo build --release
 cargo run --release
 ```
 
+## Capabilities
+
+Halloy supports the following IRCv3.2 capabilities
+
+- `batch`
+- `server-time`
+- `labeled-response`
+- `echo-message` (server must also support `labeled-response`)
+
 ## Why?
 <div align="center">
   <a href="https://xkcd.com/1782/">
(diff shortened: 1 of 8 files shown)


casperstorm (approved):
LGTM but not tested

tarkah:
I've tested against https://testnet.ergo.chat/ which has support for `batch`, `labeled-response` & `echo-message` and all 3 are working.

Ideally a server impements `labeled-response` which allows us to route WHOIS batches and to the proper buffer that initiated the WHOIS command. We can extend this to future command / responses in the future.

Without `labeled-response`, WHOIS / batches needs to be mapped manually via response codes and then based off our logic for current focused buffer for routing the responses. This is less than ideal.

Without `labeled-response`, we don't want to enable `echo-message` as there's no reliable way to know if the echo was sent from our client or not (multiple clients to same bouncer).

PR #143, opened 2023-07-11
Print more information for TLS errors

ljrk0:
The aatxe/irc crate defines a crate-wide Error enum in irc/src/error.rs, wrapping different errors from various kinds of sources, such as TLS errors by tls-native.  When `.to_string()` is invoked on such an enum, the error message using `#[error("...")]` is displayed.  However, they wrap the actual Tls error enum inside with its own `.to_string()` function.  In case of such errors, also include this information in our error message which is much more helpful for debugging.

image

Note that perhaps it'd be better to fix this upstream, it seems that the macro can at least also access fields from the enum, as seen here: https://github.com/aatxe/irc/blob/develop/src/error.rs#L76. Maybe something like

```
#[error("a TLS error occured: {}", tlserror)
```

with a corresponding field in the enum upstream would work, but my Rust foo is quite limited.

### DIFF
--- a/data/src/stream.rs
+++ b/data/src/stream.rs
@@ -98,12 +98,17 @@ pub async fn run(server: server::Entry, mut sender: mpsc::Sender<Update>) -> Nev
                         };
                     }
                     Err(e) => {
+                        let e = match e {
+                            // unwrap Tls-specific error enums to access more error info
+                            irc::error::Error::Tls(e) => format!("a TLS error occured: {e}"),
+                            _ => e.to_string(),
+                        };
                         log::warn!("[{server}] connection failed: {e}");
 
                         let _ = sender
                             .send(Update::ConnectionFailed {
                                 server: server.clone(),
-                                error: e.to_string(),
+                                error: e,
                             })
                             .await;
 
--- a/data/src/stream.rs
+++ b/data/src/stream.rs
@@ -98,17 +98,18 @@ pub async fn run(server: server::Entry, mut sender: mpsc::Sender<Update>) -> Nev
                         };
                     }
                     Err(e) => {
-                        let e = match e {
+                        let error = match e {
                             // unwrap Tls-specific error enums to access more error info
                             irc::error::Error::Tls(e) => format!("a TLS error occured: {e}"),
                             _ => e.to_string(),
                         };
-                        log::warn!("[{server}] connection failed: {e}");
+
+                        log::warn!("[{server}] connection failed: {error}");
 
                         let _ = sender
                             .send(Update::ConnectionFailed {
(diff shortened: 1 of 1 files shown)


ljrk0:
I've created an upstream PR as the code was much more easy than anticipated :)

tarkah:
Thanks for opening the PR!

It's idiomatic to not expose all error context in the `Display` implementation at each level. It's up to the consumer to decide if they want to dig into this context more, and if so, we can easily pattern match for the `Error::Tls` variant and print it's inner error. 

I'd prefer to see the original approach vs having the maintainer pollute the top-level error display with all context, which isn't suitable for all consumers.

So without changing upstream, a good solution might be:

```rs
let error = match e {
  irc::Error::Tls(e) => format!("a TLS error occurred: {e}"),
  _ => e.to_string(),
};
```

ljrk0:
Thank you for your comment & time!

Ah, I see – I was just mirroring what other errors in the irc crate do, but I'm a noob when it comes to app design :)  From what I can understand, you propose having the "complete" error in the log and the "top level" error in the UI?  That would definitely work for me, especially given those TLS errors often even being multiline.

tarkah:
> Thank you for your comment & time!
> 
> Ah, I see – I was just mirroring what other errors in the irc crate do, but I'm a noob when it comes to app design :)  From what I can understand, you propose having the "complete" error in the log and the "top level" error in the UI?  That would definitely work for me, especially given those TLS errors often even being multiline.

No problem! I think it's fine to use that new error string in both places. I think it'd be helpful to see it in the UI buffer since it's probably a common error where the context might help.

The match based example I provided would be idiomatic and work well.

data/src/stream.rs
```diff
@@ -98,12 +98,19 @@ pub async fn run(server: server::Entry, mut sender: mpsc::Sender<Update>) -> Nev
                         };
                     }
                     Err(e) => {
+                        let e = e.to_string() + &match e {
+                            // unwrap Tls-specific error enums to access more error info
+                            irc::error::Error::Tls(ne) => {
+                                ": ".to_string() + &ne.to_string()
+                            },
+                            _ => "".to_string()
+                        };
```
tarkah, line 107: You can do the following which will reduce the # of allocations and is a little easier to read:

```rs
let error = match e {
  irc::Error::Tls(e) => format!("a TLS error occurred: {e}"),
  _ => e.to_string(),
};
```

tarkah (changes requested):

tarkah:
@ljrk0 Looks great, thanks for the contribution :tada:

tarkah (approved):

PR #144, opened 2023-07-11
simplify colors

casperstorm:
This PR simplifies the our `Subpalette` - with the goal of making it simpler to manage colors.
We are moving from a long and detailed set of variants, to a more simple list of just a few variants.

Now
```rust
    pub base: Color,
    pub light: Color,
    pub lighter: Color,
    pub lightest: Color,
    pub dark: Color,
    pub darker: Color,
    pub darkest: Color,
    pub low_alpha: Color,
    pub med_alpha: Color,
    pub high_alpha: Color,
```

One Issue previously to this PR was that light themes looked washed out, and hard to read in a few places. This is now fixed by handling transparency on a theme color level. If the theme is dark, we use different values than if its light.

This is probably not perfect just yet, we could adjust value even further, but main idea here is that's its easier to adjust going forward than before.

To test, try both dark and light themes and make sure they are readable as expected.

### DIFF
--- a/data/src/palette.rs
+++ b/data/src/palette.rs
@@ -96,14 +96,6 @@ pub fn alpha(color: Color, alpha: f32) -> Color {
     Color { a: alpha, ..color }
 }
 
-pub fn lightness(color: Color, amount: f32) -> Color {
-    let mut hsl = to_hsl(color);
-
-    hsl.lightness = amount;
-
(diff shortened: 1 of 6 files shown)


tarkah:
LGTM!

tarkah (approved):

PR #145, opened 2023-07-11
Route whois on last sent buffer

### LINKED ISSUE #132: Whois only goes to channel window if channel has focus
Halloy 2023.2

Doing whois on a user in a window that is not "selected" will send whois to server window.
It should probably go to the window where the whois was performed, no matter the window focus.

tarkah:
Resolves #132 

Instead of routing whois based on focused buffer, we add some state to track the last buffer to send WHOIS command and use that. This solution works for both context menu WHOIS when focused / unfocused and using text input command.


casperstorm (approved):
Perfect!

PR #146, opened 2023-07-11
Rename config key to buffer

tarkah:
Small fix now that #127 is merged. Renames `new_buffer` in config to `buffer`. I've kept the alias in on `new_buffer` so old config formats don't break

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,11 @@
 
 Added:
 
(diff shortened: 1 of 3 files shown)


casperstorm:
LGTM. Will approve.
Small thing before merge; could you update CHANGELOG where `new_buffer` is referenced in the unreleased section?

casperstorm (approved):

PR #150, opened 2023-07-13
Update `iced` to official `master`

hecrj:
After https://github.com/iced-rs/iced/pull/1912 and https://github.com/iced-rs/iced/pull/1953, we can finally rely on the official `master` branch of `iced`.


tarkah (approved):
Thanks @hecrj! It feels good to be back on `master`  🎉

PR #151, opened 2023-07-13
map to our own `Config` struct

casperstorm:
This PR creates our own `Config` struct instead of wrapping the internal `irc::client::data::Config`.
We then use a `impl From` to go back to a `irc::client::data::Config`.  

In this process i've also added a hardcoded `CTCP VERSION` so when people requests it from a user who use Halloy they get `Halloy {VERSION}`.

### DIFF
--- a/data/src/server.rs
+++ b/data/src/server.rs
@@ -1,8 +1,9 @@
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashMap};
 use std::fmt;
-use std::ops::Deref;
 
 use serde::{Deserialize, Serialize};
+
+const VERSION: &str = include_str!("../../VERSION");
 
 #[derive(Debug, Clone, Hash, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
 pub struct Server(String);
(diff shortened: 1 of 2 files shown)


data/src/server.rs
```diff
@@ -20,19 +21,107 @@ impl AsRef<str> for Server {
...
+            ping_time: config.ping_time,
+            ping_timeout: config.ping_timeout,
+            burst_window_length: config.burst_window_length,
+            max_messages_in_burst: config.max_messages_in_burst,
+            should_ghost: config.should_ghost,
+            ghost_sequence: config.ghost_sequence,
+            channel_keys: config.channel_keys,
+            version: Some(format!("Halloy {VERSION}")),
+            owners: vec![],
+            source: None,
```
casperstorm, line 119: Not entirely sure how to do a `CTCP SOURCE` request, and what it should display.

tarkah (approved):
Makes sense. Not all fields we need exposed to users. And now we can extend this with our own settings if need be. One small nit.

data/src/server.rs
```diff
@@ -20,19 +21,107 @@ impl AsRef<str> for Server {
...
 #[derive(Debug, Clone, Default, Deserialize)]
-pub struct Config(irc::client::data::Config);
-
-impl From<irc::client::data::Config> for Config {
-    fn from(config: irc::client::data::Config) -> Self {
-        Self(config)
-    }
+pub struct Config {
+    /// The client's nickname.
+    pub nickname: Option<String>,
```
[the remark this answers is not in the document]
casperstorm: Can you check [b2f9955](https://github.com/squidowl/halloy/pull/151/commits/b2f99550c40269734e2e1b11d1aa424fdadc19a8), @tarkah.

data/src/config/server.rs
```diff
@@ -0,0 +1,110 @@
+use std::collections::HashMap;
+
+use serde::Deserialize;
+
+const VERSION: &str = include_str!("../../../VERSION");
```
tarkah, line 5: This is available as `data::environment::VERSION` already

data/src/config.rs
```diff
@@ -8,20 +8,24 @@ use thiserror::Error;
 pub use self::buffer::Buffer;
 pub use self::channel::Channel;
 pub use self::dashboard::Dashboard;
+pub use self::server::Server;
+
+use crate::environment;
 use crate::palette::Palette;
-use crate::{environment, server};
+use crate::server::Map;
```
tarkah, line 15: Map here has no context, I'd keep the module namespace for clarity... `server::Map` vs `Map`

Issue #141, opened 2023-07-09
Portable mode


da3dsoul1:
Is there any way to run this program portable? For it to store the configs inside the exe folder on windows?

tarkah:
Not currently, but support should be easy. Is there a standard way for one to invoke a program into "portable mode" on Windows? As in, a common environment variable or CLI flag?

casperstorm:
@smakkerlak do you know anything about this on Windows?

smakkerlak:
For a portable windows app i would just assume executable and config file in a folder. Just run as current user, nothing fancy going on. I don't think you are currently doing anything that would require privilege escalation so you could do the minimum effort of checking for a config local to the executable and move on to the default system location if nothing is found.

You could also look into a configuration hierarchy, where closest in scope means highest precedence, essentially overlaying conf elements. I'm not sure how much value an application like Halloy would gain from this, but it would be neato. Could be a lot of cross platform fiddling if there isn't an off the shelf fix.

I'll help test if you need it.

Daeraxa:
There are a couple of solutions that I've seen for this, one of which I think is exactly what @smakkerlak is advocating.

The easiest is to literally just throw the `config.yaml` into the same directory as the binary, if Halloy sees it there then it prefers that over any other file and writes its other data to the same place.

Another method (just adding in terms of possibility rather than necessarily thinking it is the best idea) would be to have an environmental variable that controls the location manually. So for example you could have a variable `HALLOY_CONFIG` which allows you to manually control where Halloy would look to create or find its config and data. This isn't really great for portability though and certainly for Linux it would be better to just support XDG properly instead.

tarkah:
I think side by side config to binary is the easiest approach 👍

I believe this should only be in scope for Windows. In the unix world this isn't as much of an issue or concern.

Daeraxa:
> I believe this should only be in scope for Windows. In the unix world this isn't as much of an issue or concern.

Out of interest, why not? The code for it (I imagine) would be fairly universal in terms of the actual logic and I generally find it is easier to not have to handle exceptions to the rule in the long term.

da3dsoul1:
Thank you so much!

PR #163, opened 2023-07-16
notifications (and windows msi installer)

### LINKED ISSUE #203: Deskto notifications?
Hi, thanks for the great IRC client.

I was wondering if you were interested in adding support for desktop notifications? On Linux, there's the `libnotify` system, and a simple binary named `notify-send`. This would help with getting a message while working with another application, a common user behavior.

Thanks! :smile:

casperstorm:
This PR adds notification support, along a few events: `Connected`, `Reconnected` and `Disconnected`.
We should add more, but for now the base should be ready.

Reason this PR also includes a new installer for Windows is that in order to get a icon on the notification on Windows we need a `app_id`, and the easiest way (that i know of) is to do it through a `.wxs` file. For that reason we might as well produce a `.msi` installer as our "primary" windows installation. Going forward we could then use a default `.exe` asset for our "portable" installer (#141) - this would then be similar to how Alacritty does releases and assets. 

In any case, with the app installed through the `.msi` installer notifications has a small icon which looks great.

### Sounds
The configuration can take a string to a sound. Each OS has a different set of default sounds, depending on OS version as well. However in order for the user to know what to pick, we need to make a entry to our wiki with the different defaults and how to find them (see below).

This entire branch is currently untested on Linux - @tarkah could you test this?

#### macOS (`/System/Library/Sounds/`)
```
Basso
Blow
Bottle
Frog
Funk
Glass
Hero
Morse
Ping
Pop
Purr
Sosumi
Submarine
Tink
```

#### Windows (https://github.com/allenbenz/winrt-notification/blob/main/src/lib.rs#L70-L75)
```
Default
IM
Mail
Reminder
SMS
```

#### Unix (https://github.com/hoodie/notify-rust/blob/main/src/hints.rs#L68)
```
message-new-instant
[...]
```

Fixes #203


casperstorm:
@tarkah what do you think we should add before merge?
- DM notification?
- Highlight notification?

data/src/client.rs
```diff
@@ -57,6 +62,7 @@ pub enum Event {
     Single(message::Encoded, Nick),
     WithTarget(message::Encoded, Nick, message::Target),
     Brodcast(Brodcast),
+    Notification(Notification),
```
tarkah, line 65: We need to add `message::Encoded` here so we still record the message in history once returned from here

data/src/client.rs
```diff
@@ -683,6 +717,15 @@ impl Client {
                 Retry,
             }
 
+            match self.highlight_blackout {
+                HighlightBlackout::Blackout(instant) => {
+                    if now.duration_since(instant) >= HIGHLIGHT_BLACKOUT {
```
tarkah, line 722: This can go above / outside the `for` loop

data/src/client.rs
```diff
@@ -364,10 +374,19 @@ impl Client {
...
-                    if user.nickname() == self.nickname() && context.is_some() {
+                    // Highlight notification
+                    if message::reference_user(user.nickname(), self.nickname(), text)
+                        && self.highlight_blackout.allow_highlights()
+                    {
+                        events.push(Event::Notification(
+                            message.clone(),
+                            self.nickname().to_owned(),
+                            Notification::Highlight(user, channel.clone()),
+                        ));
```
tarkah, line 387: Now that we include encoded message here, you can directly return here and remove the let events = vec![]; above

tarkah (approved):

PR #166, opened 2023-07-17
Feat/message filtering

tarkah:
Adds the ability to specify which type of messages you don't want to see. Currently supports `part`, `quit` and `join`.

This refactors message type and is a breaking change. Future changes to message shouldn't break, however my dumb decision to serialize in `bincode` forced us into this. Now I serialize w/ `json` which allows us to use a lot of `serde` magic for forwards compatible changes.


casperstorm (approved):

casperstorm:
Only thing missing is CHANGELOG - but approved it :)

PR #167, opened 2023-07-17
autocomplete joined channels

### LINKED ISSUE #138: feature request: channel completion
extending #69 and #93, it would be great if pressing tab in the input field also triggered completion for the channels the user is in.

casperstorm:
Fixes #138.

This adds the possibility to autocomplete channels. To trigger you have to write a <kbd>#</kbd> followed by <kbd>Tab</kbd>.   
Eg: <kbd>#</kbd><kbd>h</kbd><kbd>Tab</kbd> => `#halloy`.

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -397,6 +397,12 @@ impl Map {
             .unwrap_or_default()
     }
 
+    pub fn get_our_channels<'a>(&'a self, server: &Server) -> &'a [String] {
+        self.connection(server)
+            .map(|connection| connection.channels())
+            .unwrap_or_default()
+    }
+
     pub fn iter(&self) -> std::collections::btree_map::Iter<Server, State> {
         self.0.iter()
     }
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -397,7 +397,7 @@ impl Map {
             .unwrap_or_default()
     }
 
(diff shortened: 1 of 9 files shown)


data/src/client.rs
```diff
@@ -397,6 +397,12 @@ impl Map {
             .unwrap_or_default()
     }
 
+    pub fn get_our_channels<'a>(&'a self, server: &Server) -> &'a [String] {
```
tarkah, line 400: Nit: not sure `our` is needed vs `get_channels` since we never store channels that _arent_ joined.

src/widget/input/completion.rs
```diff
@@ -22,18 +22,36 @@ impl Completion {
...
+            .chars()
+            .last()
+            .map(|last| last.is_whitespace())
+            .unwrap_or_default();
+        let words: Vec<&str> = input.split_whitespace().collect();
+        let is_channel = !last_char_is_space
+            && words
+                .last()
+                .map(|word| word.starts_with('#'))
+                .unwrap_or_default();
```
tarkah, line 36: Let's push this complexity into `Text::process` which can delegate to `process_users` or `process_channels` internally.

data/src/client.rs
casperstorm: Good call.

tarkah (approved):
Works great!

PR #173, opened 2023-07-20
Provide an appimage build

Daeraxa:
So this is something I've been working on for probably a bit too long as I kept have issues (mostly surrounding the metadata file).

This is very much a draft as I suspect there are a lot of things here surrounding that metainfo file and the .desktop files (and possibly the icons too) that should be discussed as to how best to handle them with as few manual steps as possible.

So really this is a functional proof of concepts:

Prereq: https://github.com/TheAssassin/appimagecraft

Appimagecraft is a tool for building appimages that makes it really easy to provide a simple build script - it uses linuxdeploy to handle the building of the appdir then bundles it all up for you into your appimage.

It is provided as an appimage itself so what I did was just copy it to my `.local/share/bin` (which is on my path) and renamed it just `appimagecraft` so it was easy to run the commands but you can obviously just do whatever you think best to run it.

Appimagecraft requires a new file in the root of the repo - `appimagecraft.yml` which is used to store the config.

To break it down:
- `project` & `version`: kind of self explanatory but for version I decided to be fancy and automatically grab the version info from the `cargo.toml` - there are a billion other ways to do this obviously.
- `build`: essentially just runs `cargo build --release` to build the binary. If you already have it built as part of the release process this can just be skipped (I think).
- `scripts`: handles all the non-automated stuff. Creation of a .desktop file (if needed) along with the metainfo file handling.
- `appimage`: Controls the actual building of the appimage with linuxdeploy - basically an automated way of telling it where the binary is, the icons and the .desktop file

Then to run it you just run the `appimagecraft` appimage in the repo root.

Other changes I've needed to make to get this to work properly:
- `assets/linux` as some of the data in the flatpak metadata file and .desktop just don't seem to work. I've made a specific version of the metadata file for the appimage and the .desktop is being made by the script but that is easy enough to omit and just let linuxdeploy handle it directly.
- `logos` I've created a couple of scaled versions of the logos in different sizes. Ideally we could have a scaleable one here instead but this is what I've needed to do to get this working at least.

So yeah, happy to take any advice on how this should be handled if you still want to progress with an appimage. This is just one solution of many potential ones but I tried to go the route to make it as simple as possible and as easily understood as possible.


appimagecraft.yml
```diff
@@ -0,0 +1,34 @@
+version: 1
+
+project:
+  name: org.squidowl.halloy
+  # version: 0.1.0
+  version_command: awk -F'[ ="]+' '$1 == "version" { print $2;exit;}' "$PROJECT_ROOT"/Cargo.toml
```
tarkah, line 6: We have a `VERSION` file in the root of the repo

assets/linux/org.squidowl.halloy.appdata.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
```
tarkah, line 2: This file already exists at `assets/halloy.appdata.xml`

appimagecraft.yml
```diff
@@ -0,0 +1,34 @@
...
+      Comment=IRC client written in Rust
+      Type=Application
+      Keywords=IM,Chat;
+      Categories=Network;IRCClient;
+      Exec=halloy
+      Icon=logo
+      StartupWMClass=halloy
+      MimeType=x-scheme-handler/irc;x-scheme-handler/ircs;
+      Terminal=false
+      EOF
```
tarkah, line 30: This file already exists at `assets/halloy.desktop`

appimagecraft.yml
```diff
@@ -0,0 +1,34 @@
...
+
+appimage:
+  linuxdeploy:
+    extra_args: -d "$BUILD_DIR"/halloy.desktop -e "$PROJECT_ROOT"/target/release/halloy -i "$PROJECT_ROOT"/assets/linux/512/logo.png -i "$PROJECT_ROOT"/assets/linux/256/logo.png -i "$PROJECT_ROOT"/assets/linux/128/logo.png
```
tarkah, line 34: Instead of committing these icon files to the repo,  can we generate them during `post_build` step? We have a `scripts/generate-icons.sh` that can be adapted to maybe create multiple sizes w/ a configurable output folder (like under $BUILD_DIR)

tarkah:
Thanks @Daeraxa this is awesome! I left some comments, I'm sure we can figure out how to fix our existing `.desktop` / `appstream` files so we don't need any duplication.

appimagecraft.yml
```diff
@@ -0,0 +1,34 @@
...
+  script:
+    commands:
+      - cargo build --release
+
+scripts:
+  post_build:
+    - mkdir -p "$BUILD_DIR"/AppDir/usr/share/metainfo
+    - cp "$PROJECT_ROOT"/assets/linux/org.squidowl.halloy.appdata.xml "$BUILD_DIR"/AppDir/usr/share/metainfo/
+    - |2
+      cat > "$BUILD_DIR"/halloy.desktop <<EOF
```
tarkah, line 18: I suspect the reason the desktop file and meta file weren't working in conjunction is because the desktop file should be named `org.squidowl.halloy.desktop` since our meta file has `  <launchable type="desktop-id">org.squidowl.halloy.desktop</launchable>` set.

I noticed in your custom meta file, you changed it to be ` <launchable type="desktop-id">halloy.desktop</launchable>`.

If we use the original `.desktop` and `appstream.xml` files, but name this `org.squidowl.halloy.desktop` here instead of `halloy.desktop`, does it work?

appimagecraft.yml
```diff
@@ -0,0 +1,34 @@
...
+
+build:
+  script:
+    commands:
+      - cargo build --release
+
+scripts:
+  post_build:
+    - mkdir -p "$BUILD_DIR"/AppDir/usr/share/metainfo
+    - cp "$PROJECT_ROOT"/assets/linux/org.squidowl.halloy.appdata.xml "$BUILD_DIR"/AppDir/usr/share/metainfo/
```
tarkah, line 16: You can use `install` for a single command:

```
install -Dm644 ./assets/halloy.appdata.xml "$BUILD_DIR"/AppDir/usr/share/metainfo/org.squidowl.halloy.appdata.xml
```

appimagecraft.yml
```diff
@@ -0,0 +1,34 @@
+version: 1
+
```
tarkah, line 2: Can we put this file under an `assets/appimage` folder?

appimagecraft.yml
tarkah: It's also possible we need to put it under `$BUILD_DIR"/AppDir/usr/share/applications/org.squidowl.halloy.desktop`

tarkah:
So when building the `flatpak` CI I referenced the `wezterm` repo since he has it all setup very nicely over there. I'd suggest we copy the approach used in his CI script here. This is self-contained and we won't need an external tool `appcraft` since all it's handling for us is copying some files over.

https://github.com/wez/wezterm/blob/main/ci/appimage.sh

PR #175, opened 2023-07-23
added `notify-away` support

casperstorm:
This adds support for `notify-away` extension.
We call `WHO` when joining a channel to get initial state, and then keeps it synced with `AWAY` responses.
Away users is appear slightly transparent. This idea is stolen from Textual.

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -65,6 +65,7 @@ pub struct Client {
     registration_step: RegistrationStep,
     listed_caps: Vec<String>,
     supports_labels: bool,
+    supports_away_notify: bool,
 }
 
 impl fmt::Debug for Client {
@@ -111,6 +112,7 @@ impl Client {
             registration_step,
             listed_caps: vec![],
             supports_labels: false,
+            supports_away_notify: false,
         }
     }
 
@@ -268,6 +270,9 @@ impl Client {
 
                     let contains = |s| self.listed_caps.iter().any(|cap| cap == s);
 
+                    if contains("away-notify") {
+                        requested.push("away-notify");
+                    }
                     if contains("server-time") {
                         requested.push("server-time");
                     }
@@ -307,6 +312,9 @@ impl Client {
 
                 if caps.contains(&"labeled-response") {
                     self.supports_labels = true;
+                }
+                if caps.contains(&"away-notify") {
+                    self.supports_away_notify = true;
                 }
 
                 let supports_sasl = caps.iter().any(|cap| cap.contains("sasl"));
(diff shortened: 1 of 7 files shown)


data/src/client.rs
```diff
@@ -469,6 +482,35 @@ impl Client {
...
+                            list.insert(user);
+                        }
+                    }
+                }
+            }
+            Command::AWAY(args) => {
+                let away = args.is_some();
+                let user = message.user()?;
+                let channels = self.user_channels(user.nickname());
+                for channel in channels.iter() {
```
tarkah, line 505: Instead of using `self.user_channels` here, I'd just `self.chanmap.values_mut()`, this should save some allocations.

data/src/client.rs
```diff
@@ -461,6 +469,11 @@ impl Client {
                 }
             }
             Command::JOIN(channel, _) => {
+                if self.supports_away_notify {
+                    // Sends WHO to get away state on users.
+                    let _ = self.sender.try_send(command!("WHO", channel));
+                }
```
tarkah, line 475: I noticed we're sending WHO every-time someone joins a channel. We should ensure this is only sent when we're the user: `user.nickname() == self.nickname()`

tarkah (commented):
Looks great @casperstorm! A couple minor things. Also, we may want to handle `RPL_NOWAWAY` and `RPL_UNAWAY` so when we issue an `/away` command, it'll update our nick in the nicklists. I noticed these are what's sent when we're the ones sending the `/away`.

tarkah:
I noticed a message from libera about "rate-limiting" so I'm not sure `WHO` was sent on each of my joined channels. Thoughts around how we should handle this? Do we know what other clients do to circumvent this issue?

data/src/client.rs
casperstorm: Ah yes, good call.

data/src/client.rs
casperstorm: Ah great optimization.

casperstorm:
> I noticed a message from libera about "rate-limiting" so I'm not sure `WHO` was sent on each of my joined channels. Thoughts around how we should handle this? Do we know what other clients do to circumvent this issue?
> 
> Maybe we keep track of a map of channels and the last time they're polled for `WHO`. If they haven't been polled, we can execute WHO. We check this every X seconds / minutes. We only ensure it's sent once if `away-notify` is set, otherwise we can use this as a mechanism to poll every X if that feature isn't enabled?

This might be because of the WHO being called for all users joining previously. Can you test again? I don't think we have to make a special case for it - doesn't seem others do that 🤔

tarkah:
> > I noticed a message from libera about "rate-limiting" so I'm not sure `WHO` was sent on each of my joined channels. Thoughts around how we should handle this? Do we know what other clients do to circumvent this issue?
> > Maybe we keep track of a map of channels and the last time they're polled for `WHO`. If they haven't been polled, we can execute WHO. We check this every X seconds / minutes. We only ensure it's sent once if `away-notify` is set, otherwise we can use this as a mechanism to poll every X if that feature isn't enabled?
> 
> This might be because of the WHO being called for all users joining previously. Can you test again? I don't think we have to make a special case for it - doesn't seem others do that thinking

I'm still getting it:

```
13:43:29.812:TRACE -- Message received => Message { tags: [Tag { key: "time", value: Some("2023-07-24T20:43:29.722Z") }], source: Some(Server("erbium.libera.chat")), command: Numeric(RPL_TRYAGAIN, ["tarkah", "WHO", "This command could not be completed because it has been used recently, and is rate-limited."]) }
```

casperstorm:
Gotcha, yeah let's then add a system which keeps track which we can also use for polling.

tarkah (approved):
Awesome feature!

PR #176, opened 2023-07-26
Add portable mode

### LINKED ISSUE #141: Portable mode

tarkah:
Resolves #141 

Checks if there is a `config.yaml` in the same directory as the running executable. If so, it sets the halloy config & data directories to that directory. This allows users to have a "portable" mode where all data is used / saved alongside the executable.


casperstorm (approved):
Awesome! Can you add entry to changelog and then merge?

Issue #66, opened 2023-06-30
Create a shortcut to config directory


casperstorm:
As a user it can be cumbersome to find the config.yaml. I would be nice to have a quick way to open the directory from within the app. Possible in conjunction with #24.

ineeee:
is it possible to extend this to also have a "reload config" command? (i don't know if the config can be reloaded live)

Issue #24, opened 2023-06-13
Create a Slack like command-bar


casperstorm:
As a user, I want to be able to press a shortcut (<kbd>Ctrl</kbd> + <kbd>K</kbd> (macOS: <kbd>CMD</kbd> + <kbd>K</kbd>) to bring up a Slack-like command-bar. From the command-bar I can do different commands like closing the focused buffer, splitting into a new buffer.

<img src="https://github.com/squidowl/halloy/assets/2248455/4b9226bf-8db6-4d22-ac48-c706c3be3ad9" width="600px">

Acceptance Criteria:

* Pressing the specified shortcut should activate the Command-bar.
* The Command-bar should be anchored to the top of the window and overlay over everything.
* Users should be able to enter commands in the Command-bar.

PR #177, opened 2023-07-26
Add a Command Bar

jhff:
Added a Slack-like Command Bar to tackle issue #24.

Currently with a single option: open the config directory, which also tackles #66, but should be pretty easy to extend the command list from now on, e.g. change theme.

**Pressing Ctrl/Cmd + K should open it.**

halloy_hguHE9Cc0Q

I feel like placing it in the bottom looks better than at the top, unless the design is changed or we make it "pop over" the dashboard at the top. But let me know what you think.


tarkah:
Code looks great! I'll defer to @casperstorm for styling, but here's some ideas:

- command bar is in floating modal with background semi transparent 
- drop down from top / center, we still probably want to semi-transparent background

In either case, I think the command bar should be Shrink.

jhff:
@tarkah those last changes are amazing 👌

tarkah:
> @tarkah those last changes are amazing ok_hand

Thanks!

tarkah:
@jhff I've fixed an issue where closing the command bar would leave your buffer text input unfocused. Now, when clicking away / hitting ESC, the focused buffer has it's text input refocused.

This will be very important for chaining commands -> typing

jhff:
Missed that, well spotted!

casperstorm:
Looks great!
I pushed a small commit ([43f8495](https://github.com/squidowl/halloy/pull/177/commits/43f84953dc557cf6c4c0b54fe6edfb0165e95a60))

- Removed `height_margin` so we now position command bar on a offset to the first buffer on macOS. I think it gives a better "hover" effect:
- Slightly updated theme for combobox menu

Screenshot 2023-07-28 at 20 54 05

tarkah:
@casperstorm Looks great! I say we add some initial options before merging so it's initially useful. Do we have a list of desired commands?

casperstorm:
Here's a list of ideas I had in mind (not all has to be in this PR)

- [x] Open config directory 
- [ ] Toggle zen mode (hide side menu)
- [ ] Maximize/restore buffer
- [ ] Change buffer (populate it with our channels, eg "Change to: #halloy")
- [ ] Open new buffer (open a empty buffer)

jhff:
Added:

- Maximize/restore buffer
- Open new buffer (open a empty buffer)
- Close buffer

@casperstorm what do you mean by:
> Change buffer (populate it with our channels, eg "Change to: #halloy")

Change currently focused buffer to a different channel?

casperstorm:
> Change currently focused buffer to a different channel?

Yeah that's what I mean!

jhff:
halloy_XTUQv3HEqW

src/screen/dashboard.rs
```diff
@@ -302,6 +302,24 @@ impl Dashboard {
...
                                 }
                                 command_bar::Buffer::Close => self.close_pane(),
+                                command_bar::Buffer::Replace(buffer) => {
+                                    let mut commands = vec![];
+
+                                    if let Some(pane) = self.focus {
+                                        if let Some(state) = self.panes.get_mut(&pane) {
+                                            state.buffer = Buffer::from(buffer);
+                                            self.last_changed = Some(Instant::now());
+                                            self.focus = None;
```
tarkah, line 312: Nit: you can use `self.focus.take` so you don't need to set it to `None` after.

src/screen/dashboard/command_bar.rs
```diff
@@ -125,12 +132,53 @@ impl std::fmt::Display for Command {
...
+                }
+                data::client::State::Disconnected => {}
+            }
+        }
+
+        let mut buffers = vec![Buffer::Maximize, Buffer::New, Buffer::Close];
+        buffers.extend(channels.iter().cloned().map(Buffer::Replace));
+        buffers
+    }
+}
```
tarkah, line 154: Now that #180 is merged, there is a helper method in `src/dashboard` for getting all buffer entries, `all_buffers`. I'd rather we use that and pass in the list of buffers here so this doesn't depend on anything complex. 

Also, it includes the server & query buffer types which we probably want to include in command bar for switching

tarkah: Note, we may want to remove the `open_buffers` from `all_buffers` to generate a `closed_buffers` list to populate command bar. We shouldn't allow opening the same target in multiple buffers.

In the future, we can maybe add a `switch with <buffer>` command which handles already open buffers.

casperstorm:
The channel switching through commandbar is great.

casperstorm (approved):

src/screen/dashboard.rs
jhff: Done

src/screen/dashboard/command_bar.rs
jhff: Done with `all_buffers`.

Also removed buffer commands when no buffer is focused, except `New buffer`

tarkah (approved):
Killer feature! Thanks, @jhff

PR #183, opened 2023-07-28
Ci/package linux

tarkah:
This packages a linux archive as part of our release process and uploads it to the release.

Confirmed working: https://github.com/squidowl/halloy/actions/runs/5696256849


casperstorm (approved):
Worked first try I bet 😁

PR #184, opened 2023-07-28
Don't spam server history w/ our WHO request

tarkah:
We shouldn't store the WHO replies in history that we receive when we query WHO on launch. This is spam the user didn't request themselves.


casperstorm (approved):
😬 good point

Issue #119, opened 2023-07-07
Feature request: Click to insert user as reply into "Send message" box


Daeraxa:
You can currently right click in the user list to do a `Whois` about or `Message` to a user within a channel but left clicking currently does nothing (although shows a "hand" cursor").

Could left clicking maybe be used to insert `User:` into the message box prompt if empty? Or if not empty insert it at the start of the message? Or could this be worth adding to the right click menu as `Reply` or `Mention`?

casperstorm:
I like this, maybe:

* single click: insert user into text_input
* double click: open query

PR #186, opened 2023-07-29
double clicking a user

casperstorm:
I was implementing https://github.com/squidowl/halloy/issues/119 - initially I wanted to insert user to text input on single click, and open query on double click. But I didn't like having two actions on left click during testing. I found the single click feature quite useless now we have good autocompletion of nicknames, it simply feel faster to just type it rather than using your mouse, and then move hands back to the keyboard to continue the sentence.

However, double clicking to open a query was quite common across other clients so i've kept that.


tarkah:
Agreed that having different actions tied to single and double click is not a great pattern. However, I'm thinking the single click is more useful than double click. We already have DM in the right click context menu. I think single click to populate name in text input is the better complement and will be used more by users who are familiar with mouse interactions. 

For single click: If text input is empty, enter `nick: ` otherwise we just enter `nick` at the end of the current content?

tarkah (approved):
Looks great!

PR #189, opened 2023-08-02
on connect

### LINKED ISSUE #67: Ability to run a command on connect
As a user it can be nice to execute a command as soon as you connect to a server.  
This can be used to AUTH, among other things.

casperstorm:
This adds the option to add commands which will be executed once connected.
I've kept it to use same format as our commands in Halloy to make it similar.

Eg;
```yaml
servers:
  libera:
    nickname: casperstorm
    server: irc.libera.chat
    port: 6697
    use_tls: true
    on_connect:
      - "/msg tarkah hello halloy is pretty great"
      - "/msg NickServ IDENTIFY foo bar"
```

Fixes #67

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -411,15 +411,6 @@ impl Client {
                 let nick = args.first()?;
                 self.resolved_nick = Some(nick.to_string());
 
-                // Loop on connect commands.
-                for command in self.config.on_connect.iter() {
-                    if let Ok(cmd) = crate::command::parse(command, None) {
-                        if let Ok(command) = proto::Command::try_from(cmd) {
-                            let _ = self.sender.try_send(command.into());
-                        };
-                    };
-                }
-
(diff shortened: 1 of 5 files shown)


data/src/client.rs
```diff
@@ -411,6 +411,15 @@ impl Client {
...
 
+                // Loop on connect commands.
+                for command in self.config.on_connect.iter() {
+                    if let Ok(cmd) = crate::command::parse(command, None) {
+                        if let Ok(command) = proto::Command::try_from(cmd) {
+                            let _ = self.sender.try_send(command.into());
+                        };
+                    };
+                }
+
```
tarkah, line 422: Can we set this _after_ the nick password / ghost attempt? Just in-case user has that configured, so these commands are sent after they auth / recover

tarkah (approved):
Looks great! Just one small suggestion

PR #193, opened 2023-08-03
invite-notify (and userhost-in-names)

casperstorm:
This adds support for `invite-notify` and `userhost-in-names`.
For the latter, nothing has to be done.

Screenshot 2023-08-03 at 22 00 36


tarkah (approved):
Looks good!

PR #196, opened 2023-08-10
maximum message length

### LINKED ISSUE #128: irc lines shouldn't go beyond 512 bytes
right now its possible to send lines which exceed rfc1459's `510 bytes + CRLF`, which means messages appear to go through but they get cut off by the irc daemon.

i think there is a irc command (`003` or `004`?) that tells the client the max length of irc lines, but maybe it was for channel names or something like that.

environment: halloy 2023.2, windows 10, rizon via znc.

casperstorm:
Fixes #128.

This ensures we do not go above a certain message byte length.
- If its a regular message, we truncate it.
- If its a command, we alert to user that command is too long.

### DIFF
--- a/data/src/input.rs
+++ b/data/src/input.rs
@@ -1,16 +1,25 @@
 use chrono::Utc;
 use irc::proto;
+use irc::proto::format;
 
 use crate::time::Posix;
 use crate::user::NickRef;
 use crate::{command, message, Buffer, Command, Message, Server, User};
 
-pub fn parse(buffer: Buffer, input: &str) -> Result<Input, command::Error> {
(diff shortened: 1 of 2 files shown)


casperstorm:
@tarkah can you take a look at this?

tarkah:
@casperstorm Checkout 7177351. I've updated the strategy to check the "encoded" string against a `512` byte limit. We now validate that any user inputted value will fit this `512` limit or else we display an error to them.

I personally think the error is better than truncating the message as it allows the user to truncate themselves vs them thinking it may be a bug in our client.

tarkah (approved):

PR #209, opened 2023-11-05
fix config error on missing nickname color

fawni:
config errors if `buffer.nickname.color` is missing, but `Color::Unique` variant is set as default.

06_11_2023-00_26_halloy_iBQ9nwq


casperstorm (approved):
Thanks 👏🏻

PR #210, opened 2023-11-06
WIP: IRC formatting

rniii:
Adds support for formatting via IRC escapes

Bots especially use these for formatting their messages, and currently they render as boxes plus numbers, which is a bit annoying

## Progress

Currently parses most escapes except hex colors. Only the 0-15 colors are currently implemented, everything else is ignored. A default for these colors also needs to be discussed (I just put the mIRC ones for now). Let me know if this looks good and I'll continue on!

### DIFF
--- a/src/buffer/channel.rs
+++ b/src/buffer/channel.rs
@@ -1,14 +1,11 @@
 use data::server::Server;
 use data::{channel, client, history, message, Config};
-use iced::{
-    color,
-    widget::{column, container, row, vertical_space},
-};
+use iced::widget::{column, container, row, vertical_space};
 use iced::{Command, Length};
 
 use super::{input_view, scroll_view, user_context};
-use crate::theme;
 use crate::widget::{selectable_text, Collection, Element};
+use crate::{message::format_message, theme};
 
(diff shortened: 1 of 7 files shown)


rniii:
I'm not sure how to set the layout bounds of a specific element, is that possible? because otherwise this happens when wrapping

image

tarkah:
> I'm not sure how to set the layout bounds of a specific element, is that possible? because otherwise this happens when wrapping
> 
> image
> 

Thanks for the PR! You've just stumbled upon why this feature hasn't been implemented yet. Without native rich text support from Iced, we can't produce a layout (line break) aware version at this level.

The good news is that rich text support isn't far off on iced! I'd suspect something will land in the next couple months.

Once we have that, it'll be easy to express formatting with the iced rich text API.

rniii:
I see, do you think we should strip out sequences for now?

tarkah:
> I see, do you think we should strip out sequences for now?

I think that'd be sane behavior until we can support those format codes natively 👍

casperstorm:
Feel free to open if still needed. Close due inactivity.

PR #215, opened 2023-11-28
[DRAFT] Fix #214: Separate user cert-chain from PKCS#8 key

### LINKED ISSUE #214: CertFP broken on macOS
Hi there,

thank you for the most awesome IRC client :-)

# Problem Statement

I've run into an issue when trying to setup SASL External on macOS.  The code in question is part of the internal IRC implementation:

https://github.com/squidowl/halloy/blob/b4dfd138ecf3e577b8436237be21c2b43a76f347/irc/src/connection.rs#L60

The function `from_pkcs8` receives two arguments, a PEM-encoded `pem` certificate-chain and the PKCS#8 (also basically PEM encoding) `key`.  The former should contain the full certificate chain from the CA down to the user/leaf certificate, the latter the key for the leaf certificate.  Using the instructions from Libera linked in the Wiki yields one PEM file combining private key and self-signed cert.

The macOS implementation is rather strict:

https://github.com/sfackler/rust-native-tls/blob/5db1b7772604d255381d847db5dd3a5b501c30a6/src/imp/security_framework.rs#L91

1. It requires `-----BEGIN PRIVATE KEY` to be the first bytes in `key` the file, otherwise it'll return `errSecParam` or "One or more parameters passed to a function were not valid."
2. It doesn't accept duplicate items, you'll get "The specified item already exists in the keychain."

# Rust Native TLS API

Especially the latter completely breaks this implementation, due to the same values being passed to `key` and `pem`.  Additionally, however, since `from_pkcs8` is used on files which do not only contain key material, this may break in other ways as well.

Before https://github.com/squidowl/halloy/commit/c6d7e665da000ce2bd34f5715056a274e63ee3c5 instead of `from_pkcs8` the function `from_pkcs12` was called.  The PKCS#12 standard indeed allows for "everything" to be combined in one file, as it embeds not only the certificate chain but also can embed a PKCS#8 key file.

In result, an "Identity" in native-tls terms can be either constructed from a

1. PKCS#12 file containing both the certificate chain and the key for the leaf, or a
2. PKCS#8 file for the key only plus a PEM file with the certificate chain.

Unfortunately, this API doesn't easily allow for the format described in the Libera API which a PEM file (thus not PKCS#12) containing all required data.

# Going Forward

I suppose the ideal solution is to:

1. Add another entry `sasl.external.key` which contains the PKCS#8 key and have `sasl.external.cert` only contain the user certificate + chain.
    * In order to stay backwards-compatible, perhaps set `key = cert` if `key` is lacking?  But this'd be permanently broken on macOS
2. Perhaps add an entry `sasl.external.combined` which can contain a PKCS#12 combined file and would conflict with `.cert` or `.key` being set.

Note that it's virtually impossible to load the format described by Libera using Native TLS portably since especially the macOS version has many sanity checkers that e.g., reject certificates which also contains keys: https://github.com/sfackler/rust-native-tls/blob/5db1b7772604d255381d847db5dd3a5b501c30a6/src/imp/security_framework.rs#L213

Ideally, Native TLS would be consistent across platforms here, either *always* discarding unused data (as done with the OpenSSL backend) or complaining if a certificate contains a key, a key contains a certificate or certs are duplicated.

ljrk0:
This adds a new config option `sasl.external.key` which is required to properly support SASL External using PKCS#8.  PKCS#8 is a format that's used exclusively for keys and not user-certs.  As such, when using `from_pkcs8()` from native-tls, both a user-cert (in PEM) and the PKCS#8 key must be provided. This was implemented before through one file which contained both the PKCS#8 key and the user-cert, concatenated. Strict implementations will either complain about the key-file containing certificates, vice-versa or duplicate certificates: since "both" the key-file and cert-file contain the same certificates due to being identical.

The code is probably not ideal, e.g.,

1. the `external_cert` and `external_key` functions could possibly be merged, or
2. the fields `client_cert_path` and `client_key_path` in `Security` be wrapped as a tuple `(client_cert_path, client_key_path)` with similar changes
3. there could be support for combined PKCS#8 key files + PEM cert chains, as well as
4. support for PKCS#12 combined identity files.

Further, the documentation in the Wiki linking to Libera may need to be updated.

### DIFF
--- a/data/src/config/server.rs
+++ b/data/src/config/server.rs
@@ -74,6 +74,7 @@ impl Server {
                 accept_invalid_certs: self.dangerously_accept_invalid_certs,
                 root_cert_path: self.root_cert_path.as_ref(),
                 client_cert_path: self.sasl.as_ref().and_then(Sasl::external_cert),
+                client_key_path: self.sasl.as_ref().and_then(Sasl::external_key),
             }
         } else {
             connection::Security::Unsecured
@@ -97,8 +98,10 @@ pub enum Sasl {
         password: String,
     },
     External {
-        /// The path to PEM encoded X509 certificate for external auth
+        /// The path to PEM encoded X509 user certificate for external auth
         cert: PathBuf,
+        /// The path to PEM encoded PKCS#8 private key corresponding to the user certificate for external auth
+        key: PathBuf,
     },
 }
 
@@ -122,8 +125,16 @@ impl Sasl {
     }
 
     fn external_cert(&self) -> Option<&PathBuf> {
-        if let Self::External { cert } = self {
+        if let Self::External { cert, .. } = self {
             Some(cert)
+        } else {
+            None
+        }
+    }
+
+    fn external_key(&self) -> Option<&PathBuf> {
+        if let Self::External { key, .. } = self {
+            Some(key)
         } else {
             None
         }
--- a/data/src/config/server.rs
+++ b/data/src/config/server.rs
@@ -101,7 +101,7 @@ pub enum Sasl {
         /// The path to PEM encoded X509 user certificate for external auth
         cert: PathBuf,
         /// The path to PEM encoded PKCS#8 private key corresponding to the user certificate for external auth
-        key: PathBuf,
+        key: Option<PathBuf>,
     },
 }
 
@@ -133,8 +133,8 @@ impl Sasl {
     }
 
     fn external_key(&self) -> Option<&PathBuf> {
-        if let Self::External { key, .. } = self {
-            Some(key)
+        if let Self::External { cert, key, .. } = self {
+            Some(key.as_ref().unwrap_or(cert))
         } else {
             None
         }
--- a/irc/src/connection.rs
+++ b/irc/src/connection.rs
@@ -26,6 +26,7 @@ pub enum Security<'a> {
         accept_invalid_certs: bool,
         root_cert_path: Option<&'a PathBuf>,
         client_cert_path: Option<&'a PathBuf>,
+        client_key_path: Option<&'a PathBuf>,
     },
 }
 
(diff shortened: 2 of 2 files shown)


casperstorm:
@tarkah could you take a look at this? 👀

data/src/config/server.rs
```diff
@@ -122,12 +125,20 @@ impl Sasl {
...
         }
     }
+
+    fn external_key(&self) -> Option<&PathBuf> {
+        if let Self::External { key, .. } = self {
+            Some(key)
+        } else {
+            None
+        }
+    }
```
tarkah, line 141: Can we make `key` an Option and return cert path here if key is none? This way we preserve backwards compatibility.

tarkah (changes requested):
Sorry for the delay on this, thank you for the PR! After reviewing things further, the change makes sense.

I'd like to see us not break backwards compatibility however. See my other comment.

data/src/config/server.rs
ljrk0: Hey, sure can do!  Just to make sure I understand correctly:

The function `external_key` is currently already returning an `Option`, so what you're suggesting is to do another fallback check like this?

```diff
- } else {
+ else if let Self::External { cert, .. } = self {
+    return Some(cert) //fallback
+ } else {
      return None
  }
```

I'm currently mostly worried about odd error cases arising from this, when people use the same config from Linux to macOS/Windows and getting weird error messages -- but I don't have any better idea.

tarkah: I've implemented this patch on top of your work and confirmed it:

- Works w/ previous config (only cert defined w/ cert+key combined file)
- Works w/ separate cert + key files / entries in config

I'm not sure how to advise people on this, other than maybe we update our wiki? At least we don't have breakage for users migrating to this newer patched version and now support separately defined cert / keys for "proper" usage, which ideally people migrate to.

ljrk0: Ah, sorry for being MIA for some time!  Sounds great!

Yeah, I was wondering whether one could add some more checking, because it'd be a weird error-case to have a config that works on Linux but you use it on macOS and it breaks (with weird TLS errors, too).  Ideally, maybe one could issue a warning for combined files, going forward?

PR #216, opened 2023-12-17
Focus Pane on Launch

andymandias:
No pane is focused on launch, even if panes are restored from a previous session.  When in this state, clicking a channel in the sidebar a opens a new pane even when ReplacePane is configured as the default action.  It is rare, in my experience, to enter into a no-pane-selected state while using Halloy, so this launch behavior is unexpected from my point of view.  This small change is to address that.

### DIFF
--- a/src/screen/dashboard.rs
+++ b/src/screen/dashboard.rs
@@ -64,7 +64,12 @@ impl Dashboard {
     pub fn restore(dashboard: data::Dashboard) -> (Self, Command<Message>) {
         let mut dashboard = Dashboard::from(dashboard);
 
-        let command = dashboard.track();
+        let command;
+        if let Some((pane, _)) = dashboard.panes.panes.iter().next() {
+            command = Command::batch(vec![dashboard.focus_pane(*pane),dashboard.track()]);
+        } else {
+            command = dashboard.track();
+        }
 
         (dashboard, command)
     }


src/screen/dashboard.rs
```diff
@@ -64,7 +64,12 @@ impl Dashboard {
     pub fn restore(dashboard: data::Dashboard) -> (Self, Command<Message>) {
         let mut dashboard = Dashboard::from(dashboard);
 
-        let command = dashboard.track();
+        let command;
+        if let Some((pane, _)) = dashboard.panes.panes.iter().next() {
+            command = Command::batch(vec![dashboard.focus_pane(*pane),dashboard.track()]);
```
anarsoul, line 69: Missing space after comma

anarsoul:
@casperstorm any comments on this?

src/screen/dashboard.rs
andymandias: New to Rust and forgot to run the modified code through rustfmt.  Fixed now.  Thank you for the review.

src/screen/dashboard.rs
```diff
@@ -64,7 +64,12 @@ impl Dashboard {
     pub fn restore(dashboard: data::Dashboard) -> (Self, Command<Message>) {
         let mut dashboard = Dashboard::from(dashboard);
 
-        let command = dashboard.track();
+        let command;
+        if let Some((pane, _)) = dashboard.panes.panes.iter().next() {
+            command = Command::batch(vec![dashboard.focus_pane(*pane), dashboard.track()]);
+        } else {
+            command = dashboard.track();
+        }
```
casperstorm, line 72: Instead of first defining `command`, you can do the following!

```rust
let command = if let Some((pane, _)) = dashboard.panes.panes.iter().next() {
    Command::batch(vec![dashboard.focus_pane(*pane), dashboard.track()])
} else {
    dashboard.track()
};
```

casperstorm:
Make sure to update CHANGELOG as well, and this looks good.

src/screen/dashboard.rs
andymandias: Thank you for the review and the pointer!

casperstorm (approved):

casperstorm:
Thanks for great work, and sorry for being slow to review.

Issue #94, opened 2023-07-04
Smart join message filter


Chemiseblanc:
It would be nice to have functionality similar to [weechat](https://blog.weechat.org/post/2008/10/25/Smart-IRC-join-part-quit-message-filter) where join/part/quit messages are only displayed if the user has spoken in a channel within some period of time.

casperstorm:
This seems like a nice third way of handling parts/join/quit.

I see it as a enum of 3 different variants a user can choose from:

1. All
2. None 
3. Smart(Time)

lodenrogue:
Can we make All the default behavior though?

casperstorm:
> Can we make All the default behavior though?

Yes.

PR #221, opened 2024-01-22
Add Smart Message Filter

andymandias:
This is an attempt to address issue #94, which works locally but may need improving.  In particular, I am concerned it will be brittle, since it relies on the formatting of the join/part/quit server messages.  But, I didn't see any other way to get the username for the server message without modifying the data structure of the server message.  If it would be preferable to modify the data structure of the server message, or if there's a better way to get the user nickname of a join/part/quit server message that I've missed, then I'd be happy to rework this implementation.

Serde was updated because I could not get it to recognize the seconds config for the smart filter otherwise.

### DIFF
--- a/data/src/history/manager.rs
+++ b/data/src/history/manager.rs
@@ -451,9 +451,9 @@ impl Data {
                         Exclude::None => true,
                         Exclude::Smart(seconds) => {
                             let nick = match source {
-                               message::source::Server::Join(nick) => nick,
-                               message::source::Server::Part(nick) => nick,
-                               message::source::Server::Quit(nick) => nick,
+                                message::source::Server::Join(nick) => nick,
+                                message::source::Server::Part(nick) => nick,
+                                message::source::Server::Quit(nick) => nick,
                             };
 
                             if nick.is_some() {
@@ -552,16 +552,16 @@ fn smart_filter_message(
     seconds: &i64,
     most_recent_message_server_time: Option<&DateTime<Utc>>,
 ) -> bool {
-    if most_recent_message_server_time.is_some() {
-        let duration_seconds = message
-            .server_time
-            .signed_duration_since(*most_recent_message_server_time.unwrap())
-            .num_seconds();
+    let Some(server_time) = most_recent_message_server_time else {
+        return true;
+    };
 
-        duration_seconds > *seconds
-    } else {
-        true
-    }
+    let duration_seconds = message
+        .server_time
+        .signed_duration_since(*server_time)
+        .num_seconds();
+
+    duration_seconds > *seconds
 }
 
 #[derive(Debug, Clone)]
--- a/data/src/history/manager.rs
+++ b/data/src/history/manager.rs
@@ -450,17 +450,11 @@ impl Data {
                         Exclude::All => false,
                         Exclude::None => true,
                         Exclude::Smart(seconds) => {
-                            let nick = match source {
-                                message::source::Server::Join(nick) => nick,
-                                message::source::Server::Part(nick) => nick,
-                                message::source::Server::Quit(nick) => nick,
-                            };
-
-                            if nick.is_some() {
+                            if let Some(nick) = source.nick() {
                                 !smart_filter_message(
                                     message,
                                     &seconds,
-                                    most_recent_messages.get(&nick.to_owned().unwrap()),
+                                    most_recent_messages.get(nick),
                                 )
                             } else if let Some(nickname) =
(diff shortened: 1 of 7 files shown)


andymandias:
I started work on a version that modifies the the Server struct in data/source.rs, but the changes were significant enough that I figure I should wait for an official ruling before going any further.  Have been using this PR's version of the smart message filter for the last couple weeks and it has been working well locally, so I think it could potentially be used in the meantime.

data/src/history/manager.rs
```diff
@@ -439,13 +439,53 @@ impl Data {
...
+                                            message::source::Server::Part => {
+                                                format!("⟵ {}", previous_message_user.nickname())
+                                            }
+                                            message::source::Server::Quit => {
+                                                format!("⟵ {}", previous_message_user.nickname())
+                                            }
+                                        };
+
+                                        if message.text.starts_with(&substring) {
+                                            break 'search_previous_messages true;
```
tarkah, line 474: I don't love that we need to match on message text here because we don't persist the user information associated to the server join / part / quit messages. 

We should revisit message::Source and see if we can add user somewhere so we have strong data to check against

data/src/history/manager.rs
```diff
@@ -439,13 +439,53 @@ impl Data {
...
             .iter()
-            .filter(|message| {
+            .enumerate()
+            .filter(|(message_index, message)| {
                 if let message::Source::Server(Some(source)) = message.target.source() {
-                    !server_messages.get(source).exclude
+                    match server_messages.get(source).exclude {
+                        Exclude::All => false,
+                        Exclude::None => true,
+                        Exclude::Smart(seconds) => 'search_previous_messages: {
```
tarkah, line 448: We've got quite a lot of nesting here, can we refactor this smart check to a function?

data/src/history/manager.rs
```diff
@@ -439,13 +439,53 @@ impl Data {
...
+                            for previous_message in
+                                messages.iter().rev().skip(messages.len() - *message_index)
+                            {
```
tarkah, line 451: Complexity grows here with the number of messages we need to iterate over.

Maybe a more stateful approach using Iter::scan to store the last message time for each user in a map would reduce that complexity. Though that has its own memory cost 🤔

data/src/history/manager.rs
andymandias: Thank you for weighing in on this.  I agree entirely.  I will plow forward with my attempt to alter message::Source then, after addressing other issues with this PR :)

data/src/history/manager.rs
tarkah: Looks good!

andymandias:
Put a warning in the commit message but it's probably prudent to do so here as well;  the latest commit stores the nickname with join/part/quit messages to check against, but that change results in chat history being clobbered when there's an error when deserializing it.  I.e. it will wipe your chat history if you run it.  Once that's fixed I think it might be good to go.

tarkah:
Awesome! We can prevent deser failure with the right serde magic. I'll take a look tomorrow 👍

andymandias:
The last couple commits have (de)serialization working for me;  no longer destroys old logs made using the old Server enum.  Added a fallback to the server message filtering so it should work for server messages deserialized from the old Server enum as well (uses the old method that attempts to get the nickname from the server message itself, by necessity).

If there's better serde magic to be employed then I'm happy to wait for that;  was just learning a bit more about serde and saw a potential fix.

data/src/history/manager.rs
```diff
@@ -507,6 +547,23 @@ impl Data {
...
+    if most_recent_message_server_time.is_some() {
+        let duration_seconds = message
+            .server_time
+            .signed_duration_since(*most_recent_message_server_time.unwrap())
+            .num_seconds();
+
+        duration_seconds > *seconds
+    } else {
+        true
+    }
```
casperstorm, line 564: Perhaps you could use let else here:
(psuedo code)

```rust
    let Some(server_time) = most_recent_message_server_time else {
        return true
    };

    let duration_seconds = message.server_time.signed_duration_since(*server_time).num_seconds();
    duration_seconds > *seconds
```

andymandias: Looks cleared to me, thanks for the suggestion.  Incorporated :)

tarkah:
@andymandias Checkout the changes I just pushed. It retains backwards compatibility, but uses some more well defined types vs serde magic. By using `untagged`, we can decode as the old value or the new value. All code paths now set the enum to the new value.

tarkah (approved):
Awesome! LGTM

PR #225, opened 2024-02-06
fix: create configuration directory correctly

a-kenji:
Correctly creates the configuration directory, if it does not exist yet.

This fixes the following confusing error:
```
thread 'main' panicked at data/src/config.rs:66:18:
expected permissions to create config folder: Os { code: 2, kind: NotFound, message: "No such fi}

```


a-kenji:
Added to the changelog and squashed the commits.

casperstorm (approved):

Issue #231, opened 2024-02-20
Feature Request: Add username color highlighting to channel user list


SamStrongTalks:
This is extending the work done in #15 to keep the colors consistent in the interface.

PR #232, opened 2024-02-20
Add nick coloring to nick_list

SamStrongTalks:
- addresses #231

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+Added:
+
+- Option to colorize nicks in the nick list (and an option to control it separately from in the buffer)
+
 Fixed:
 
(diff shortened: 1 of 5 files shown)


casperstorm:
This looks great, however it should be customizable from the config similar to how we have [user color configuration](https://github.com/squidowl/halloy/wiki/Configuration#color). I think it should be done under `buffer > channel > users > color` similar to user color mentioned before.

casperstorm:
I think we should have a seperate configuration under channel > buffer > users > color instead of using buffer > nicknname 🤔 
Otherwise it looks great.

casperstorm:
Could you also add a section to CHANGELOG with you change?

casperstorm (approved):

casperstorm:
Thanks 🥳

PR #234, opened 2024-02-24
expose scale_factor from iced

### LINKED ISSUE #233: Setting zoom / font size?
I was wondering if there's a way to set the font size or the application zoom setting somehow? I'd love to give it a go but it's way too small on my screen. Thanks!

casperstorm:
Fixes #233.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Added:
 
 - Option to colorize nicks in the nick list (and an option to control it separately from in the buffer)
+- Option to control application scale factor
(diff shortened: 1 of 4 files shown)


tarkah (approved):
LGTM! We may want to add a clamp to ensure it stays within some range, but that's not a blocker

casperstorm:
> LGTM! We may want to add a clamp to ensure it stays within some range, but that's not a blocker

Clamped between 0.1 (as 0.0 and below crashes) and 3.0.

Issue #236, opened 2024-02-25
Feature Request: Topic Banner


andymandias:
On quiet channels, or when closing/opening Halloy repeatedly for whatever reason, the channel history can build up a backlog of `RPL_TOPIC` and `RPL_TOPICWHOTIME` messages (often the same ones over and over) that swamp out the more infrequent messages that are left.  I'd like to add the option to move those messages to a banner at the top of a channel pane, showing only the latest topic (along with who it was set by and when).

I've created a possible implementation [andymandias/halloy/topic-banner](https://github.com/andymandias/halloy/tree/topic-banner).
![Screenshot from 2024-02-24 15-44-39](https://github.com/squidowl/halloy/assets/5768180/0791e664-fe9b-43e0-be2b-3817fc6e457c)

casperstorm:
I think that is a good idea, feel free to open a PR and we can work together on the design on it.

casperstorm:
Thanks for implementing this, @andymandias! 🚀 
Done in #237.

PR #237, opened 2024-02-25
Topic Banner

andymandias:
Initial implementation for #236.  Currently it builds on #221, but it can be disentangled if that complicates the process.


andymandias:
Currently the topic bar searches the channel history (new to old) to find the topic, then puts it in a scrollable at the top of the channel pane.  If no topic is found, then no scrollable is created.  Included in the topic bar is the user that set it and when it was set, with the user's nick being able to be clicked on if they are in the channel.  I made it background-color text on an info color scrollable to keep it distinct from the other views, but picked the darkest shade of the info color to try to reduce how distracting it can be.  A maximum height can be set in the config to limit the height of the topic banner, with the default aiming to be two lines tall.

casperstorm:
It's hard to review it since it builds on top of another PR. I would, if not too complicated, like to have it separate so we can work on them separately. Idea is great though, so I would like to see if we can land this feature.

andymandias:
Absolutely, sorry about that.  Wasn't sure how GitHub would handle the PR, but I can see now it's a bit of a mess.  Will disentangle then force push.

andymandias:
Disentangled :)

andymandias:
Rebased and modified to take advantage of the serde code @tarkah provided in #221.

config.yaml
```diff
@@ -86,6 +86,17 @@ buffer:
   # - Focused: Only show input when the buffer is focused
   input_visibility: Always
 
+  # Topic banner settings:
+  topic_banner:
```
casperstorm, line 90: We should just name this `topic`

src/buffer/channel.rs
```diff
@@ -108,19 +113,151 @@ pub fn view<'a>(
...
-        column![
-            vertical_space(4),
-            input_view::view(
-                &state.input_view,
-                buffer,
-                users,
-                channels,
-                input_history,
-                is_focused
+    let topic_banner = config.buffer.topic_banner.enabled.then(|| {
```
casperstorm, line 116: I feel like this part is quite complicated. We have nom inside view, and is doing quite a lot of calculations. We should tidy this up.

casperstorm (changes requested):
Screenshot 2024-03-01 at 20 06 58

Also, i played a bit with the styling and think we can tone it a bit down from what you have currently.

config.yaml
```diff
@@ -86,6 +86,17 @@ buffer:
...
+    # Show topic messages in a topic banner at the topic of a pane instead
+    # of server messages.  Topic changes will still show up as a server
+    # message.
+    # - Default is false.
+    enabled: false
```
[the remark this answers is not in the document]
andymandias: The more descriptive naming is a good idea.  Thanks & implemented :)  Wound up making the max height a struct field since I couldn't work out how to give it a default value otherwise.

config.yaml
```diff
@@ -86,6 +86,17 @@ buffer:
...
+    # Maximum height of the topic banner.
+    # - Default is 42
+    max_height: 42
```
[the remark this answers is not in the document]
andymandias: I briefly tried to have it set automatically, but I couldn't see how to get iced to give me the line height of the enclosed text (or to let me render offscreen so I can measure the line height).  I found the results where the max_height doesn't reveal an integer multiple of line height (plus padding) to be less than ideal.  I'll dig into it further.

config.yaml
andymandias: Definitely clearer.  Implemented :)

src/buffer/channel.rs
andymandias: Moved who/time line styling into its own function and removed nom for a much simpler split.  Had trouble finding documentation specifying the format for RPL_TOPICWHOTIME, but the simple split seems sufficient for our needs.

config.yaml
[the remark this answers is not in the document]
andymandias: Found `double_pass.rs` to use for calculating the rendered dimensions, such that the configuration can be in lines rather than requiring the user to know/determine the line height.  Might be too many passes to first find a window height percentage via `FillPortion` then the nearest number of lines though?

casperstorm:
@tarkah mind taking a look at this as well?

tarkah:
Let me look tomorrow, sorry for my delay on this

data/src/message/source.rs
```diff
@@ -55,6 +66,26 @@ pub mod server {
...
+
+        pub fn text(&self) -> Option<&str> {
+            match self {
+                Server::Kind(_) => None,
+                Server::Details(details) => {
+                    if let Some(text) = &details.text {
+                        Some(text.as_str())
+                    } else {
+                        None
+                    }
```
tarkah, line 78: try details.text.as_deref()

data/src/history/manager.rs
```diff
@@ -499,6 +515,62 @@ impl Data {
...
     }
 
+    fn history_banner_view(
+        &self,
+        server: &server::Server,
+        kind: &history::Kind,
+    ) -> Option<history::BannerView> {
+        let History::Full { messages, .. } = self.map.get(server)?.get(kind)? else {
+            return None;
+        };
```
tarkah, line 525: I don't think we should rely on our history to extract the topic at all. 

When we join a channel, which happens every time we start up, we should always receive RPL_TOPIC from the server. I think we can just hold a map of latest topic messages per channel in our client state and update it when those messages come in.

Then we can can query the topic from our client state vs dealing with history.

We can then remove all the metadata from message we've added here except for a single new server kind for Topic so we can filter those messages out if the banner option is chosen.

I think this will simplify a lot. Let me know if you have any questions or if you'd prefer I implemented it, but if you want to give it a shot, let me know!

tarkah (changes requested):

data/src/history/manager.rs
andymandias: Sounds like a good idea to me.  I will have to learn more about how commands/messages get passed around in iced/Halloy, but that's something I want to learn so I'd like to give it a shot.  I will likely have questions for you, but will see what I can figure out on my own first.

tarkah: Great! You can start here in data::client where we handle all messages:

https://github.com/squidowl/halloy/blob/cb6c624c77ef363df2068fa4c9002148b427dbe5/data/src/client.rs#L182

You can setup some new state on Client, update it when Topic messages come in, then expose a method to get the current topic for a server / channel and call it from view!

andymandias: This was a *very* helpful set of pointers, thank you!  I have made a first attempt at the recommended change.  I think it's quite a bit clearer/simpler and it seems to be working well here.

tarkah (approved):

data/src/history/manager.rs
tarkah: That is much better! Glad it was helpful.

I've pushed a few changes to get this across the finish line, let me know what you think.

@casperstorm LGTM from a technical standpoint, you can do a final design review

andymandias: Looks good to me.  Very instructive, thank you :)

The only "change" I have is a reversion of a change I made previously.  I had removed a `vertical_space` from before the `input_view` when switching to `spacing` because I couldn't figure out a clean way to include the same padding to the topic banner.  You found a clean way to do so for the topic banner and removed the `spacing` appropriately, but that meant there wasn't any spacing between the `input_view` and the `history_view` anymore.  So I put that back.

src/buffer/channel.rs
```diff
@@ -123,15 +123,18 @@ pub fn view<'a>(
...
-            buffer,
-            users,
-            channels,
-            input_history,
-            is_focused
-        )
-        .map(Message::InputView)]
+        column![
+            vertical_space(4),
+            input_view::view(
```
tarkah, line 128: Why are we adding this? The other buffers (Server and Query) don't add additional space here

tarkah (changes requested):

src/buffer/channel.rs
tarkah: Edit: I just saw comment https://github.com/squidowl/halloy/pull/237#discussion_r1516813482, will reply there

data/src/history/manager.rs
tarkah: Can we add that same space to the other buffers then so we're consistent?

casperstorm:
I am trying a few different designs tests, will report back tomorrow with some suggestions. Good job on this @andymandias.

data/src/history/manager.rs
tarkah: Man I'm sorry, total brain fart from me on this one.

tarkah (approved):

data/src/history/manager.rs
andymandias: Happens to the best of us :)

casperstorm (changes requested):
Screenshot 2024-03-08 101704

I've played around with styling a bunch, and I think, for now, what I like the most is actually to keep it as simple as possible, just like the `nick_list` which uses: `theme::Scrollable::Hidden` as well as `theme::Text::Transparent` to blend it more with the pane header. That way we can also remove `theme::Scrollable::Banner` and `theme::Text::Banner`. When playing around with this, I found that changing the padding slightly at `src\buffer\channel.rs:160` worked the best:

```diff
    container(body)
        .width(Length::Fill)
        .height(Length::Fill)
-       .padding(8)
+       .padding([4, 8, 8, 8])
        .into()
```



Regarding layout i think we should go for a layout like this:
```
| topic   | users |
|---------|       |
| message |       |
```
Because it can look a little strange if topic isn't long enough and user list starts below:

image

Third, and last thing, I think we should add a pane button. Notice the button on the top image. Ive tried a bunch and kinda like the following:

```rust
pub fn topic<'a>() -> Text<'a> {
    to_text('\u{f5af}')
}
```
I think, with a button like this we can simplify the config a little as well and make it similar to Users:

```yaml
buffer:
    users:
      visible: true
```

Perhaps we could simply do:
```yaml
buffer:
    topic:
      visible: false
      max_lines: 2
```

Functionality should just be 1 to 1 with showing and hiding `nick_list`.
Let me know what you think about these suggetions, @andymandias.

andymandias:
Thank you for the notes @casperstorm.

I think these are all good changes, and mesh into the Halloy style better.

Re: padding.  By removing the background color from the topic banner we can shrink the padding on it without any aesthetic changes.  Over here there's been a downside to the topic banner's current padding where it would show a sliver of the third line when only two lines were otherwise showing.  For example with the currently-committed padding I see:

Screenshot from 2024-03-08 14-32-56

So, unless there's an issue I'm going to remove the vertical padding from the topic banner.  I believe that will be equivalent to the padding change you've suggested in `channel.rs`, so I'm going to hold off on making that change.  I have no objection to changing the padding further, I just don't want to overshoot what you're suggesting.

I am slightly concerned that there's not a lot differentiating the topic from the rest of the channel text.  Would using something like a light horizontal rule fit within the Halloy style?

Screenshot from 2024-03-08 14-23-22

A seasoned user probably wouldn't need it, but a newbie might be able to use the extra affordance.

I like the pane button idea quite a bit.  I'm assuming toggling the topic button should show/hide the inline topic messages accordingly.  Is that right?  Or would it only toggle the visibility of the banner?

I think matching configuration options is ideal, but I am concerned it could be confusing to users if they set `visible: false` and they see inline topic messages.  Perhaps the options could be Inline, Banner, or false?

casperstorm:
1. A divider between topic and content is all good to me. It looks good as you have it.
2. Re the sliver of the third line. Could a solution be adding a border and or background in the same color as the pane so it looks transparent but is colored? Just an idea, haven't tested if it actually works.
3. Re button. My initial thought was that config setting only affected banner, and button was a toggle for it. Similar to user list where you can set `visible: false`, but still quickly toggle to briefly check. What we could do is consider adding a `topic` variant to [server_messages](https://github.com/squidowl/halloy/wiki/Configuration#server-messages) which controls inline, and then rename `topic` under buffer to `topic_banner` with `visible: boolean, max_lines: u16`? That could solve all cases perhaps.
4. Re padding. I think we should aim to get the spacing between pane header, topic divider and from topic divider to content equal. See below image. Its a super small nit. 
image

andymandias:
Thank you for the clarifications!  I'll make the changes, going with both banner and server messages configuration options as you describe.  No sweat if we wind up changing the configuration options again :)

andymandias:
Some chores left to be done (e.g. example config update, code clean-up pass, etc), but I believe this has the new configuration and functionality we're looking for.  I struggled a little with the under-topic padding since its apparent height can be influenced by scroll state of the messages in the channel;  if I didn't get it right please feel free to fix (or let me know how far off I am).

casperstorm (approved):
I updated padding slightly, example config file, CHANGELOG and then rebased on main. We should update wiki once we merge.

I don't have anything else - this looks awesome, and what awesome feature. Do you have anything you want to clean up before we merge @andymandias?

andymandias:
Did a quick review and didn't see anything I made too messy.  Looks good to merge to me.

Thanks for taking care of the rest of the chores @casperstorm, and for all the guidance from you and @tarkah.

PR #239, opened 2024-02-28
Minor Username-Click → Input Tweaks

andymandias:
Two minor QoL changes to how clicking on a username adds the username to the input.  These are very minor and slightly related, so I'm grouping them together to hopefully be less busywork.

1.  Remove the space after the username from the click target.  (The only reason I noticed this is because that space is where I aim when starting a selection for an entire message.)

2.  If clicking on a username and there is already text in the input, then check whether the input ends in a space;  if it doesn't end in a space then insert a space followed by the username (current behavior), otherwise insert just the username (no space).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Fixed:
 
 - Input not visible on Server and Query (DM) buffers
 
+Changed:
+- Improved user experience in text input when auto-completing a nickname.
+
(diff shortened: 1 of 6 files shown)


casperstorm:
Good changes, thanks @andymandias.
Do you feel like it deserves a CHANGELOG update?

casperstorm (approved):

PR #242, opened 2024-02-28
Changed default value for ServerMessage and updated changelog

casperstorm:
* I changed the default value to `None` so if a user doesn't setup anything they see server messages, I think that makes the most sense.
* Updated changelog.

@andymandias you happy with this?


andymandias:
Agreed re: default value.  Looks good to me :)

PR #243, opened 2024-02-29
Remove Message Wrapping Offset Introduced by PR #239

andymandias:
With the commits added in #239 I forgot to account for each `selectable_text` being used for text wrapping.  So when wrapping a message, the first line is indented by a space but none of the following lines are indented.  This PR removes the offset by moving the space between the username and message to its own `selectable_text`.  All the style changes are from `rustfmt`.


casperstorm (approved):
Thanks @andymandias!

PR #254, opened 2024-03-06
feat: use height of text as button height in context menu

casperstorm:


casperstorm:
@andymandias do you mind testing this if context menus works on linux?

andymandias:
Confirmed on IRC but will also confirm here for posterity, it works perfectly.  Thank you!

data/src/config.rs
```diff
@@ -58,10 +58,40 @@ impl From<ScaleFactor> for f64 {
...
+        }
+
+        let MaybeFont { family, size } = MaybeFont::deserialize(deserializer)?;
+        let Font { family: default_family, size: default_size } = Font::default();
+
+        Ok(Font {
+            family: family.unwrap_or(default_family),
+            size: size.unwrap_or(default_size),
+        })
+    }
```
tarkah, line 94: `#[serde(default_with = "default_font_size")]`

```rs
fn default_font_size() -> f32 {
  13.0
}
```

tarkah: Then you can keep `default` below on Font

src/screen/dashboard/side_menu.rs
```diff
@@ -238,9 +246,12 @@ fn buffer_button<'a>(
                 ),
             };
 
+            // Height based on font size, with some margin added.
+            let height = LineHeight::default().to_absolute((font.size + 8.0).into());
+
             button(text(content).style(theme::Text::Primary))
                 .width(length)
-                .height(length)
+                .height(height)
```
tarkah, line 254: This seems super hacky. Let me play around. We should just be able to use `Shrink` height and have it work when font size increases....

PR #255, opened 2024-03-07
Fix float precision issue in double pass

tarkah:
If second pass is fill and we provide the fixed
size from the first pass, it's possible fill doesn't return the same layout within those limits due to
flex / float precision. Expanding the limits by 1
pixel appears to completely resolve this.


andymandias:
Looks good, all context menus render as intended.

casperstorm (approved):

PR #262, opened 2024-03-12
Add "Drafts" to Input

andymandias:
This PR is to automatically store the contents of the input view, such that non-sent messages will persist when switching between channels/queries/servers.  I.e. if I type out a part of a message for a channel, switch to another channel to reference other messages, then when I switch back to the original channel the part of the message that I typed will be restored.  Previously, when switching channels the input is cleared and anything that has been typed is lost.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 Added:
 
 - Configuration option to enable a topic banner in channels. This can be enabled under `buffer.channel.topic`
-- Messages typed into the input will persist until sent.  Typed messages are saved when switching a pane to another buffer, then
+- Messages typed into the input will persist until sent. Typed messages are saved when switching a pane to another buffer, then
   are restored when that buffer is returned to.
 
 Fix:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 Added:
 
 - Configuration option to enable a topic banner in channels. This can be enabled under `buffer.channel.topic`
+- Messages typed into the input will persist until sent.  Typed messages are saved when switching a pane to another buffer, then
+  are restored when that buffer is returned to.
 
 Fix:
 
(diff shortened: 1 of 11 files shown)


src/screen/dashboard.rs
```diff
@@ -783,6 +785,15 @@ impl Dashboard {
         self.panes.get_mut(pane).map(|state| (pane, state))
     }
 
+    fn get_focused_mut_with_history(
```
casperstorm, line 788: Nit: By default we normally name mutable functions `_mut` at the end.
```rust
fn get_focused_with_history_mut
```

casperstorm:
This looks and feels good, similar to how Discord, Slack and other applications behave.
Didn't really pay too much attention to this before, because I normally run in a split plane without but I welcome this change.
Could you add a entry to CHANGELOG?

src/screen/dashboard.rs
andymandias: Makes sense, thanks for the note :)

andymandias:
Oops, sorry about that.  CHANGELOG entry added.

Happy to provide the single-pane-user perspective.

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 Added:
 
 - Configuration option to enable a topic banner in channels. This can be enabled under `buffer.channel.topic`
+- Messages typed into the input will persist until sent.  Typed messages are saved when switching a pane to another buffer, then
```
casperstorm, line 6: Nit: Double space.

casperstorm (approved):
I'll approve, and wait to see if @tarkah has any additional comments.

tarkah (approved):
Awesome as always @andymandias! I've done a little cleanup, mainly removing the unneeded state on `input_view::State` since we can just update / reference the draft value from our stored history manager -> input widget.

Issue #226, opened 2024-02-06
Many Server Connection Reports


andymandias:
Sometimes when waking my laptop from sleep I get many `connection to server lost` and `connection to server restored` messages in a row all with the same timestamp (example screenshot attached).  I'm unable to reliably reproduce the behavior, and I currently don't have any reason to believe that they're false reports, but they do take up a big block in the backlog.  I don't believe we want to suppress that information entirely, but I think it would be nice if the messages could be condensed into a single pair of `connection to server lost` and `connection to server restored` messages.
   
![Screenshot from 2024-02-05 14-27-45](https://github.com/squidowl/halloy/assets/5768180/2d67b4fb-e312-4ca8-9efd-6128bc79432d)

andymandias:
Looking into this a bit more, these (dis)connections are all happening within a tenth of a second.  It seems like these shouldn't happen since I haven't changed the reconnect delay from the default 10s.  I've written a simple message filter to coalesce the messages into single a disconnect/connect pair, but maybe it'd be better to stop them at the source?

casperstorm:
I wonder if @tarkah has any thoughts about this.

tarkah:
Make sure to set reconnect delay https://github.com/squidowl/halloy/blob/main/data%2Fsrc%2Fconfig%2Fserver.rs#L39-L41 in your server config

tarkah:
Relevant code is here, we always try to delay before reconnecting. All branches which disconnect set last retry to now. So either reconnect delay is somehow zero or something is weird with Instant and sleep.

https://github.com/squidowl/halloy/blob/main/data%2Fsrc%2Fstream.rs#L85-L92

andymandias:
Thank you for your thoughts and the code reference.  I've added some diagnostic prints around reconnection to the debug log and will report back with details the next time it occurs.

andymandias:
The delay is happening between `data/src/stream.rs` and `src/main.rs` (as expected, but best to confirm in my experience).  I don't know why the wait time between send and receive is sometimes very long (~1.5 hours the most recent time this issue occurred), but I expect it's out of the scope of Halloy.  I've drafted a PR which I believe will give users accurate reconnect/disconnect times;  just waiting for the issue to occur again to test abnormal behavior (normal behavior appears to be effectively unchanged, as desired).  I think we can close this issue with PR #270 (or something like it).

andymandias:
[...]

PR #270, opened 2024-03-14
Add Send Time to Broadcast Messages

andymandias:
This is an attempt to address the Halloy side of #226 by adding a `sent_time` to those broadcast messages.  Other broadcast messages I expect will not have such long times between send and receive, so I have not added `sent_time` to those messages, instead it defaults to `Utc::now()` at receive time (previous behavior).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Fixed:
 
 - Accept '@' in usernames to support bouncers that use the user@identifier/network convention
+- Prevent rare scenario where broadcast messages' timestamp would not match time the messages are received
 
 # 2024.5 (2024-03-21)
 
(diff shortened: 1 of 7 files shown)


andymandias:
Shows the correct connect/disconnect times in the backlog for me.

casperstorm (commented):
This makes sense to me, however is there a good way to test it?

andymandias:
Okay, I can now consistently reproduce broadcast timestamps not matching when the event occurred with these steps (these may not be the minimum required yet, but they are consistently working for me):
1.  Sleep laptop with Halloy in another workspace.
2. Wake laptop, but do not return to the workspace with Halloy.
3.  On another computer enter the same channel and perform a broadcast action (e.g. use the webchat to log in to the channel and change nickname).
4.  Return to Halloy workspace.

tarkah (approved):
LGTM. Makes sense we would show server time on broadcast messages for better accuracy just like other messages, since they're based on a server message.

casperstorm (approved):

PR #273, opened 2024-03-15
More Linux Icons

andymandias:
In my experience the scalers in Linux DEs can use as much help as Windows/macOS.  I picked the icon sizes based on the sizes inside `assets/windows/halloy.ico` and the conversion options are unchanged in `scripts/generate-icons.sh`;  works well over here, but I'm not attached to any of the details.  The directory structure is a bit of a beast, but is based on the [XDG directory layout specification](https://specifications.freedesktop.org/icon-theme-spec/latest/ar01s03.html) so that most users who build Halloy themselves could e.g. `cp -r assets/linux/icons ~/.local/share/` to install all the icons in one go.


tarkah:
LOL @andymandias we were working on this at the same time:

https://github.com/squidowl/halloy/pull/274

casperstorm:
Hahahaha

PR #277, opened 2024-03-15
Fix `Solid` nickname color

neilalexander:
This was accidentally changed to the action color in #232.

### DIFF
--- a/src/theme/text.rs
+++ b/src/theme/text.rs
@@ -1,6 +1,7 @@
-use super::Theme;
 use data::theme::{alpha, randomize_color};
 use iced::widget::text::{Appearance, DefaultStyle};
+
+use super::Theme;
 
 impl DefaultStyle for Theme {
     fn default_style(&self) -> Appearance {
@@ -49,6 +50,17 @@ pub fn transparent(theme: &Theme) -> Appearance {
 }
 
 pub fn nickname(theme: &Theme, seed: Option<String>, transparent: bool) -> Appearance {
+    let dark_theme = theme.colors().is_dark_theme();
+
+    if seed.is_none() {
+        let color = match transparent {
+            true => theme.colors().text.med_alpha,
+            false => theme.colors().text.base,
+        };
+
+        return Appearance { color: Some(color) };
+    }
+
     let original_color = theme.colors().action.base;
     let randomized_color = seed
         .as_deref()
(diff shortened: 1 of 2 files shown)


casperstorm:
Hmm, this was because the nicknames in chat was `action` if solid was selected, but nicknames in sidemenu was `text`. I think we should consider `accent` for both instead:

Screenshot 2024-03-17 at 06 42 17

neilalexander:
This is probably one of those subjective/theme-specific things — to my eyes, the "blending together" was what made the nick list disappear into the periphery rather than stand out too much. For example, this stands out quite a lot:

Screenshot 2024-03-17 at 10 53 39

... whereas this felt more in-keeping to me with the rest of the theme:

Screenshot 2024-03-17 at 10 54 37

Accent colour might be OK, it doesn't look as though it's used by much else apart from text selection?

Another alternative is that we put it to `text` and add some permanent alpha to it, so that it's subtly different but not by much, for example at alpha 0.6:

Screenshot 2024-03-17 at 10 57 21

casperstorm:
Colors are hard :)
Perhaps the solution is simply to do `text.base` for both nicklist and nicknames in buffer.
It will blend the most, which i assume people who use this option want.

neilalexander:
Right now there are separate `colors:` options for the buffer and nick list, so they can be controlled separately. That seems like a nice option to keep? Then you can have colours in the buffer but not nick list (like I do), both or neither.

Presumably both use the `nickname()` function?

casperstorm:
> Right now there are separate `colors:` options for the buffer and nick list, so they can be controlled separately. That seems like a nice option to keep? Then you can have colours in the buffer but not nick list (like I do), both or neither.
> 
> Presumably both use the `nickname()` function?

Correct, but initially the nickname color in the buffer was `action` and the colors in nicklist was `text`.
Then in `main` both is `action`, and this PR changes both to `text`. 😄 

I am good with that, now that I've given it some thoughts.

src/theme/text.rs
```diff
@@ -49,14 +50,23 @@ pub fn transparent(theme: &Theme) -> Appearance {
...
 pub fn nickname(theme: &Theme, seed: Option<String>, transparent: bool) -> Appearance {
+    let dark_theme = theme.colors().is_dark_theme();
+
+    if seed.is_none() {
+        let mut color = theme.colors().text.base;
+        if transparent {
+            color = alpha(color, if dark_theme { 0.2 } else { 0.4 })
+        }
+        return Appearance { color: Some(color) };
+    }
```
casperstorm, line 61: What about:

```rust
    if seed.is_none() {
        let color = match transparent {
            true => theme.colors().text.med_alpha,
            false => theme.colors().text.base,
        };

        return Appearance { color: Some(color) };
    }
```

neilalexander: That's much nicer and more compact, will do that!

casperstorm:
Could you also add a small line to CHANGELOG? Then it should be good to go.

neilalexander:
Done!

casperstorm (approved):
Thanks a bunch @neilalexander

PR #278, opened 2024-03-15
feat: yaml to toml

casperstorm:
- [ ] Change from yaml to toml
- [ ] Migrate entire github wiki to halloy wiki
- [ ] Migration page on halloy wiki
- [ ] Migrate all user themes to halloy wiki
- [ ] Changelog


.github/workflows/publish-book.yml
```diff
@@ -0,0 +1,25 @@
+name: Publish
+on:
```
tarkah, line 2: Sorry but I find it hilarious that the first file to review on this PR is a `yaml` file :laughing:

data/src/config.rs
```diff
@@ -1,29 +1,29 @@
...
 pub mod channel;
-pub mod dashboard;
 mod keys;
 pub mod notification;
 pub mod server;
+pub mod sidebar;
 
-const CONFIG_TEMPLATE: &[u8] = include_bytes!("../../config.yaml");
-const DEFAULT_THEME_FILE_NAME: &str = "ferra.yaml";
+const CONFIG_TEMPLATE: &[u8] = include_bytes!("../../config.toml");
```
tarkah, line 25: Use `include_str!`  and then you wont need to `String::from_utf8_lossy` below

data/src/config/buffer.rs
```diff
@@ -52,15 +44,29 @@ impl ServerMessages {
...
+#[derive(Debug, Copy, Clone, Deserialize)]
+#[serde(rename_all = "kebab-case")]
 pub struct ServerMessage {
+    #[serde(default = "default_bool_true")]
+    pub enabled: bool,
     #[serde(default)]
-    pub exclude: Exclude,
+    pub smart: Option<i64>,
     #[serde(default)]
     pub username_format: UsernameFormat,
```
tarkah, line 55: `username-format` should be `username_format` since we're going default rust field / snake_case?

data/src/buffer.rs
```diff
@@ -60,19 +60,36 @@ impl From<config::Buffer> for Settings {
 }
 
 #[derive(Debug, Clone, Copy, Default, Deserialize)]
-pub enum InputVisibility {
+#[serde(rename_all = "kebab-case")]
+pub struct TextInput {
```
tarkah, line 64: Remove `kebab-case` since we're going snake_case. We only want `kebab-case` on enum variants

PR #281, opened 2024-03-16
Make passwords readable from file

funkeleinhorn:
It would be nice to be able to put passwords in extra files with secure permissions instead of including them in plaintext in the config as some users might share the config via version control etc. and this would leak passwords.

### DIFF
--- a/data/src/config/server.rs
+++ b/data/src/config/server.rs
@@ -11,6 +11,8 @@ pub struct Server {
     pub nickname: String,
     /// The client's NICKSERV password.
     pub nick_password: Option<String>,
+    /// The client's NICKSERV password file.
+    pub nick_password_file: Option<String>,
     /// Alternative nicknames for the client, if the default is taken.
     #[serde(default)]
     pub alt_nicks: Vec<String>,
@@ -25,6 +27,8 @@ pub struct Server {
     pub port: u16,
     /// The password to connect to the server.
     pub password: Option<String>,
+    /// The file with the password to connect to the server.
+    pub password_file: Option<String>,
     /// A list of channels to join on connection.
     #[serde(default)]
     pub channels: Vec<String>,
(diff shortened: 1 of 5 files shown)


data/src/config/server.rs
```diff
@@ -96,6 +100,8 @@ pub enum Sasl {
         username: String,
         /// Account password,
         password: String,
```
casperstorm, line 102: Shouldn't this be changed to `Optional<String>` so we only need to fill one of them when using Plain Sasl?

casperstorm (changes requested):
Rebased to master. Could you also add entry to CHANGELOG and update /book/src/configuration/server.md with updated argument.

data/src/client.rs
```diff
@@ -108,7 +109,14 @@ impl Client {
             let user = config.username.as_ref().unwrap_or(nick);
             let real = config.realname.as_ref().unwrap_or(nick);
 
-            if let Some(pass) = config.password.as_ref() {
+            if let Some(pass_file) = config.password_file.as_ref() {
+                match fs::read_to_string(pass_file) {
+                    Ok(pass) => {
```
tarkah, line 114: We shouldn't block here as this is all run on the main UI thread. Instead, we should load the password into memory at our entry point in main and pass it in. This can be done right after we load the config file.

funkeleinhorn: Moved the file reading to Config::load();

data/src/config/server.rs
funkeleinhorn: Made this Optional and check at runtime that one of password or password_file is set.

funkeleinhorn:
I addressed the requested changes. @casperstorm @tarkah please have a look again :)

casperstorm:
> I addressed the requested changes. @casperstorm @tarkah please have a look again :)

Looking at this today!

tarkah (approved):

tarkah:
Great call putting it in that load path, looks great!

casperstorm (approved):

PR #282, opened 2024-03-16
User Context Tweaks

andymandias:
Two minor tweaks related to the new user context functionality:

1.  Only show the operator actions when the user is an operator.  
2.  The context menu for usernames in the history view would always show the user with default access levels.  As far as I can tell, that's because the user from `message::Source::User(user)` is being constructed from proto::User, which always has the default access levels.  I didn't see a simple way to add the user lookup to `message::Source::User` construction, so I added it to the channel view.

I can split these up if one is desirable and the other is not.

### DIFF
--- a/src/buffer/channel.rs
+++ b/src/buffer/channel.rs
@@ -1,4 +1,5 @@
 use data::server::Server;
+use data::user::Nick;
 use data::User;
 use data::{channel, history, message, Config};
 use iced::widget::{column, container, row};
@@ -34,10 +35,9 @@ pub fn view<'a>(
     let input = history.input(&buffer);
     let our_nick = clients.nickname(&state.server);
 
-    let users = clients.get_channel_users(&state.server, &state.channel);
-    let our_user = users
-        .iter()
-        .find(|user| user.nickname() == our_nick.unwrap());
+    let our_user = our_nick
+        .map(|our_nick| User::from(Nick::from(our_nick.as_ref())))
+        .and_then(|user| clients.resolve_user_attributes(&state.server, &state.channel, &user));
 
     let messages = container(
         scroll_view::view(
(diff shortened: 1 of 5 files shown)


tarkah:
The message::Source::User should be resolved with attributes. We do this in data::message when constructing it.

The reason you might see attributes missing is because this uses the attributes from our client state at the time we receive the message. If you first load Halloy and get a backlog of messages from your bouncer, these probably come in before we resolve WHO messages to update client state with attributes.

@casperstorm and I are aware of this potential gap. The upside of doing it this way however is old messages more accurately reflect permissions at the time the message was sent (ignoring the initial backlog timing issue). 

We need to align on if we want to address this and maybe use some type safety to ensure we're always working with the desired user type.

andymandias:
Ahh, my apologies you're absolutely correct.  I was testing with messages sent before attribute levels were changed.  And, as you mention, the way I implemented this will reflect an inaccurate message history wrt attributes.  Maybe the op context menu items could be only present in the nick_list?  Regardless, I'll remove that part of this PR.

src/buffer/channel.rs
```diff
@@ -34,6 +34,11 @@ pub fn view<'a>(
     let input = history.input(&buffer);
     let our_nick = clients.nickname(&state.server);
 
+    let users = clients.get_channel_users(&state.server, &state.channel);
+    let our_user = users
+        .iter()
+        .find(|user| user.nickname() == our_nick.unwrap());
```
tarkah, line 40: We have a clients.resolve_attributes you can use

andymandias: Thanks for the heads up!

casperstorm:
Can you add a entry to CHANGELOG as well?

andymandias:
I promise one of these days I'll remember to update changelog :sweat_smile:

casperstorm (approved):

casperstorm:
Rebased on master. Tested and works great.

Issue #279, opened 2024-03-16
Icons are not included in the 2024.4 release


txtsd:
```
λ tree src/
src/
├── halloy
├── org.squidowl.halloy.appdata.xml
└── org.squidowl.halloy.desktop
```

The previous release had the single icon file, but this one doesn't have anything. Please do a release with the `icons` folder included.

PR #283, opened 2024-03-17
Fix for #279

andymandias:
Feel free to just close this if there's already something in the works for this.


casperstorm (approved):
@tarkah can you confirm this as well?

tarkah (approved):
Thanks @andymandias. I've gone ahead and updated to use the proper installation paths for linux, so now these paths can be simply installed to `PREFIX` on the users system.

andymandias:
Technically `~/.icons` is a valid Freedesktop path, but nobody should be encouraged to use it.  Looks good to me.

PR #294, opened 2024-03-23
added who poll interval as server config

### LINKED ISSUE #293: WHO being rate-limited
I repeatedly get the following message in my Halloy server console:
```
* End of /WHO list.
WHO This command could not be completed because it has been used recently, and is rate-limited.
```

Why does Halloy need to poll WHO so often? It's slowing down my connection.

casperstorm:
Fixes #293

### DIFF
--- a/data/src/config/server.rs
+++ b/data/src/config/server.rs
@@ -1,8 +1,9 @@
 use std::collections::HashMap;
 use std::path::PathBuf;
+use std::time::Duration;
 
 use irc::connection;
-use serde::Deserialize;
+use serde::{Deserialize, Deserializer};
 
 #[derive(Debug, Clone, Default, Deserialize)]
 pub struct Server {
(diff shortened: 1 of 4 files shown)


data/src/config/server.rs
```diff
@@ -141,6 +148,56 @@ impl Sasl {
...
+#[derive(Debug, Clone)]
+pub struct WhoPollInterval(Duration);
+
+impl WhoPollInterval {
+    pub fn duration(&self) -> Duration {
+        self.0
+    }
+}
+
+impl Default for WhoPollInterval {
```
tarkah, line 160: Instead of newtyping both to implement Deserialize, you can tag the field with #[serde(deserialize_with = "")] and just define the single deser function to apply to both fields

data/src/config/server.rs
```diff
@@ -65,6 +66,18 @@ pub struct Server {
...
+        deserialize_with = "deserialize_duration_from_u64"
+    )]
+    pub who_poll_interval: Duration,
+    /// WHO retry interval for servers without away-notify.
+    #[serde(
+        default = "default_who_retry_interval",
+        deserialize_with = "deserialize_duration_from_u64"
+    )]
+    pub who_retry_interval: Duration,
 }
```
tarkah, line 81: Make sure we implement Default for this struct vs derive so the durations can be defaulted properly when this entire struct is omitted / defaulted

tarkah (approved):

PR #295, opened 2024-03-24
Accept ZNC Client Identifier Format Usernames

andymandias:
ZNC allows usernames of the format `username@identifier/network` where the identifier tells ZNC which client is accessing the bouncer.  From what I can tell adding `@` to the list of special characters accepted for nicknames seems to make that format work with Halloy, with the expected `!` for the username protecting us from accidentally adding the hostname to the nickname.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+Fixed:
+
+- Accept '@' in usernames to support bouncers that use the user@identifier/network convention
+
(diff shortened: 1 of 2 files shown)


casperstorm:
Could you add a test for it as well? And update changelog 😅

andymandias:
Apologies!  Added 😅

casperstorm (approved):

tarkah (approved):
LGTM, `!` protects us as you mention

PR #302, opened 2024-03-27
Sign macOS builds

petergam:
Adds signing and Apple notarization for macOS release builds


scripts/build-macos.sh
```diff
@@ -36,10 +36,3 @@ cp -fRp "$APP_TEMPLATE" "$APP_DIR"
 cp -fp "$APP_BINARY" "$APP_BINARY_DIR"
 touch -r "$APP_BINARY" "$APP_DIR/$APP_NAME"
 echo "Created '$APP_NAME' in '$APP_DIR'"
-
-# package dmg
-echo "Packing disk image..."
-ln -sf /Applications "$DMG_DIR/Applications"
-hdiutil create "$DMG_DIR/$DMG_NAME" -volname "Halloy" -fs HFS+ -srcfolder "$APP_DIR" -ov -format UDZO
-echo "Packed '$APP_NAME' in '$APP_DIR'"
```
petergam, line 44: @casperstorm I moved the package stuff into its own script that can be called after signing. Not sure if this will cause issues (for instance if the `build-macos.sh` script is using in other places)

casperstorm: It's not, so should be fine

casperstorm (approved):
This is great!!

PR #304, opened 2024-03-28
feat: display version in command bar

casperstorm:
Display version, as well as latest remote version (if its different from current version).
Note: fixed a few lints in CHANGELOG.

Screenshot 2024-03-28 at 19 43 41


tarkah (approved):
LGTM, nifty!

Issue #296, opened 2024-03-24
nick_password gets treated as nickname


JuliusBairaktaris:
Hi,

I am trying to connect to #openwrt-devel on oftc.net but I get an error when I try to log in

```
20:07 NickServ This nickname is registered and protected.  If it is your nickname, you may
20:07 NickServ authenticate yourself to services with the IDENTIFY command.  You are
20:07 NickServ getting this message because you are not on the access list for the
20:07 NickServ JuliusBairaktaris nickname.
20:07 NickServ Nickname <password> is not registered.  The nickname you specify must be a
20:07 NickServ registered nickname.  See HELP REGISTER for more information.
```

I have configured halloy as follows:

```
[servers.oftc]
nickname = "JuliusBairaktaris"
nick_password = "<password>"
server = "irc.oftc.net"
realname =  "Julius Bairaktaris"
port = 6697
use_tls = true
channels = ["#openwrt", "#openwrt-devel"]
```

The password is correct. I double-checked it. (Replaced in this issue for obvious reasons)

YouFoundAlpha:
Hello @JuliusBairaktaris 
You should use `on-connect` option instead of nick_password:
https://halloy.squidowl.org/configuration/servers.html
for example like this:
on-connect: ["/msg NickServ IDENTIFY foo bar"]

PR #308, opened 2024-04-01
Choose IDENTIFY Syntax Based on Resolved Nick or Server

andymandias:
This is to address #296.  As far as I was able to find, there doesn't seem to be an official standard for NickServ.  From a quick search, the more common syntax seems to be `IDENTIFY nickname password` (Atheme, Anope, Libera, Freenode), while OFTC uses `IDENTIFY password nickname`.  All servers (that I found) accept the `IDENTIFY password` syntax when identifying as the current nickname.

So, the added logic aims to use `IDENTIFY password` whenever possible, then uses the syntax for servers that are known, then finally defaults to the more common `IDENTIFY nickname password` syntax.

### DIFF
--- a/data/src/config/server.rs
+++ b/data/src/config/server.rs
@@ -13,6 +13,8 @@ pub struct Server {
     pub nick_password: Option<String>,
     /// The client's NICKSERV password file.
     pub nick_password_file: Option<String>,
+    /// The server's NICKSERV IDENTIFY syntax.
+    pub nick_identify_syntax: Option<IdentifySyntax>,
     /// Alternative nicknames for the client, if the default is taken.
     #[serde(default)]
     pub alt_nicks: Vec<String>,
@@ -103,6 +105,13 @@ impl Server {
             security,
         }
     }
+}
+
+#[derive(Debug, Clone, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub enum IdentifySyntax {
+    NickThenPassword,
+    PasswordThenNick,
 }
 
 #[derive(Debug, Clone, Deserialize)]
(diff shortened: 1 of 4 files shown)


andymandias:
Sounds good to me.  I implemented a slightly different version because I misremembered your suggestion.  I can align it with your suggestion, but just to be sure:  we want to avoid the nick-less syntax entirely?  It has the highest coverage from what I saw (everything I looked into supported it), so it would probably provide the most seamless out of the box experience for users;  on the other hand, it's a bit more complicated and might confuse users.  At the moment it's set up only to be used if a specific syntax has not been configured.

data/src/config/server.rs
```diff
@@ -105,6 +107,13 @@ impl Server {
     }
 }
 
+#[derive(Debug, Clone, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub enum IdentifySyntax {
+    NickThenPassword,
+    PasswordThenNick,
+}
```
tarkah, line 115: Nit: maybe just `NickPassword` / `nick-password`, etc works since it's kebab case?

tarkah (approved):
LGTM, one small nit.

I was talking to @casperstorm today about his comment above, and I agree we should limit how much we expose here since users can generally just use the on_connect commands to do whatever they want.

However, this specifically uses a password which may be sourced from a file, so it's not really doable here unless we add some tempting to on_connect to replace some variables with in-memory values. 

So maybe something we move to in the future. In the meantime, this looks great!

andymandias:
Makes sense to me.  Thanks for the reviews!

data/src/config/server.rs
tarkah: Thanks :D Don't forget to update the docs!

andymandias: Shoot, thanks for the reminder!

tarkah (approved):

tarkah:
Thanks!

Issue #287, opened 2024-03-20
Feature Request: Option to Indicate Absent Users in Channel


andymandias:
I think it would be nice if there was an option to change their nickname appearance of absent users in the the channel history, such as

![Screenshot from 2024-03-19 21-26-26](https://github.com/squidowl/halloy/assets/5768180/e24abaea-5be4-4609-8516-4d6642c70222)

where nicknames for users that are not present appear as `theme.colors().info.low_alpha` (not particular to the color, just wanted it to be distinguished from the away state coloring while mimicking away's transparent cue).  The aim of this feature, from my perspective, is to give an option to further suppress server messages by providing another mechanism to quickly see whether a user is currently present.

casperstorm:
What is the difference between absent users and away users?

andymandias:
My (maybe incorrect understanding) was that away users are still in the channel but have away set (the IRC command/status), while what I'm calling absent users would be ones that are no longer in the channel.  My assumption is that users that are still in the channel but away could be waiting for a reply, while users that are no longer on the channel probably won't see a reply.  Seeing that status right next to the message (rather than searching through the nick list) would be handy in my use, but I'm not sure if it's has wide enough appeal.

andymandias:
To avoid messing with the historical accuracy of the chat logs, how about this instead: a hover state for usernames which shows a small pane with the user's current status (present, away, absent, and access levels)?

andymandias:
Got delayed by pondering the Rust heap, but I had some time to put together a mock up of the tooltip version of this feature.  Pictured are absent/present state tooltips.  There is also an away state that looks similar to the present state, just with an info colored dot (and it says Away instead of Present);  couldn't find a user to test with before my dev time is up.

![absent_tooltip](https://github.com/squidowl/halloy/assets/5768180/debd6e9d-99fb-4f9a-b401-12d731ffc141)
![present_tooltip](https://github.com/squidowl/halloy/assets/5768180/73796a29-8f45-44d1-b604-24df268b8a44)

casperstorm:
> absent users would be ones that are no longer in the channel

I am still a bit unsure about this. I feel like irc users operates on two states; away and not-away.
However, a tooltip hovering users (in nicklist) is a fantastic idea with useful information about them.

andymandias:
[...]

PR #316, opened 2024-04-03
Username Tooltips

andymandias:
Implementation for the feature discussed in #287.  Feature intends to be exactly as was discussed, but when adding it to the changelog I noticed there is a configuration option for application button tooltips.  So I added configuration option to the tooltips section;  made the option separate from button tooltips, since it I imagine users might want one without the other.


casperstorm:
@andymandias,
I've played with this for a while now and I am not 100% sold on it after using it in practise.
I think it becomes a little confusing when moving the mouse around and this tooltip pops up left and right.
Perhaps, we should consider using the context menu? 🤔 

Screenshot 2024-04-09 at 21 05 20

andymandias:
Originally I had intended to put it behind left-click, but then I remembered that's already in use (and I'm not sure we want to slow down its current use case).  So inside the context menu makes sense to me :).  Just as long as there's enough space in there for the information.  I should be able to switch it over and update the PR later today, then we can give it a test spin.

A possible alternative that comes to mind: we could add a delay to the username tooltips, so they only appear after hovering on the user for ~1s.  I think I would prefer inside the context menu, but if it turns out it's too crowded in there.

casperstorm:
This is better @andymandias! Actually really like this - how about you?
About UI (sorry for always being nit picky haha) what if we did:

> | Online&nbsp;&nbsp;&nbsp;&nbsp; • |
> | Offline&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;• |
> | Away &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;• |

That way the text will left align with the other buttons below.

andymandias:
Nit picking always welcome, we wind up with better PRs as a result :)  In this case I think I like the layout quite a bit better with your suggestion;  moving the symbol to the right gets it out from under the cursor, easier to see at glance.

I have an iced layout issue I'm having trouble resolving.  Hoping you can help with seeing what mistake's being made.  In particular, the symbol on the right does not line up with the end of the text from the hostname above.  For example:

Screenshot from 2024-04-12 15-34-41

Where I temporarily made the operator symbol `a` (instead of `@`) to clarify the comparison with the hostname (and re-included access level info in the context menu for testing purposes, that can be removed again).

I've nominally fixed the misalignment by adding a 1 pixel padding to the right side of the rows with symbols, but that's less than ideal.  The current commit uses the width of the first text box to set the spacing, but I also tried putting a `horizontal_space()` between the two row entries instead.  Unfortunately `horizontal_space()` produced the same misalignment.  Only tested two fonts, but it seems that it happens regardless of font.

andymandias:
Well, I think I found out what's going on.  `double_pass` expands the limits to eliminate floating-point issues, making it one pixel wider/taller than it otherwise would be.  That puts the right side of the rows one pixel to the right of the longest text in the context menu.  If I remove the horizontal expansion from `double_pass` and set `symbol_padding` to `[0, 0, 0, 0]` in `user_context` then everything is lined up again.  I don't think that we want to mess with the expansion in `double_pass`.  I'll push a potential fix, but would be happy to hear of a better one.

casperstorm:
Do we really need the host? Or could we KISS and just do the status for now? 🤔 hard to make it look good with so long host unless we limit it size which then makes it a quite useless.

andymandias:
Sounds good to me, the status is all I need there.  Just wasn't sure where you had landed on the other info.  Pushed the simplification.

casperstorm:
> Sounds good to me, the status is all I need there. Just wasn't sure where you had landed on the other info. Pushed the simplification.

This is why I love working with you @andymandias - really happy that we can go forward and backwards on a feature like this.

src/buffer/user_context.rs
```diff
@@ -80,42 +84,103 @@ pub fn view<'a>(
...
+                .on_press(message)
+                .into()
+        } else {
+            column![]
+                .push(container(content).padding(5).width(length))
+                .push(
+                    row![]
+                        .push(horizontal_rule(1))
+                        .padding([0, 5])
+                        .width(length),
```
casperstorm, line 133: Does this have to be a `row![]`?

src/buffer/user_context.rs
```diff
@@ -80,42 +84,103 @@ pub fn view<'a>(
...
+                )
+                .into()
+        }
     })
 }
+
+fn button_text(content: &str) -> Element<'_, Message> {
+    text(content).style(theme::text::primary).into()
```
casperstorm, line 141: `.style(theme::text::primary)` is the default style for text, so i dont think we need this.

src/buffer/user_context.rs
andymandias: I made it a `row![]` to be able to apply the `padding` and `width` to it.  I know how to move the `padding` to the other widgets, but I don't know how to get the effects of `width` without the `row![]` (without `width` the `horizontal_rule` makes the context menu very wide).

src/buffer/user_context.rs
andymandias: When I remove `.style(theme::text::primary)` the text windows up being black.  Not sure if I've messed with the text context somehow, such that the default has been lost.

andymandias:
I've rearranged the logic a bit to try to improve the legibility of the code and remove some of the cruft.  I'm not entirely sure whether the changes address the reviews that were made, but hopefully they help.

src/buffer/user_context.rs
casperstorm: Ah I forgot. It's because theme will override text color.

casperstorm (approved):

PR #317, opened 2024-04-03
Use text fragment to avoid double allocation

tarkah:
Saves about 2MB of heap usage 

Had to update iced so more theme changes


casperstorm (approved):

casperstorm:
2mb is 2mb!

PR #320, opened 2024-04-04
Switch tokio-native-tls -> tokio-rustls

oldgalileo:
- Enables using PEM with certificates and private key side-by-side
- Resolves problem unique to macOS's SecureTransport resulting in users seeing:
    `a TLS error occurred: Unknown format in import.`

@tarkah I've confirmed that this does fix the issue on macOS

Fixed feature regressions:
- [x] Disabling certificate validation no longer works (although in the original it was always disabled 👻 )
- [x] Adding a specific root cert no longer works
- [x] Splitting the certificate and the private key no longer works


data/src/server.rs
```diff
@@ -77,6 +77,7 @@ impl Map {
                         *password = Some(pass);
                     },
                     Sasl::Plain { password: Some(_), password_file: None, .. } => {},
+                    Sasl::External { .. } => {},
```
casperstorm, line 80: I think we can turn the entire `match sasl` into a `if let`:

```rust
if let Sasl::Plain { password: password @ None, password_file: Some(pass_file), .. } = sasl {
    let pass = fs::read_to_string(pass_file)?;
    *password = Some(pass);
}
```

tarkah: resolved by f47da01

tarkah (approved):
Huge thanks @oldgalileo!

irc/src/connection/tls.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    for cert in rustls_native_certs::load_native_certs()? {
+        roots.add(cert).unwrap();
+    }
+    if let Some(cert_path) = root_cert_path {
+        let cert_bytes = fs::read(&cert_path).await?;
+        let certs =
+            rustls_pemfile::certs(&mut Cursor::new(&cert_bytes)).collect::<Result<Vec<_>, _>>()?;
+        roots.add_parsable_certificates(certs);
+    }
```
oldgalileo, line 33: This is only relevant if `accept_invalid_certs == false`

tarkah: Good point 364db2e

Issue #305, opened 2024-03-28
Option to move connection broadcasts to server's buffer instead of channels.


qauff:
Can we have an option to hide/move the `connection to server lost (ping timeout)` and `connection to server restored` server messages? 

Currently they show up in my bouncer's history and it's hard to follow sometimes.

<img width="448" alt="Screenshot 2024-03-28 at 13 07 00" src="https://github.com/squidowl/halloy/assets/15840269/1b23d00d-1231-4265-9d49-44ea144e1129">

Preferably I only want to see these in the server buffer, and not my channel buffers. If that isn't possible, maybe we can get options similar to the topic, join, part, and quit options.

I'm not sure what the best naming convention would be, but something along the lines of:
- `buffer.server_messages.disconnected`
- `buffer.server_messages.reconnected`

PR #322, opened 2024-04-05
Option to configure display of server status success & errors.

qauff:
Added options to configure server status success & error messages. I was getting a lot of noisy `connection to server lost` and `connection to server restored` messages in my channels. Figured I'd try my hand at adding the options to Halloy.

Screenshot 2024-03-28 at 13 07 00

Added two new configuration settings:
- `buffer.server_messages.status_success`
- `buffer.server_messages.status_error`

Both use the existing `buffer.server_messages.x` options. Specifically:

```
[buffer.server_messages.status_success]
enabled = true
smart = 7200

[buffer.server_messages.status_error]
enabled = true
smart = 7200
```

Related to issue #305

### DIFF
--- a/data/src/config/buffer.rs
+++ b/data/src/config/buffer.rs
@@ -31,6 +31,10 @@ pub struct ServerMessages {
     pub part: ServerMessage,
     #[serde(default)]
     pub quit: ServerMessage,
+    #[serde(default)]
+    pub status_success: ServerMessage,
+    #[serde(default)]
+    pub status_error: ServerMessage,
 }
 
 impl ServerMessages {
@@ -40,6 +44,8 @@ impl ServerMessages {
             source::server::Kind::Part => Some(self.part),
             source::server::Kind::Quit => Some(self.quit),
             source::server::Kind::Join => Some(self.join),
+            source::server::Kind::StatusSuccess => Some(self.status_success),
+            source::server::Kind::StatusError => Some(self.status_error),
         }
     }
 }
--- a/data/src/config/buffer.rs
+++ b/data/src/config/buffer.rs
@@ -19,6 +19,8 @@ pub struct Buffer {
     pub channel: Channel,
     #[serde(default)]
     pub server_messages: ServerMessages,
+    #[serde(default)]
+    pub internal_messages: InternalMessages,
 }
 
 #[derive(Debug, Clone, Default, Deserialize)]
@@ -31,10 +33,6 @@ pub struct ServerMessages {
     pub part: ServerMessage,
     #[serde(default)]
     pub quit: ServerMessage,
-    #[serde(default)]
-    pub status_success: ServerMessage,
-    #[serde(default)]
-    pub status_error: ServerMessage,
 }
 
 impl ServerMessages {
@@ -44,8 +42,6 @@ impl ServerMessages {
             source::server::Kind::Part => Some(self.part),
             source::server::Kind::Quit => Some(self.quit),
             source::server::Kind::Join => Some(self.join),
-            source::server::Kind::StatusSuccess => Some(self.status_success),
-            source::server::Kind::StatusError => Some(self.status_error),
         }
     }
 }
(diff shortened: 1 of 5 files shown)


data/src/message/source.rs
```diff
@@ -70,11 +64,13 @@ pub mod server {
         Part,
         Quit,
         ReplyTopic,
+        StatusSuccess,
+        StatusError,
```
tarkah, line 68: We shouldn't need these new new message kinds, we already cover this by tagging the messages as `Source::Internal`

data/src/history/manager.rs
```diff
@@ -491,6 +498,43 @@ impl Data {
...
+                                    }
+                                }
+                            };
+
+                            return !smart_filter_message(
+                                message,
+                                &seconds,
+                                most_recent_messages.get(&nick),
+                            );
+                        }
```
tarkah, line 533: Smart filtering does nothing for these messages since they're internal. No user is related to these messages.

data/src/config/buffer.rs
```diff
@@ -40,6 +44,8 @@ impl ServerMessages {
             source::server::Kind::Part => Some(self.part),
             source::server::Kind::Quit => Some(self.quit),
             source::server::Kind::Join => Some(self.join),
+            source::server::Kind::StatusSuccess => Some(self.status_success),
+            source::server::Kind::StatusError => Some(self.status_error),
```
tarkah, line 48: Let's add a new method `get_internal` which matches on `source::Internal` since that's what these messages actually are. Then we can get rid of this unneeded server source.

tarkah: Alternatively, we can add a new `InternalMessages` / `buffer.internal_messages.success / error` since that better maps to our data model and better describes what these messages are. They aren't coming from the upstream server.

Or maybe we rename `server_messages` -> `messages` so it can encapsulate both. (we can alias `messages` to `server_messages` during deserialization as to not break backwards compatibility).

@casperstorm thoughts?

data/src/history/manager.rs
```diff
@@ -491,6 +498,43 @@ impl Data {
...
+                Source::Internal(source::Internal::Status(status)) => {
+                    let source: source::Server;
+                    match status {
+                        source::Status::Success => {
+                            source = source::Server::new(source::server::Kind::StatusSuccess, None);
+                        }
+                        source::Status::Error => {
+                            source = source::Server::new(source::server::Kind::StatusError, None);
+                        }
+                    };
```
tarkah, line 510: Once the other comment is addressed, this is unnecessary since we should be able to use `Source::Internal(internal)` value to figure out if the config option is set

tarkah (changes requested):
Thanks for your first contribution to halloy! A few comments and one design decision we need to make, but this looks great!

data/src/history/manager.rs
tarkah: Maybe we need a new filter option which suppresses "duplicate" messages within some time frame?

qauff: Hm. I was thinking of fetching the last internal message and matching against that. But maybe duplicate messages would be better. Thinking about how I'd like this smart feature to work, it would make sense to show the last errors. But then, if you see a `reconnected` but not the `disconnected` that may be odd. I was more focused on hiding all of these in the channel buffer than I was thinking how best to display these 😅 @tarkah and team, would love to hear thoughts and suggestions on this!

data/src/config/buffer.rs
qauff: I like the explicit `InternalMessages` / `buffer.internal_messages.success / error`. I had considered originally doing something like `buffer.server_messages.status.error`. So I like the internal_messages approach you laid out here!

data/src/history/manager.rs
tarkah: Let's just do `enabled` for now then as I agree, I'm not sure another approach makes sense here. We can always add it down the road if a need arises

qauff: I realized it didn't feel right to show nothing at all. So I added logic to the `smart` option to display if the message was received within the duration configured 🙏 Hopefully this is okay for now!

data/src/config/buffer.rs
qauff: Refactored! This was how I should have written it in the first place. Hope it's this is more acceptable.

data/src/history/manager.rs
qauff: You were right! Got rid of this section in the refactor.

data/src/message/source.rs
qauff: Addressed in the latest refactor push!

data/src/config/buffer.rs
casperstorm: I think `internal_messages` works well.

data/src/history/manager.rs
```diff
@@ -491,6 +498,19 @@ impl Data {
 
                     true
                 }
+                message::Source::Internal(message::source::Internal::Status(status)) => {
+                    if let Some(server_message) = buffer_config.internal_messages.get(&status) {
```
casperstorm, line 502: You can rename `server_message` here to `internal_message`.

qauff: Done!

casperstorm (approved):
Congrats on your first contribution!

tarkah (approved):
Great changes, this looks super clean!

tarkah:
@qauff can you fix the clippy error? https://github.com/squidowl/halloy/actions/runs/8576207366/job/23635454340?pr=322

`cargo clippy --workspace`

qauff:
> @qauff can you fix the clippy error? https://github.com/squidowl/halloy/actions/runs/8576207366/job/23635454340?pr=322
> 
> 
> 
> `cargo clippy --workspace`

Done! Pushed up the fix!

Issue #321, opened 2024-04-04
password_file option not working in Flatpak version


shombando:
Through extended trial and error with @tarkah we discovered that the Flatpak version `halloy 2024.5` is not passing along the password correctly.

This works:
```toml
username = "shom"
password = "hunter2"
```
But storing the password in a file does not work:
```sh
echo "hunter2" > ./passwd
```
```toml
username = "shom"
password_file = "./passwd"
```
This is also true when using a full path `"/home/shom/.var/app/org.squidowl.halloy/config/halloy/passwd"`

The same config file works fine with `halloy 2024.5 (cc8e287)`. 

---
`RUST_LOG=trace ~/.local/share/flatpak/exports/bin/org.squidowl.halloy`: [Log](https://0x0.st/Xii3.txt)
What the interface shows: [screesnshot](https://0x0.st/XiiQ.png)
`config.toml`: [full config file](https://0x0.st/Xi-i.txt)

casperstorm:
Thanks for the bug report, i have not been able to reproduce it on my machines with macOS and Windows so it must indeed be a Flatpak thing.

shombando:
Thanks for checking into it and the support on irc! 

Looks like the password_file feature came in with PR#281 which landed after the flatpak version (release tag 2024.5). I'll wait for a flatpak release that includes 281 and then verify it works as intended and then close the issue.

A couple suggestions:
- Might be good to error on unrecognized parameter on TOML (but that could be undesirable when testing multiple version with same config file, but a thought) 
- The website documentation could be versioned so the user can pick a specific release to see available features

tarkah:
> The website documentation could be versioned so the user can pick a specific release to see available features

We've addressed this already and will only update website in lock-step w/ each release

tarkah:
@shombando thanks again for sticking with this. I've added #323 to at least prevent the most common issue which will be trailing newline / whitespace making the password incorrect.

shombando:
@tarkah that sounds great! Thanks. 
In the future, if/when halloy is in multiple package repos then the there will be some delta time where the versions will be out of sync. And if the package is moving fast then there will always a gap. One option/idea would be to package the website and distribute it so the documentation (locked with version) will be available locally for the user.

tarkah:
@casperstorm how hard would it be to offer a versioned mdbook? Default url goes to latest but maybe we add a version slug which routes to versioned?

Then our open wiki button in halloy can always open to the respective version

PR #323, opened 2024-04-05
Trim password files to remove leading / trailing whitespace

tarkah:
Resolves https://github.com/squidowl/halloy/issues/321


casperstorm (approved):

funkeleinhorn:
I guess this does not resolve #321 as the error there was an outdated version and not an error with the password file functionality. Also I think it is a bad practice to trim the passwords when read from file as some people might use whitespace intentionally in their passwords. At least we should document if we tinker with the passwords so it is transparent to the user.

PR #328, opened 2024-04-08
simplified onboarding

casperstorm:
Previously we create a `config.template.toml` and had the user rename it.
This wasn't super elegant, so I changed it slightly so we just create a `config.toml` for them directly.
I also updated the text on the welcome screen to match this new change.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+Changed:
+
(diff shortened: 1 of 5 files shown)


airstrike:
this is neat, thank you. next we just need a config edit screen with a text editor (or, even better, with widgets) ;-)

tarkah (approved):
LGTM, nice change!

PR #329, opened 2024-04-09
xdg path for macos

### LINKED ISSUE #327: Support standard $XDG_CONFIG_HOME setup on Mac
Currently config is expected in Application Support, but a lot of modern tools respect $XDG_CONFIG_HOME environment variable on macs as well.

casperstorm:
Fixes #327.

@tarkah can you verify i didn't break anything on linux here?
Main idea here is that `dirs-next` targets `~/Library/Application Support` for macOS, so i've added support to also look at `$HOME/.config` which is also used on macOS. Apple recommends using `Application Support` so I kept it our default location.

### DIFF
--- a/data/src/environment.rs
+++ b/data/src/environment.rs
@@ -18,11 +18,7 @@ pub fn formatted_version() -> String {
 }
 
 pub fn config_dir() -> PathBuf {
-    portable_dir().unwrap_or_else(|| {
-        dirs_next::config_dir()
-            .expect("expected valid config dir")
-            .join("halloy")
-    })
+    portable_dir().unwrap_or_else(platform_specific_config_dir)
 }
 
 pub fn data_dir() -> PathBuf {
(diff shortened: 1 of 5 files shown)


data/src/environment.rs
```diff
@@ -43,3 +35,49 @@ fn portable_dir() -> Option<PathBuf> {
...
+}
+
+#[cfg(target_os = "macos")]
+fn xdg_config_dir() -> Option<PathBuf> {
+    let Some(config_dir) = xdg::BaseDirectories::with_prefix("halloy")
+        .ok()
+        .and_then(|xdg| xdg.find_config_file(CONFIG_FILE_NAME))
+    else {
+        return None;
+    };
```
tarkah, line 80: Nit: Can use `?`

data/src/environment.rs
```diff
@@ -43,3 +35,49 @@ fn portable_dir() -> Option<PathBuf> {
...
+        dirs_next::config_dir()
+            .expect("expected valid config dir")
+            .join("halloy")
+    }
+}
+
+fn platform_specific_data_dir() -> PathBuf {
+    #[cfg(target_os = "macos")]
+    {
+        xdg_config_dir().unwrap_or_else(|| {
```
tarkah, line 59: As we discussed, let's not change `data_dir` behavior. We especially DO NOT want to use XDG_CONFIG_HOME to store data files.

tarkah (approved):
LGTM, changelog?

PR #333, opened 2024-04-11
reload configuration

casperstorm:


src/main.rs
```diff
@@ -219,21 +217,37 @@ impl Application for Halloy {
...
+
                 // Retrack after dashboard state changes
                 let track = dashboard.track();
 
-                Command::batch(vec![
+                if let Some(event) = event {
+                    match event {
+                        dashboard::Event::ReloadConfiguration => {
+                            // TODO: 
+                            // 1. show error screen if fails.
```
tarkah, line 236: 1. Would be nice if we could flash a quick message and have it dissapear? Not sure we need to take over the whole screen vs just briefly notifying the user.
2. `self.servers` is what drives the servers. So any _new_ server in the config that isn't in `self.servers` can be added and that server will auto-magically connect / show up (we return `self.servers` in `Application::subscription` to create the stream subscription per server). Existing servers, I don't think we _should_ do anything w/ them as that'd cause us to reconnect, is that desired?

casperstorm:
@tarkah give this branch a go and see how you like it.

src/main.rs
```diff
@@ -560,11 +592,23 @@ impl Application for Halloy {
...
+
+        let content = container(screen)
             .width(Length::Fill)
             .height(Length::Fill)
             .style(theme::container::primary)
-            .into()
+            .into();
+
+        match modal {
+            Some(modal) => widget::modal(content, modal).into(),
```
tarkah, line 609: I'll look into this, but we don't want to change the shape of view at this root level because all our text caching lives in the state tree.

When view changes and state tree is diffed, it'll cause everything to get dropped and recreated (expensive!!)

Ideally we have a way to toggle the modal dynamically without changing the view tree to prevent this from happening.

casperstorm: Sounds good, wanted to ask you about this because I did notice some slow down. Happy if you can look into it.

tarkah (approved):
This is great!! This will be a huge UX boost. And the modal looks :ok_hand:

PR #337, opened 2024-04-15
Dismiss Context Menu with Escape

andymandias:
A small change to allow context menus to be dismissed by pressing escape.  This is my first time implementing an operation in iced, but it seems to work.


casperstorm:
Great work! Could you add entry to CHANGELOG?

andymandias:
Done.  Sorry, wasn't sure if it warranted an entry.  Thanks for the review!

Issue #309, opened 2024-04-01
High Memory Usage when Using ZNC Clientbuffer Module


andymandias:
When using Halloy connected to ZNC with the Clientbuffer module memory usage grows continually during use.  When otherwise running Halloy I see memory usage start around 100MB and grow up to around 150MB, but when connected to ZNC with the Clientbuffer module memory grows unbounded.  In around an hour memory is often up to 250MB and the UI slows down, but memory use will continue to grow until the application stalls and must be force quit.  The largest memory use I've seen so far is 3.5GB, but usually I don't let it get that high.

It doesn't matter whether I use the new nickname format or not (i.e. using an `@` in my username to identify the client is not necessary to produce the behavior).  It seems to progress faster when two clients are connected to the ZNC bouncer, but it still happens when only one client is connected.  Happens on Linux (both x86_64 and aarch64, though it progresses faster on the latter).  I tried reproducing it under DHAT (valgrind), but annoying it does not seem to reproduce no matter how long I wait.

One thing I have noticed, is that the server buffer gets a lot of `RPL_WHOREPLY` messages in it.  I think this is somewhat expected when two (or more) clients are connected, since `last_who` will not (to my knowledge) reflect the `WHO` requests from another client (i.e. all the `WHO` polls sent from another client will show up in the server buffer).  For some reason this also happens when only one client is connected.  Since the server buffer history is capped at 10000 messages (as far as I know) I don't expect it to causing the memory leak, but it's the only different behavior I've been able to spot so far.

tarkah:
Can you try running it with `heaptrack`? And debug mode might make the stack traces make more sense. Or compile release mode w/ debug info enabled.

You're right that we cap how many messages we store in memory and how many messages we render to the scrollable. I'll try to think if there's anywhere we allow memory to grow unbounded.

andymandias:
Running with a `cargo build` executable from built from `main`, running with `heaptrack` then waiting for memory usage to grow to ~200MB, then quitting Halloy normally.  `heaptrack` is reporting memory leaks, but the biggest one is `alloc::alloc::alloc` clocking in at 18.2MB.  But following it up the backtrace just leads to iced `load_font`, so I'm guessing it's not a real leak just the font not being explicitly unloaded.

There are apparently ~88MB of Vulkan allocations at peak memory usage (where `heaptrack` reports the total memory at peak as 149MB).  I guess that might be a likely candidate, but I don't know if there's any useful information in the related backtrace:
```
[...]

PR #340, opened 2024-04-17
feat: updated iced to master

casperstorm:
@andymandias could you test this branch re #309.
@hecrj updated iced with https://github.com/iced-rs/iced/pull/2389.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1925,7 +1925,7 @@ dependencies = [
 [[package]]
 name = "iced"
 version = "0.13.0-dev"
-source = "git+https://github.com/iced-rs/iced?rev=ba705d63dd52160e5c0b90024bb2d6f95e3f6762#ba705d63dd52160e5c0b90024bb2d6f95e3f6762"
+source = "git+https://github.com/iced-rs/iced?rev=a05b8044a9a82c1802d4d97f1723e24b9d9dad9c#a05b8044a9a82c1802d4d97f1723e24b9d9dad9c"
 dependencies = [
  "iced_core",
  "iced_futures",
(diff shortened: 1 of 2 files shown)


andymandias:
Can do.  Should I test with Vulkan disabled to rule out any issues from there?  I can do both, but it takes a bit longer to get a clear signal when Vulkan is disabled.

hecrj:
You should update to `a05b8044a9a82c1802d4d97f1723e24b9d9dad9c`. I fixed a worker thread panic related to the backpressure changes.

andymandias:
From initial testing (with `tiny-skia` backend to exclude Vulkan issues), growth appears slower but hasn't shown signs of stopping yet.  I'm going to continue to test to verify that it's doesn't just have a higher steady-state memory usage than I'm expecting (as well as continue to test configurations such that it can be more reliably/quickly tested;  I *think* I'm getting close to reproducing the issue without ZNC involved).

I'm reporting back a little bit early because, for the first time since tracking this issue, I occasionally see the memory decrease while in use.  It seems like when I open the server buffer, for one of the servers that has a high rate of `WHO` polling, that is when memory usage might decrease;  but that doesn't happen every time I open the server buffer, and the memory use continues to rise afterward.

tarkah:
I'd suspect a lot of this comes down to the allocator used by default on Linux with Rust and allocated pages not getting released back to the kernel.

You can try compiling Halloy with jemalloc which prioritizes reducing memory fragmentation which means it returns unused pages of memory back to the kernel much more aggressively. 

I think this is why memory reported by top can be a lot higher than actual usage you see reported by heaptrack.

andymandias:
Overnight test and it looks like memory use has just settled at a higher level than I was expecting.  :partying_face:   And @tarkah's explanation for the behavior makes perfect sense to me as to why memory use would appear larger than it actually is.  I think we can consider this as solving #309, and I can look into whether I need to open issues for Vulkan further upstream.  Thank you everyone for looking into this!  I think we got a lot of blood out of that stone :slightly_smiling_face:

PR #342, opened 2024-04-18
CAP NEW and CAP DEL Subcommands

andymandias:
A first implementation for adding support for the IRCv3 `CAP NEW` and `CAP DEL` subcommands.  The intention is to follow the existing logic for the other `CAP` subcommands, with the one exception being that `sasl` is ignored when provided via `CAP NEW` since I assumed it would be too late to authenticate via SASL.

Works with my ZNC bouncer (v1.9.0), which only provides `away-notify` via `CAP NEW` (resulting in `WHO` polling for Libera which was not needed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Added:
 
 - Allow configuration of internal messages in buffer (see [buffer configuration](https://halloy.squidowl.org/configuration/buffer.html#bufferinternal_messages-section))
 - User information added to context menu.
+- Support for IRCv3 CAP NEW and CAP DEL subcommands.
 
 Changed:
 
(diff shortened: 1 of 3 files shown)


casperstorm:
Really nice one, i appreciate the help with CAPS.
Do we need to update this list with this MR? https://github.com/squidowl/halloy/blob/main/book/src/README.md?plain=1#L7

tarkah (commented):
LGTM.! Do we need to add `cap-notify` to our requested capabilities so the server knows we support this?

I haven't tested but it all makes sense otherwise. We have a refactor opportunity to make our supported capabilities strongly typed, but this works great for now.

andymandias:
My reading of https://ircv3.net/specs/extensions/capability-negotiation.html#cap-notify is that negotiating with version 302 (which I believe we do via `CAP LS 302`) should implicitly enable `cap-notify`.  Doesn't seem like we'd need to request it explicitly, unless there's a fallback state that doesn't use `CAP LS 302` which I've overlooked.

casperstorm (approved):

andymandias:
> Do we need to update this list with this MR? https://github.com/squidowl/halloy/blob/main/book/src/README.md?plain=1#L7

As far as I can tell we can claim support for `cap-notify`, so I've added it to the list.

tarkah (approved):
Awesome, thanks!

Issue #206, opened 2023-09-28
Feature request: IRCv3 CHATHISTORY support


Mikaela:
The [IRCv3 chathistory](https://ircv3.net/specs/extensions/chathistory) would make using Halloy with [Ergo IRCd](https://ergo.chat) more seamless as it's [their primary method of synchronising history to client from integrated bouncer](https://github.com/ergochat/ergo/blob/master/docs/USERGUIDE.md#history).

PR #343, opened 2024-04-19
Request `multi-prefix` when Available

andymandias:
Was looking at https://ircv3.net/software/clients to see if we can turn a few more table cells green, and I think we can enable `multi-prefix` with the current code base by just adding it to the capabilities we request when available (which is what this PR does).  `multi-prefix` (https://ircv3.net/specs/extensions/multi-prefix) modifies the output of `NAMES`, `WHO`, and `WHOIS` to include all user prefixes for each user (instead of just the highest ranking prefix).

`NAMES` (via `RPL_NAMREPLY`) passes the [prefix]nick strs to `User::try_from`, which does not make any assumptions about the number of prefixes present.

`WHO` only updates users' away state from the flags, it does not parse the rest of flags.  So the additional prefixes in flags shouldn't effect its operation.

`WHOIS` is not parsed at all, its output is just sent to the user.

All three work in my testing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,9 @@
 Added:
 
 - Allow configuration of internal messages in buffer (see [buffer configuration](https://halloy.squidowl.org/configuration/buffer.html#bufferinternal_messages-section))
-- User information added to context menu.
-- Support for IRCv3 CAP NEW and CAP DEL subcommands.
+- User information added to context menu
+- Support for IRCv3 CAP NEW and CAP DEL subcommands
+- Enable support for IRCv3 `multi-prefix`
 
 Changed:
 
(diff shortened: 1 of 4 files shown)


casperstorm:
Big thanks to help with all the IRCv3 caps, hugely appreciated. Another one we should consider is https://github.com/squidowl/halloy/issues/206.

casperstorm (approved):
LGTM, and works in my testing as well.

andymandias:
> Another one we should consider is #206.

Next on my list (after I finish another tiny PR) is to look into `RPL_ISUPPORT` support, which should help with this.  I'll put `message-tags` on my list after that (looks like it might be another quick one), then we'll be well positioned to implement `draft/chathistory` (which I'd be happy to help with).

tarkah (approved):
Sweet!

PR #344, opened 2024-04-19
Hide Output of `NAMES` that Result from `JOIN`

andymandias:
A small change to hide the results of the `NAMES` commands that are automatically triggered when the user joins a channel.  According to https://modern.ircdocs.horse/#join-message servers must send those `RPL_NAMREPLY` and `RPL_ENDOFNAMES` messages automatically.  I think this change fits with the hiding of `RPL_WHOREPLY` due to automatic polling, and it produces a nice clean server buffer.  I think we can skip a changelog entry for this one, assuming the new behavior is desirable.

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -757,6 +757,24 @@ impl Client {
                             channel.users.insert(user);
                         }
                     }
+
+                    // Don't save to history if names list was triggered by JOIN
+                    if !channel.names_init {
+                        return None;
+                    }
+                }
+            }
+            Command::Numeric(RPL_ENDOFNAMES, args) => {
+                let target = args.get(1)?;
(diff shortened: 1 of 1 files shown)


andymandias:
I wasn't particularly satisfied with `names_init` either, but I also couldn't think of a better name :sweat_smile:

casperstorm:
> I wasn't particularly satisfied with `names_init` either, but I also couldn't think of a better name :sweat_smile: 

ROFL, I just knew it! Also why I brought it up - I don't have a better name as well so I'm all fine with this haha.

tarkah (approved):
Lol naming is definitely the hardest part of programming. This is long overdue, thanks!

PR #346, opened 2024-04-20
Handle `ERR_NEEDREGGEDNICK` (aka `ERR_NOCHANMODES`)

andymandias:
When logging in via NickServ IDENTIFY, channels that require registration are joined before the user is logged in.  As a result, channels that require registration reject the join and the channel doesn't appear on the sidebar.  I think this behavior is correct, since most channels don't require registration and waiting for the user to log in would (in my experience) significantly slow down the typical load time.  This PR instead has Halloy look for a [`ERR_NEEDREGGEDNICK`](https://defs.ircdocs.horse/defs/numerics.html#err-needreggednick-477) message, then adds the channel to a vector of channels to be joined after the user has logged in.  After receiving `RPL_LOGGEDIN` those channels will be joined again.  Tested successfully with `#znc` on Libera.

Unfortunately there is a conflicting definition for numeric 477 ([`ERR_NOCHANMODES`](https://defs.ircdocs.horse/defs/numerics.html#err-nochanmodes-477)).  To try to avoid possible false positives it's checked to see that both (a) the user is not in the channel referenced by the numeric and (b) the channel is in the user's config.  I would prefer this to be more bulletproof, but I think it is a worthwhile tradeoff between improved typical user experience and potential edge case weirdness.


casperstorm:
This looks good to me, @tarkah do you have any thing for this?

casperstorm (approved):

data/src/client.rs
```diff
@@ -1175,17 +1201,19 @@ pub enum WhoStatus {
 }
 
 /// Group channels together into as few JOIN messages as possible
-fn group_joins(config: &config::Server) -> impl Iterator<Item = proto::Message> + '_ {
+fn group_joins<'a>(
+    config: &'a config::Server,
+    channels: &'a [String],
```
tarkah, line 1206: Can we update this to take keys instead of the entire config?

It's confusing (and a code smell) we pass in self.config and self.config.channels.

tarkah (changes requested):

data/src/client.rs
tarkah: lets do `channels: &'a [String], keys: &'a [String]`

andymandias: An artifact of the first implementation that I neglected to revisit.  You're absolutely right of course, the interface doesn't really make sense.  Thank you for catching it!  :)

tarkah (approved):

casperstorm (approved):

Issue #334, opened 2024-04-12
sha256sums on release


KaiKorla:
Please create a sha256sum.txt or something similar with the sha256sums of each file in a release.

YouFoundAlpha:
I'm currently finding a solution to this using Github actions
I'll make a pull request as soon as I have tested the solution

casperstorm:
Sorry for late response on this, but it is now added to releases.

YouFoundAlpha:
> Sorry for late response on this, but it is now added to releases.
> Thanks for the help work, @YouFoundAlpha.

The usual names for the SHA256 hashes files are SHA256SUM or SHA256SUMS or checksums.txt
Could you maybe change it @casperstorm ?

PR #347, opened 2024-04-20
Add SHA256

YouFoundAlpha:


YouFoundAlpha:
PR for #334

YouFoundAlpha:
Keep in mind that the action only can get turned on manually and the files gets placed in the root of the repo

casperstorm:
I was actually about to just add this functionally to our existing release action so the file is only added to the release and not in the repo. Will finish it up asap.

YouFoundAlpha:
> I was actually about to just add this functionally to our existing release action so the file is only added to the release and not in the repo. Will finish it up asap. 

Sounds good

casperstorm:
Added a manual action for now so I could target previous releases: https://github.com/squidowl/halloy/blob/main/.github/workflows/checksum.yml
I can add it to auto run on published releases later, in any case this is now solved :)

YouFoundAlpha:
> Added a manual action for now so I could target previous releases: https://github.com/squidowl/halloy/blob/main/.github/workflows/checksum.yml
> I can add it to auto run on published releases later, in any case this is now solved :)

The usual names for the SHA256 hashes files are SHA256SUM or SHA256SUMS or checksums.txt

PR #348, opened 2024-04-21
Revert "Trim password files to remove leading / trailing whitespace"

funkeleinhorn:
Reverts squidowl/halloy#323

As stated in the conversation on #323 trimming the passwords is bad as some people might use white-space intentionally in their passwords, which get removed then without this behavior being documented. Instead of removing white-space from password files to help people enter their password correctly we should add more documentation on how to specify your password via a file (e.g. do not add white-space like a new line to the file as the whole file contents are treated as the password).


casperstorm:
Do you want to add the additional documentation in this PR as well, @funkeleinhorn? I agree we need it.

casperstorm (approved):
🚀

PR #349, opened 2024-04-21
ISUPPORT Support

andymandias:
Initial implementation for ISUPPORT parameter parsing.  Leans toward leniency in parsing where simple & possible.  Includes support for all active parameters & all known draft parameters, while support for deprecated parameters is limited to those that appear to be in use and/or easy to support.

While this PR attempts to be reasonably comprehensive in ISUPPORT parameter *parsing*, it is not aiming to be comprehensive in ISUPPORT parameter *utilization*.  Regardless, it would be a good test to implement some of the simpler parameter utilization.

Planned parameter utilization:
- [x] CNOTICE
- [x] CPRIVMSG
- [x] KNOCK
- [x] SAFELIST
- [x] USERIP
- [x] WHOX


andymandias:
The proof of concept implementations are nominally complete.  I don't know if we want to do any automatic handling of `RPL_WHOSPCRPL` (`WHOX`) replies since they're not guaranteed to have away status.  Would we want to use `WHOX` to minimize the size of the data sent for `WHO` polling?

I noticed that the command completions are not comprehensive, and I assume that is intentional (too many commands available there might be overwhelming).  So, while I don't know if any of the ISUPPORT gated commands I've added should be added to the completion list or not, I have added them for now.

I think this could be considered completing support for (`WHOX`)[https://ircv3.net/specs/extensions/whox] and (`UTF8ONLY`)[https://ircv3.net/specs/extensions/utf8-only] (since we're using Rust `String` I think everything we're sending is UTF-8 and everything we receive can be UTF-8).

data/src/isupport.rs
```diff
@@ -0,0 +1,614 @@
...
+    TARGMAX(Vec<CommandTargetLimit>),
+    TOPICLEN(u16),
+    UHNAMES,
+    USERIP,
+    USERLEN(u16),
+    UTF8ONLY,
+    VLIST(String),
+    WATCH(u16),
+    WHOX,
+    Negation(String),
```
tarkah, line 68: `Negation` isn't a parameter, it's an operation. Let's have `isupport::Operation::Add(Parameter) / Remove(String)`.

data/src/isupport.rs
```diff
@@ -0,0 +1,614 @@
...
+    'g'
+}
+
+pub fn default_deaf_letter() -> char {
+    'D'
+}
+
+pub fn default_invite_exception_letter() -> char {
+    'I'
+}
```
tarkah, line 570: Let's make these `const` variables:

```rs
const BAN_EXCEPTION_CHANNEL_LETTER: char = 'e'`;
```

data/src/isupport.rs
```diff
@@ -0,0 +1,614 @@
...
+    }
+}
+
+fn parse_required_letter(value: &str, default_value: Option<char>) -> Result<char, &'static str> {
+    if let Some(value) = value.chars().next() {
+        Ok(value)
+    } else if let Some(default_value) = default_value {
+        Ok(default_value)
+    } else {
+        Err("value required to be a letter")
```
tarkah, line 596: This error message doesn't line up w/ the logic: "required to be a letter" but we are actually checking it's a valid utf8 (implied by str) >1 char string.

data/src/isupport.rs
```diff
@@ -0,0 +1,614 @@
...
+impl<'a> TryFrom<&'a str> for Parameter {
+    type Error = &'static str;
+
+    fn try_from(isupport: &'a str) -> Result<Self, Self::Error> {
+        if isupport.is_empty() {
+            return Err("empty ISUPPORT parameter not allowed");
+        }
+
+        match isupport.chars().next() {
+            Some('-') => Ok(Parameter::Negation(isupport[1..].to_string())),
```
tarkah, line 88: Possible index error, let's ensure len > 1 (we definitely don't want to panic due to malformed messages)

data/src/isupport.rs
```diff
@@ -0,0 +1,614 @@
...
+
+impl TryFrom<String> for Parameter {
+    type Error = &'static str;
+
+    fn try_from(isupport: String) -> Result<Self, Self::Error> {
+        Self::try_from(isupport.as_str())
+    }
+}
+
+impl<'a> TryFrom<&'a str> for Parameter {
```
tarkah, line 79: Let's use `FromStr` implementation instead so we get `.parse()` available which I think is nicer

data/src/isupport.rs
```diff
@@ -0,0 +1,614 @@
...
+                        "BOT" => Ok(Parameter::BOT(parse_required_letter(value, None)?)),
+                        "CALLERID" => Ok(Parameter::CALLERID(parse_required_letter(
+                            value,
+                            Some(default_caller_id_letter()),
+                        )?)),
+                        "CASEMAPPING" => match value {
+                            "ascii" => Ok(Parameter::CASEMAPPING(CaseMap::ASCII)),
+                            "rfc1459" => Ok(Parameter::CASEMAPPING(CaseMap::RFC1459)),
+                            "rfc1459-strict" => Ok(Parameter::CASEMAPPING(CaseMap::RFC1459_STRICT)),
+                            "rfc7613" => Ok(Parameter::CASEMAPPING(CaseMap::RFC7613)),
```
tarkah, line 115: are these always went lowercase or should we call `value.lowercase()` just in case?

data/src/isupport.rs
```diff
@@ -0,0 +1,614 @@
...
+                                            prefix: prefix.to_string(),
+                                            limit: None,
+                                        });
+                                    } else if let Ok(limit) = limit.parse::<u16>() {
+                                        channel_limits.push(ChannelLimit {
+                                            prefix: prefix.to_string(),
+                                            limit: Some(limit),
+                                        });
+                                    }
+                                }
```
tarkah, line 134: `prefix` should be `chars()` and each distinct channel prefix character should be stored. Each character being a separate `CHANTYPE` 

https://modern.ircdocs.horse/#chanlimit-parameter

data/src/isupport.rs
```diff
@@ -0,0 +1,614 @@
...
+                            value
+                                .split(',')
+                                .for_each(|client_tag_denial| {
+                                    match client_tag_denial.chars().next() {
+                                        Some('*') => {
+                                            client_tag_denials.push(ClientOnlyTags::DenyAll)
+                                        }
+                                        Some('-') => {
+                                            client_tag_denials.push(ClientOnlyTags::Allowed(
+                                                client_tag_denial[1..].to_string(),
```
tarkah, line 180: Another possible index error

data/src/isupport.rs
```diff
@@ -0,0 +1,614 @@
...
+                        "NETWORK" => Ok(Parameter::NETWORK(value.to_string())),
+                        "NICKLEN" | "MAXNICKLEN" => {
+                            Ok(Parameter::NICKLEN(parse_required_positive_integer(value)?))
+                        }
+                        "OVERRIDE" => Ok(Parameter::OVERRIDE),
+                        "PREFIX" => {
+                            let mut prefix_maps = vec![];
+
+                            if let Some((modes, prefixes)) = value.split_once(')') {
+                                let modes = &modes[1..];
```
tarkah, line 311: Another possible index error,  we didn't confirm modes len > 0

data/src/client.rs
```diff
@@ -87,6 +87,7 @@ pub struct Client {
     supports_labels: bool,
     supports_away_notify: bool,
     highlight_blackout: HighlightBlackout,
+    isupport_parameters: HashMap<String, isupport::Parameter>,
```
tarkah, line 90: After removing `Negation`, we should just make this a `HashSet<Parameter>` so it's always strongly typed.

data/src/client.rs
```diff
@@ -1169,11 +1276,15 @@ pub struct Topic {
...
-    Requested(Instant),
+    Requested(Instant, Option<u16>),
     Receiving,
     Done(Instant),
 }
 
+fn who_polling_token() -> u16 {
+    9
+}
+
```
tarkah, line 1287: This can be made a `const` var, similar to the other comment

data/src/client.rs
```diff
@@ -644,8 +646,21 @@ impl Client {
 
                     if let Some(state) = self.chanmap.get_mut(channel) {
                         // Sends WHO to get away state on users.
-                        let _ = self.handle.try_send(command!("WHO", channel));
-                        state.last_who = Some(WhoStatus::Requested(Instant::now()));
+                        if self.isupport_parameters.get("WHOX").is_some() {
+                            let _ = self.handle.try_send(command!(
+                                "WHO",
+                                channel,
+                                "tcnf",
```
tarkah, line 653: I'm not sure we need `c` since we specify `WHO channel`

data/src/client.rs
```diff
@@ -689,6 +704,38 @@ impl Client {
...
+                            // H = Here, G = gone (away)
+                            let flags = args.get(4)?.chars().collect::<Vec<char>>();
+                            let away = *(flags.first()?) == 'G';
+
+                            let lookup = User::from(Nick::from(args[3].clone()));
+
+                            if let Some(mut user) = channel.users.take(&lookup) {
+                                user.update_away(away);
+                                channel.users.insert(user);
+                            }
```
tarkah, line 732: Let's put this logic in the same block as `if request_tag == tag` and only `return None` here. 

This is identical to the logic above in RPL_WHO, maybe we factor this logic into a new function so we don't need to maintain it twice?

data/src/client.rs
```diff
@@ -816,6 +863,47 @@ impl Client {
                 #[cfg(feature = "dev")]
                 return None;
             }
+            Command::Numeric(RPL_ISUPPORT, args) => {
+                args[1..].iter().for_each(|arg| {
```
tarkah, line 867: Another index error, let's ensure args > 0

tarkah:
Question: should we parse ISUPPORT params we don't plan on leveraging in some way? I realize you were probably just being proactive, but we're now storing a bunch of "dead data" that isn't being used.

I'd like us to stay slim and reduce overhead (cognitive as much as performance) so my gut says we should only deal w/ parsing ones we actual care about & use

data/src/client.rs
```diff
@@ -816,6 +863,47 @@ impl Client {
...
+                    if arg != "are supported by this server" {
```
tarkah, line 868: I don't think this is valid since the text may be something else (see @casperstorm comment).

We should skip the _last_ arg if it's a non-valid parameter.

data/src/client.rs
```diff
@@ -1169,11 +1276,15 @@ pub struct Topic {
...
 #[derive(Debug, Clone, Copy)]
 pub enum WhoStatus {
-    Requested(Instant),
+    Requested(Instant, Option<u16>),
     Receiving,
     Done(Instant),
 }
 
+fn who_polling_token() -> u16 {
+    9
```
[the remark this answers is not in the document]
andymandias: I chose it mostly arbitrarily (picked a single digit to make the response messages as short as possible).  It's only being set to help distinguish the `RPL_WHOSPCRPL` from a potential user performed `WHOX`. It's required to be 3 digits or fewer but otherwise does not have a restriction on it.  So we could potentially set it to something else, or have a new token be generated for every poll.

irc/proto/src/command.rs
```diff
@@ -229,6 +243,10 @@ impl Command {
             Command::USERHOST(params) => params,
             Command::WALLOPS(a) => vec![a],
             Command::BATCH(a, rest) => std::iter::once(a).chain(rest).collect(),
+            Command::CNOTICE(a, b, c) => vec![a, b, format!(":{}", c)],
+            Command::CPRIVMSG(a, b, c) => vec![a, b, format!(":{}", c)],
```
tarkah, line 247: Don't add `:` here, it's handled during `proto::format::parameters`

data/src/client.rs
andymandias: I originally tried `"tnf"` because my reading of the `WHOX` documentation was that we would get the requested channel in the reply automatically, but when testing on Libera the individual `RPL_WHOSPCRPL` messages do not include the channel.  But, we could use `"tnf"` regardless if we use the token to distinguish the channel.

data/src/client.rs
tarkah: https://github.com/squidowl/halloy/pull/349#issuecomment-2073438081

data/src/client.rs
tarkah: Let's just leave it then, seems like servers aren't consistent on implementation, but adding it doesn't cause an additional channel arg to come across

data/src/client.rs
tarkah: OR, we have `isupport::Kind` enum which lists just the param names which we care to track in Halloy and then `HashMap` that. `HashSet<Parameter>` is actually dangerous because it'd allow storing >1 parameter kind if the values are different inside the parameter.

src/widget/input/completion.rs
```diff
@@ -112,11 +119,30 @@ impl Commands {
...
+                    {
+                        &WHOX_COMMAND
+                    } else {
+                        command
+                    }
+                })
+                .chain(isupport_parameters.iter().filter_map(|isupport_parameter| {
+                    isupport_parameter_to_command(isupport_parameter)
+                }))
+                .collect::<Vec<_>>();
```
tarkah, line 139: Nice :+1: Love that we can now dynamically enable stuff via isupport

andymandias:
> Question: should we parse ISUPPORT params we don't plan on leveraging in some way? I realize you were probably just being proactive, but we're now storing a bunch of "dead data" that isn't being used.
> 
> I'd like us to stay slim and reduce overhead (cognitive as much as performance) so my gut says we should only deal w/ parsing ones we actual care about & use

Indeed the intention was mostly to be proactive, and to start work on features planned for implementation (which should probably get their own PR).  I'm completely on board with axing things if they're just getting in the way, though I think it would be best if we don't produce warnings in the log for ISUPPORT parameters that are known to be valid.  These commits will live on in the repository's history, so if there's ever any need to dredge up the code it can be done without much difficulty.

andymandias:
> Testing it out on libera works great so far, however in logs i am seeing the following:
Thanks for the test @casperstorm!  Bitten by ZNC producing a different trailing message.  No surprise, @tarkah has a great suggestion for this.

data/src/isupport.rs
andymandias: Great suggestion, thank you!  Implemented.

data/src/isupport.rs
andymandias: Good point!  Updated to require an ASCII alphabetic.

data/src/isupport.rs
andymandias: Didn't know `FromStr` would give us `.parse()`.  Very cool!  Implemented.

data/src/isupport.rs
andymandias: For some reason I thought it would result in an `Error` instead of a panic.  Thank you for the heads up!

data/src/isupport.rs
andymandias: My reading is that they're always sent lowercase, but I don't see any reason we need to be strict about that.  Added `to_lowercase()`.

data/src/client.rs
andymandias: I see the value in `HashMap<Kind,Parameter>`, though it's a shame we lose some of the simplicity that `Remove(String)` had.  I'll probably add a function to `Operation` that returns a `Kind`, but will play around to see if there's an even cleaner way first.

irc/proto/src/command.rs
andymandias: And *now* I understand what `:` means.  Thank you!

tarkah:
> I added a proof of concept for adding tooltips to the command list as a way to utilize the `*LEN` parameters. Would be happy to add the others if we like the functionality, or cut it if the tooltips are too much.

I think the tooltips are pretty nifty! It's a bit of a hidden feature, but cool nonetheless.

data/src/client.rs
```diff
@@ -679,35 +698,68 @@ impl Client {
...
+                                    let flags = args.get(4)?.chars().collect::<Vec<char>>();
+
+                                    if flags
+                                        .iter()
+                                        .all(|c| matches!(c, 'H' | 'G' | '&' | '@' | '%' | '+'))
+                                    {
+                                        if let Ok(user) = User::try_from(args.get(3)?.clone()) {
+                                            channel.update_user_away(user, flags);
+                                        }
+                                    }
```
tarkah, line 748: Whys this logic different than in `RPL_WHOREPLY`? Also, we don't need to have the logic here and do the token == token logic again, just put this up after line `730` like

data/src/client.rs
```diff
@@ -845,6 +897,51 @@ impl Client {
...
+                                }
+                            };
+                        }
+                        Err(error) => {
+                            if index != args_len - 1 {
+                                log::debug!(
+                                    "[{}] unable to parse ISUPPORT parameter: {} ({})",
+                                    self.server,
+                                    arg,
+                                    error
```
tarkah, line 936: Nit: We know we're dealing with `isupport` here, we can shorten some vars: `isupport_operation -> operation` / `isupport_parameter -> parameter`.

Also, we can just call `self.isupport_parameters -> self.isupport`, I think this will reduce a lot of visual redundancy / noise.

data/src/client.rs
andymandias: I added a bit more parameter checking to this call because the user could manually invoke a `WHOX` command with different tags specified than those used in `WHOX` polling.  I left it out of the `WHO` polling as a minor optimization, but we could definitely fold it into `update_user_away` for increased code simplicity.  Should I do that?

The `request_token == token` check above is for `WhoStatus::Requested`, so I believe it will only be matched for the very first `RPL_WHOSPCRPL` received.  I attempted to model that after how `RPL_WHOREPLY` handles its messages.

data/src/client.rs
andymandias: Sounds good to me!  I tend toward increased verbosity, so the guidance is greatly appreciated. :)

tarkah: I'm definitely all for fully typed names (parameters > params, etc) but here this is just redundant namespacing we don't need since it's implied in the context we're working in imo. At least that's my reasoning

andymandias:
All tooltips I'm planning to add have been added.  For the more complicated commands in particular (e.g. `WHOX` and `ELIST`) I think they're pretty useful.  Also added an indicator for when a tooltip is available (an superscript-styled asterisk for now);  still not the most discoverable, but it helps (hopefully).

Lots of parsing improvements based directly on and inspired by the reviews.  :)

data/src/client.rs
andymandias: After changing the `flags` parsing (couldn't reasonably make the check more stringent for `RPL_WHOSPCRPL` after looking further into what `flags` is allowed to contain), there really wasn't any reason to separate the two.  Went ahead and consolidated further.  Thanks for the suggestion.

tarkah:
@andymandias Is this in a good spot for a final review?

andymandias:
I think this is a good spot @tarkah.  Lots more work to be done on parameter support, but that should be broken off into more narrowly focused PRs.  This one is too big already;  sorry about that, was just testing the tooltips and kept finding info that would be useful to have.

andymandias:
Just a simple rebase to remove the merge conflict, and include the command auto-completion changes in the changelog (assuming we keep those changes).

tarkah (approved):
Huge lift @andymandias!! Thanks for addressing all the feedback, this is looking really great.

I cleaned up one thing since control flow here was kind of redundant / weird: abf2f8cfa5e7c5880cb2f21b919e565b85d77694

andymandias:
Thanks for the review @tarkah!  The `request_token == token` check inside the `last_who` is never hit after the first reply since `last_who` is no longer `WhoStatus::Requested` but is instead `WhoStatus::Receiving`.  That's why I had the awkward looking structure in there;  as it is now we get most of the `WHOX` polling inside the server buffer.

PR #350, opened 2024-04-23
Features/proxy support

### LINKED ISSUE #301: [Improvements] Proxy support ?
I need proxy support. A socks field in the [servers] toml config would be great

ramajd:
Fixes #301.

following from #301, this PR provides the support for `socks5` proxy. the main idea is to initiate the `TcpStream` in `irc::connection` module using the `tokio-socks`'s `Socks5Stream`. 

P.S: the initial implementation might not be efficient in favor of cloning the proxy configuration a couple of times.

### DIFF
--- a/data/src/config/server.rs
+++ b/data/src/config/server.rs
@@ -27,6 +27,8 @@ pub struct Server {
     /// The port to connect on.
     #[serde(default = "default_port")]
     pub port: u16,
+    /// Proxy configuration to use for connecting to the server.
+    pub proxy: Option<ProxyConfig>,
     /// The password to connect to the server.
     pub password: Option<String>,
     /// The file with the password to connect to the server.
@@ -105,6 +107,24 @@ impl Server {
             security,
         }
     }
+}
+
+#[derive(Debug, Clone, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub enum ProxyConfigType {
+    Socks5,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct ProxyConfig {
+    #[serde(rename = "type")]
+    pub proxy_type: ProxyConfigType,
+    pub host: String,
+    pub port: u16,
+    #[serde(default)]
+    pub username: String,
+    #[serde(default)]
+    pub password: String,
 }
 
 #[derive(Debug, Clone, Deserialize)]
--- a/data/src/config/server.rs
+++ b/data/src/config/server.rs
@@ -2,8 +2,10 @@ use std::collections::HashMap;
 use std::path::PathBuf;
 use std::time::Duration;
 
-use irc::connection::{self, Proxy};
+use irc::connection;
 use serde::{Deserialize, Deserializer};
+
+use super::proxy::Proxy;
 
 #[derive(Debug, Clone, Default, Deserialize)]
 pub struct Server {
@@ -28,7 +30,7 @@ pub struct Server {
     #[serde(default = "default_port")]
     pub port: u16,
     /// Proxy configuration to use for connecting to the server.
-    pub proxy: Option<ProxyConfig>,
+    pub proxy: Option<Proxy>,
     /// The password to connect to the server.
     pub password: Option<String>,
     /// The file with the password to connect to the server.
@@ -114,37 +116,6 @@ impl Server {
 
 #[derive(Debug, Clone, Deserialize)]
 #[serde(rename_all = "kebab-case")]
-pub enum ProxyConfigType {
-    Socks5,
-}
-
-#[derive(Debug, Clone, Deserialize)]
-pub struct ProxyConfig {
-    #[serde(rename = "type")]
-    pub proxy_type: ProxyConfigType,
-    pub host: String,
-    pub port: u16,
-    #[serde(default)]
-    pub username: String,
-    #[serde(default)]
-    pub password: String,
-}
-
-impl Into<Proxy> for ProxyConfig {
-    fn into(self) -> Proxy {
-        match self.proxy_type {
-            ProxyConfigType::Socks5 => irc::connection::Proxy::Socks5 {
-                host: self.host,
-                port: self.port,
-                username: self.username,
-                password: self.password,
-            },
-        }
-    }
-}
-
-#[derive(Debug, Clone, Deserialize)]
-#[serde(rename_all = "kebab-case")]
 pub enum IdentifySyntax {
     NickPassword,
     PasswordNick,
@@ -159,7 +130,7 @@ pub enum Sasl {
         /// Account password,
         password: Option<String>,
         /// Account password file
-        password_file: Option<String>
+        password_file: Option<String>,
     },
     External {
         /// The path to PEM encoded X509 user certificate for external auth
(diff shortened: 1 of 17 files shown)


data/src/config/server.rs
```diff
@@ -99,10 +101,44 @@ impl Server {
...
+        match self.proxy_type {
+            ProxyConfigType::Socks5 => irc::connection::Proxy::Socks5 {
+                host: self.host,
+                port: self.port,
+                username: self.username,
+                password: self.password,
+            },
         }
     }
 }
```
tarkah, line 144: We should always implement From instead of Into. Into will then be automatically derived.

data/src/config/server.rs
```diff
@@ -99,10 +101,44 @@ impl Server {
...
+pub struct ProxyConfig {
+    #[serde(rename = "type")]
+    pub proxy_type: ProxyConfigType,
+    pub host: String,
+    pub port: u16,
+    #[serde(default)]
+    pub username: String,
+    #[serde(default)]
+    pub password: String,
+}
```
tarkah, line 131: Let's create a new config::proxy module and put these there. You can then just call it Proxy and Kind (proxy::Kind)

data/src/server.rs
```diff
@@ -54,6 +55,13 @@ impl Map {
         self.0.iter().map(Entry::from)
     }
 
+    pub fn update_proxy(&mut self, proxy: &Option<ProxyConfig>) -> Result<(), Error> {
+        for (_, config) in self.0.iter_mut() {
+            config.proxy = proxy.clone();
+        }
+        Ok(())
+    }
```
tarkah, line 63: Shouldn't a server level proxy config take precedence over a root proxy config? One can simply delete the server proxy to have the root one take effect.

I'd say we only set this if it's None

irc/src/connection.rs
```diff
@@ -26,16 +27,45 @@ pub enum Security<'a> {
...
+                username,
+                password,
+            }) => {
+                let proxy = (host.as_str(), port);
+                if username.trim().is_empty() {
+                    Socks5Stream::connect(proxy, target).await?.into_inner()
+                } else {
+                    Socks5Stream::connect_with_password(proxy, target, &username, &password)
+                        .await?
+                        .into_inner()
```
tarkah, line 65: Does calling into_inner eliminate the proxy? We may need to use an enum to keep the proxy type for it to work

src/screen/dashboard.rs
```diff
@@ -63,7 +64,7 @@ impl Dashboard {
             history: history::Manager::default(),
             last_changed: None,
             command_bar: None,
-            file_transfers: file_transfer::Manager::new(config.file_transfer.clone()),
+            file_transfers: file_transfer::Manager::new(config.file_transfer.clone(), proxy),
```
tarkah, line 67: I think file transfers should be aware of server level proxies as well if one is used over a root proxy?

tarkah (changes requested):
Thanks for tackling this, it looks great! 

I have some comments mostly around dealing with root vs server proxy config, let me know what you think.

irc/Cargo.toml
```diff
@@ -10,6 +10,7 @@ futures = "0.3.28"
 thiserror = "1.0.30"
 tokio = { version = "1.29", features = ["net", "full"] }
 tokio-rustls = { version = "0.26.0", default-features = false, features = ["tls12", "ring"] }
+tokio-socks = "0.5.1"
```
tarkah, line 13: FYI this crate is unmaintained.

We may want to use the `fast_socks5` crate, it seems maintained and built on top of tokio

ramajd: I just updated the `TcpStream` preparation to use the `fast_socks5` crate.

data/src/config/server.rs
ramajd: I don't know if it's a good practice to use the `From`  trait. 

Proper location for the implementation of `From` trait  is `irc::connection` module, while the `irc` crate is not dependent to the `data` crate.

the workaround would be to implement that inside `data::config::proxy` module which might be bad practice to separate the `irc::connection::Proxy` implementations in two separated locations.

tarkah: I'm not saying to relocate anything, keep this but just make it:

```rs
impl From<ProxyConfig> for Proxy
```

Instead of Into. You can do the from implementation and it _gives_ you into for free

data/src/server.rs
ramajd: the current implementation just supports a global proxy configuration. having a server specific proxy config would also be a good suggestion that we can implement and use that with a higher priority.

irc/src/connection.rs
ramajd: as I tested, the `into_inner` provides the `TcpStream` chained with the socks5 proxy.

src/screen/dashboard.rs
ramajd: I didn't get what you mean, could you please provide me more hints about the change you are proposing?

data/src/server.rs
tarkah: I'm confused, we have ProxyConfig specified in the server config, so users can set the proxy in toml for each server. Is that not intended?

src/screen/dashboard.rs
tarkah: Maybe I'm misunderstanding something, see https://github.com/squidowl/halloy/pull/350#discussion_r1578126676

data/src/config/server.rs
tarkah: I've addressed in efd7b54

data/src/server.rs
tarkah: I've fixed in efd7b54 to remove this "injection" and reference the proxy config from a single source of truth

src/screen/dashboard.rs
tarkah: I've fixed in efd7b54 and instead pass the proxy config in when new file transfer tasks are created. This is important because we don't want to hold onto a stale proxy config if the user refreshes their config while the app is running (which we now support).

tarkah (approved):
Thanks @ramajd! I've gone ahead and refactored things a tiny bit so that we always reference the single source of truth proxy config when spawning new server connections or file transfer requests.

This is a bit cleaner than "injecting" the config into the server config map or storing it in the file manager as we now have many sources of truth for this data.

Let me know what you think, but I think this is good to go! Thanks so much for tackling this.

tarkah:
I've also added docs: 

image

ramajd:
@tarkah sorry, I faced an unexpected issue and wasn't available for a couple of days. 

I just reviewed the changes, and seems to be a good refactor over my messy code :sweat_smile: ! 

just one improvement I could add would be to replace the `connect_socks5` with a method for `Proxy` enum which generalizes the proxy-based connection from outside the module: https://github.com/ramajd/halloy/commit/c2338145f38bc987eeee9df31c5cd213ad5b65e4

casperstorm:
> @tarkah sorry, I faced an unexpected issue and wasn't available for a couple of days. 
> 
> 
> 
> I just reviewed the changes, and seems to be a good refactor over my messy code :sweat_smile: ! 
> 
> 
> 
> just one improvement I could add would be to replace the `connect_socks5` with a method for `Proxy` enum which generalizes the proxy-based connection from outside the module: https://github.com/ramajd/halloy/commit/c2338145f38bc987eeee9df31c5cd213ad5b65e4

No worries that happens! Feel free to open improvements PR. Very appreciated.

ramajd:
> Feel free to open improvements PR. Very appreciated.

@casperstorm @tarkah just submitted the improvement as part of a new PR about  HTTP proxy support: #362

PR #352, opened 2024-04-23
Updated banner

YouFoundAlpha:
Makes it more easier to access the IRC channel if people want to look at it without having a IRC client on their computer
If any of you want to change the link you can find more info [here](https://libera.chat/guides/webchat#kiwiirc)


casperstorm:
I rather not do this, but instead make a feature so we open the server in Halloy. I would expect that myself.

YouFoundAlpha:
> I rather not do this, but instead make a feature so we open the server in Halloy. I would expect that myself.

I've reverted the change in the README back to normal and added the correct sized banner to this PR from #353

YouFoundAlpha:
> the new `banner-logo.png` should be in `src/images`. also you have commited a `Test` file.

Will fix

YouFoundAlpha:
> the new `banner-logo.png` should be in `src/images`. also you have commited a `Test` file.

It is fixed

casperstorm (approved):
Looks great, thanks!

PR #357, opened 2024-04-24
Added dependabot

YouFoundAlpha:


.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "cargo"
```
casperstorm, line 6: It would be beneficial to add the following package-ecosystem as well

```
- package-ecosystem: "github-actions"
  directory: "/"
  schedule:
    interval: "weekly"
```

.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
...
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    ignore:
```
casperstorm, line 10: We should ignore `iced` since we are running off a commit.

.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
...
+        # These are peer deps of Cargo and should not be automatically bumped
+        #- dependency-name: "semver"
+        #- dependency-name: "crates-io"
+    rebase-strategy: "disabled"
```
casperstorm, line 14: Why disable this?

casperstorm (changes requested):

.github/dependabot.yml
YouFoundAlpha: I took it from another repo

.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
...
+updates:
+  - package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
+    interval: "weekly"
+    ignore:
+        # These are peer deps of Cargo and should not be automatically bumped
+        - dependency-name: "iced"
+        #- dependency-name: "crates-io"
+    rebase-strategy: "enabled"
```
casperstorm, line 14: You can remove this, as it is enabled by default.

YouFoundAlpha: ok

.github/dependabot.yml
```diff
@@ -0,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
```
casperstorm, line 3: Now you are mixing things up. We want both a package-ecosystem for "github-actions" and one for "cargo".

casperstorm (changes requested):

casperstorm (changes requested):
Slow down a bit, and verify the code you do :-)
Right now it doesn't look like valid YAML, did you validate it?

YouFoundAlpha:
I'll do this later

casperstorm:
> I'll do this later

No need to close the PR, we were so close :)

PR #359, opened 2024-04-24
Removing Cargo.lock from version control

YouFoundAlpha:


tarkah:
Binary crates should always have a lock file checked in. How else would anyone build this without having a completely different outcome?

PR #362, opened 2024-04-27
Features/http proxy support

ramajd:
following from #301 and #350, this PR adds the support for HTTP proxy


casperstorm:
@ramajd could you update the book as well with the new http variable?

ramajd:
@casperstorm just updated the changelog and the book reference for proxy configuration

tarkah (approved):
I don't have a proxy to test with, but LGTM. Thanks for adding these :D

ramajd:
@tarkah I tested both, since I have to use various proxies almost everyday to bypass the internet censorship.

PR #363, opened 2024-04-28
feat: url schemes

casperstorm:
Ability to open "irc" and "ircs" url links.

- [x] Test
  - [x] Windows
  - [x] Linux
  - [x] MacOS
- [x] Add approval screen to accept opening a new connection
- [X] Add ipc to route url
- [X] Update book 
- [X] Update changelog


data/src/log.rs
```diff
@@ -30,8 +30,8 @@ fn path() -> Result<PathBuf, Error> {
...
     Io(#[from] io::Error),
-    #[error(transparent)]
-    SetLog(#[from] log::SetLoggerError),
-    #[error(transparent)]
-    ParseLevel(#[from] log::ParseLevelError),
-}
+    // #[error(transparent)]
+    // SetLog(#[from] log::SetLoggerError),
+    // #[error(transparent)]
+    // ParseLevel(#[from] log::ParseLevelError),
```
casperstorm, line 36: I am running into errors when running tests if these are added. 🤔

casperstorm:
@tarkah can you take this for a spin and see if you wanna add or change anything now?

tarkah (approved):
Awesome addition @casperstorm! 

I've cleaned up a few things, made sure that URL is handled if halloy is booted from it the first time (not already running), and also handled the case of an existing server only joining requested channels vs overriding the server w/ a new one.

I've tested on linux and it works great. Let me know if you see anything else, otherwise I think we're good to merge.

Can't wait to add `halloy://` urls for theme sharing in the future :sunglasses:

Issue #331, opened 2024-04-10
Auto-reconnect after /QUIT


ColinMcKechney:
Not sure if this is intended behavior or not, but if reconnect_delay is set you auto-reconnect to a server even after using the /QUIT command.

casperstorm:
This sounds indeed like a bug! Thanks for reporting it, @ColinMcKechney.

PR #365, opened 2024-04-30
Leave Server when Sending a `/QUIT` Command

andymandias:
This is a small change to address #331.  Checks sent user input for a `/QUIT` command, then executes the same code path that is used by the "Leave server" context menu button in that scenario.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -173,7 +173,7 @@ impl Halloy {
         )
     }
 
-    fn quit_server(&mut self, server: Server) -> Command<Message> {
+    fn quit_server(&mut self, server: Server, reason: Option<String>) -> Command<Message> {
         // Removing from servers kills stream subscription
         self.servers.remove(&server);
 
@@ -181,7 +181,7 @@ impl Halloy {
         self.clients
             .remove(&server)
             .map(move |connection| async move {
-                connection.quit().await;
+                connection.quit(reason).await;
             })
             .map(|task| Command::perform(task, |_| Message::QuitServer(server)))
             .unwrap_or_else(Command::none)
@@ -269,7 +269,7 @@ impl Application for Halloy {
                                 self.config = updated;
 
                                 for server in removed_servers {
-                                    commands.push(self.quit_server(server));
+                                    commands.push(self.quit_server(server, None));
                                 }
                             }
                             Err(error) => {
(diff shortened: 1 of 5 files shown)


src/main.rs
```diff
@@ -569,6 +569,11 @@ impl Application for Halloy {
 
                     Command::batch(commands)
                 }
+                stream::Update::QuitServer(server, reason) => {
+                    let _ = self.quit_server(server, reason);
+
+                    Command::none()
+                }
```
tarkah, line 576: I don't love that we have changed behavior here because we _don't_ return a command.

tarkah (commented):
Thanks for looking at this @andymandias! Let me think on this one for a little bit, because I believe there's a race condition between us removing the stream subscription (`self.servers.remove(..)`) which kills the stream backend and us trying to send the quit message.

I think the better approach is we don't remove the server from `self.servers` (thus dropping the stream), until we get a confirmation from the stream backend that QUIT was sent.

We should then also have that stream backend enter a Quit state so it can't do anything else until we have processed the confirmation on the frontend and removed it from `self.servers`. I think this will be more robust.

src/main.rs
andymandias: Good point, I think we can just return `self.quit_server(server, reason)` here, which would make more sense.  Works in testing so I'll push that change.  Apologies if I'm misunderstanding the issue.

andymandias:
I tried to be minimal in messing with the particulars within `quit_server`, as I didn't know the reasoning that went into its processes.  Sounds like a good plan to me, thank you for improving on the changes as usual.

You probably already have this in mind, but I will note that in testing some servers do not respond with proper `QUIT` acknowledgement (in particular UnrealIRCd sends a message that's terminated with crcrlf and as such gets caught in a parse error).

tarkah:
@andymandias Checkout my commit. I think this flow makes WAY more sense...

1. Client::send now just sends the QUIT command to the backend
1. We intercept QUIT command from the backend and we send it then forward a Quit update to the frontend
1. Backend transitions to State::Quit and does nothing (until frontend drops this stream)
1. On receiving Update::Quit from the frontend, we can safely remove the server / client (dropping the stream) and then broadcast that we quit

andymandias:
Hadn't quite grasped what you meant previously, but looking at the code it's perfectly legible.  Looks much nicer @tarkah, and I agree makes much more sense.  And it works perfectly in my testing (as expected).  Thank you for the help with this!

tarkah (approved):

PR #366, opened 2024-04-30
Message Tags Support

andymandias:
Reading through [`message-tags`](https://ircv3.net/specs/extensions/message-tags) and looking at `irc/proto/src/parse.rs` I believe tags are already rigorously parsed by the current implementation.  I believe all we need in addition to that is support for the `CLIENTTAGDENY` ISUPPORT parameter and `TAGMSG` command.

In the [client-only tags registry](https://ircv3.net/registry#tags) there appears to be one active and three draft client-only tags.  If/when we implement those features I believe we should disable them if they're covered in the `CLIENTTAGDENY` ISUPPORT parameter, but as it stands (with none of those features implemented, yet) I think parsing `CLIENTTAGDENY` (added in #349) should be sufficient.

This PR adds the `TAGMSG` command.  As with `CLIENTTAGDENY` there's nothing for us to utilize `TAGMSG` for yet, so I think its addition is all we currently need.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ Added:
 - Allow configuration of internal messages in buffer (see [buffer configuration](https://halloy.squidowl.org/configuration/buffer.html#bufferinternal_messages-section))
 - User information added to context menu
 - Support for IRCv3 `CAP NEW` and `CAP DEL` subcommands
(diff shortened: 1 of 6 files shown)


tarkah (approved):

tarkah:
Makes sense, thanks!

casperstorm (approved):

Issue #361, opened 2024-04-26
Add `profile.release` to enable link time optimizations


rawhuul:
Hi, the default `profile.release` for `Cargo.toml` is:

tarkah:
Hey @rawhuul thanks for the suggestion! I'd like to see us create a special profile specifically called `packaging` that can use these optimizations.

I use `release` profile a lot while developing and LTO will really kill compile times there, so I think we should only enable when "packaging" the app for release

rawhuul:
Yup, you are right. This can be achieved easily, this feature is stable from `rustc v1.57`

```toml
[profile.packaging]
inherits = "release"
lto = true
```

hashcatHitman:
Is this still something that is wanted? I don't see any clear reason #367 was closed, and I'd be happy to make a quick PR if this is still something we want.

auronandace:
Does #325 have an impact here?

PR #367, opened 2024-05-05
Added packaging profile for release builds

rawhuul:
As discussed in issue #361


casperstorm (approved):
@tarkah any feedback?

Cargo.toml
```diff
@@ -52,3 +52,14 @@ members = ["data", "irc", "irc/proto"]
...
+
+[profile.packaging]
+opt-level = 3
+debug = false
+lto = "fat"
+strip = "debuginfo"
+panic = 'unwind'
+incremental = false
+codegen-units = 1
+rpath = false
```
tarkah, line 64: We don't need to redefine the items that are identical to the `release` profile since we use `inherits` below. Let's remove them.

tarkah (changes requested):
We should leverage this profile in our release scripts that produce the final binary.

PR #370, opened 2024-05-06
`chathistory` Support

andymandias:
Work in progress PR to add IRCv3 CHATHISTORY support (i.e. #206).  Since it's not a widely available feature yet, the goal is to minimize the any effects on Halloy's operation when CHATHISTORY is not available.  Testing on ircs://irc.ergo.chat:6697/ to start.

Currently (more-or-less) implemented:
- Request up to the latest 500 messages or all messages since latest message in channel history when joining a channel.
- Request up to 500 messages to add to channel history when scrolling to the top of a channel buffer.
- Basic message deduplication.

 Planned:
- [x] Enable `event-playback` to allow replay of channel events (`JOIN`/`PART`/`QUIT`/etc).  `HistServ` appears to be non-standard, so all its messages are filtered out for now for parity of experience.  ~~Convert `HistServ` PRIVMSG messages into the appropriate `JOIN`/`PART`/`QUIT`/etc messages.~~
- [x] Ensure history is loaded before requesting latest messages after `JOIN`.
- [x] Ensure messages received when starting Halloy are marked as new.
- [x] Request messages when reconnecting to a server (without closing Halloy; current testing shows this is covered by `JOIN` messages request(s)). 
- [x] Improve UX for loading additional channel history (button at end of history & config option to automatically request when scrolling to top of history).
- [x] Do not move the backlog divider when adding older messages to a buffer.
- [x] Test multi-request `LATEST` functionality (artificially limit the size of batches).
- [x] Improve performance of deduplication (utilize something like an [`indexmap`](https://docs.rs/indexmap/latest/indexmap/map/struct.IndexMap.html) for message history?).  And/or, avoid the need for deduplication.
- [x] Use `TARGETS` to open Query buffers for direct messages sent while disconnected.
- [x] Add get latest messages & load older messages to Query buffers.
- [x] Get message ids for user-sent messages.
- [x] Set up logic for the possibility of a `chathistory` request timeout.

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -1499,6 +1499,12 @@ impl Client {
         isupport::MessageReferenceType::default()
     }
 
+    pub fn chathistory_request(&self, target: &str) -> Option<ChatHistoryRequest> {
+        self.chanmap
+            .get(target)
+            .and_then(|channel| channel.chathistory_request.clone())
+    }
+
     pub fn send_chathistory_request(
         &mut self,
         subcommand: ChatHistorySubcommand,
@@ -1535,7 +1541,7 @@ impl Client {
                                             server_time.checked_sub_signed(time_delta)
                                         })
                                 {
-                                    MessageReference::Timestamp(fuzzed_server_time, "".to_string())
+                                    MessageReference::Timestamp(fuzzed_server_time, ":".to_string())
                                 } else {
                                     message_reference
                                 }
@@ -1575,7 +1581,7 @@ impl Client {
                                     {
                                         MessageReference::Timestamp(
                                             fuzzed_reference_timestamp,
-                                            "".to_string(),
+                                            ":".to_string(),
                                         )
                                     } else {
                                         message_reference
@@ -1607,6 +1613,13 @@ impl Client {
         if let Some(channel) = self.chanmap.get_mut(target) {
             channel.chathistory_request = None;
         }
+    }
+
+    pub fn chathistory_before_exhausted(&self, target: &str) -> bool {
+        self.chanmap
+            .get(target)
+            .map(|channel| channel.chathistory_before_exhausted)
+            .unwrap_or_default()
     }
 
     fn topic<'a>(&'a self, channel: &str) -> Option<&'a Topic> {
@@ -1853,6 +1866,15 @@ impl Map {
             .unwrap_or_default()
     }
 
+    pub fn get_channel_chathistory_request(
+        &self,
+        server: &Server,
+        channel: &str,
+    ) -> Option<ChatHistoryRequest> {
+        self.client(server)
+            .and_then(|client| client.chathistory_request(channel))
+    }
+
     pub fn send_channel_chathistory_request(
         &mut self,
         subcommand: ChatHistorySubcommand,
@@ -1869,6 +1891,12 @@ impl Map {
         if let Some(client) = self.client_mut(server) {
             client.clear_chathistory_request(channel);
         }
+    }
+
+    pub fn get_channel_chathistory_before_exhausted(&self, server: &Server, channel: &str) -> bool {
+        self.client(server)
+            .map(|client| client.chathistory_before_exhausted(channel))
+            .unwrap_or_default()
     }
 
     pub fn get_channels<'a>(&'a self, server: &Server) -> &'a [String] {
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -1,4 +1,4 @@
-use chrono::{DateTime, TimeDelta, Utc};
+use chrono::{DateTime, Utc};
 use futures::channel::mpsc;
 use irc::proto::{self, command, Command};
 use itertools::{Either, Itertools};
@@ -337,6 +337,7 @@ impl Client {
                                                         None
                                                     }
                                                 }
+                                                ChatHistorySubcommand::Targets(_, _, _) => None,
                                             };
 
                                             finished.events.push(
@@ -1494,16 +1495,8 @@ impl Client {
 
             match subcommand {
                 ChatHistorySubcommand::Latest(target, message_reference, limit) => {
-                    let command_message_reference = match message_reference {
-                        MessageReference::Timestamp(server_time) => {
-                            TimeDelta::try_seconds(isupport::CHATHISTORY_FUZZ_SECONDS)
-                                .and_then(|time_delta| server_time.checked_sub_signed(time_delta))
-                                .map_or(message_reference, |fuzzed_server_time| {
-                                    MessageReference::Timestamp(fuzzed_server_time)
-                                })
-                        }
-                        _ => message_reference,
-                    };
+                    let command_message_reference =
+                        isupport::fuzz_start_message_reference(message_reference);
 
                     log::debug!(
                         "[{}] requesting {limit} latest messages in {target} since {}",
@@ -1520,16 +1513,8 @@ impl Client {
                     ));
                 }
                 ChatHistorySubcommand::Before(target, message_reference, limit) => {
-                    let command_message_reference = match message_reference {
-                        MessageReference::Timestamp(server_time) => {
-                            TimeDelta::try_seconds(isupport::CHATHISTORY_FUZZ_SECONDS)
-                                .and_then(|time_delta| server_time.checked_add_signed(time_delta))
-                                .map_or(message_reference, |fuzzed_server_time| {
-                                    MessageReference::Timestamp(fuzzed_server_time)
-                                })
-                        }
-                        _ => message_reference,
-                    };
+                    let command_message_reference =
+                        isupport::fuzz_end_message_reference(message_reference);
 
                     log::debug!(
                         "[{}] requesting {limit} messages in {target} before {}",
@@ -1551,52 +1536,11 @@ impl Client {
                     end_message_reference,
                     limit,
                 ) => {
-                    let (command_start_message_reference, command_end_message_reference) = match (
-                        start_message_reference.clone(),
-                        end_message_reference.clone(),
-                    ) {
-                        (
-                            MessageReference::Timestamp(start_server_time),
-                            MessageReference::Timestamp(end_server_time),
-                        ) => {
-                            if start_server_time < end_server_time {
-                                (
-                                    TimeDelta::try_seconds(isupport::CHATHISTORY_FUZZ_SECONDS)
-                                        .and_then(|time_delta| {
-                                            start_server_time.checked_sub_signed(time_delta)
-                                        })
-                                        .map_or(start_message_reference, |fuzzed_server_time| {
-                                            MessageReference::Timestamp(fuzzed_server_time)
-                                        }),
-                                    TimeDelta::try_seconds(isupport::CHATHISTORY_FUZZ_SECONDS)
-                                        .and_then(|time_delta| {
-                                            end_server_time.checked_add_signed(time_delta)
-                                        })
-                                        .map_or(end_message_reference, |fuzzed_server_time| {
-                                            MessageReference::Timestamp(fuzzed_server_time)
-                                        }),
-                                )
-                            } else {
-                                (
-                                    TimeDelta::try_seconds(isupport::CHATHISTORY_FUZZ_SECONDS)
-                                        .and_then(|time_delta| {
-                                            start_server_time.checked_add_signed(time_delta)
-                                        })
-                                        .map_or(start_message_reference, |fuzzed_server_time| {
-                                            MessageReference::Timestamp(fuzzed_server_time)
-                                        }),
-                                    TimeDelta::try_seconds(isupport::CHATHISTORY_FUZZ_SECONDS)
-                                        .and_then(|time_delta| {
-                                            end_server_time.checked_sub_signed(time_delta)
-                                        })
-                                        .map_or(end_message_reference, |fuzzed_server_time| {
-                                            MessageReference::Timestamp(fuzzed_server_time)
-                                        }),
-                                )
-                            }
-                        }
-                        _ => (start_message_reference, end_message_reference),
-                    };
+                    let (command_start_message_reference, command_end_message_reference) =
+                        isupport::fuzz_message_reference_range(
+                            start_message_reference,
+                            end_message_reference,
+                        );
 
                     log::debug!(
                         "[{}] requesting {limit} messages in {target} between {} and {}",
@@ -1609,6 +1553,32 @@ impl Client {
                         "CHATHISTORY",
                         "BETWEEN",
                         target,
+                        command_start_message_reference.to_string(),
+                        command_end_message_reference.to_string(),
+                        limit.to_string()
+                    ));
+                }
+                ChatHistorySubcommand::Targets(
+                    start_message_reference,
+                    end_message_reference,
+                    limit,
+                ) => {
+                    let (command_start_message_reference, command_end_message_reference) =
+                        isupport::fuzz_message_reference_range(
+                            start_message_reference,
+                            end_message_reference,
+                        );
+
+                    log::debug!(
+                        "[{}] requesting {limit} targets between {} and {}",
+                        self.server,
+                        command_start_message_reference,
+                        command_end_message_reference,
+                    );
+
+                    let _ = self.handle.try_send(command!(
+                        "CHATHISTORY",
+                        "TARGETS",
                         command_start_message_reference.to_string(),
                         command_end_message_reference.to_string(),
                         limit.to_string()
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -75,7 +75,8 @@ pub enum Broadcast {
 
 #[derive(Debug, Clone)]
 pub enum HistoryRequest {
-    Targets,
+    Targets(DateTime<Utc>),
+    Queries(Vec<isupport::MessageReferenceType>, DateTime<Utc>),
     Recent(String, Vec<isupport::MessageReferenceType>, DateTime<Utc>),
     Older(String, Vec<isupport::MessageReferenceType>),
 }
@@ -587,9 +588,15 @@ impl Client {
                 if caps.contains(&"chathistory") || caps.contains(&"draft/chathistory") {
                     self.supports_chathistory = true;
 
-                    return Some(vec![Event::ChatHistoryRequestFromHistory(
-                        HistoryRequest::Targets,
-                    )]);
+                    return Some(vec![
+                        Event::ChatHistoryRequestFromHistory(HistoryRequest::Targets(server_time(
+                            &message,
+                        ))),
+                        Event::ChatHistoryRequestFromHistory(HistoryRequest::Queries(
+                            self.chathistory_message_reference_types(),
+                            server_time(&message),
+                        )),
+                    ]);
                 }
             }
             Command::CAP(_, sub, a, b) if sub == "NAK" => {
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -1,4 +1,4 @@
-use chrono::{DateTime, Utc};
+use chrono::{DateTime, TimeDelta, Utc};
 use futures::channel::mpsc;
 use irc::proto::{self, command, Command};
 use itertools::{Either, Itertools};
@@ -998,7 +998,7 @@ impl Client {
 
                         if self.supports_chathistory {
                             return Some(vec![Event::ChatHistoryCommand(
-                                ChatHistorySubcommand::Latest,
+                                ChatHistorySubcommand::Latest(server_time(&message)),
                                 channel.clone(),
                                 self.get_chathistory_message_reference_type(),
                                 self.get_chathistory_limit(),
@@ -1413,40 +1413,35 @@ impl Client {
     pub fn finish_chathistory_batch(&mut self, batch: &mut Batch) {
         if let Some(chathistory_target) = &batch.chathistory_target {
             if let Some(channel) = self.chanmap.get_mut(chathistory_target) {
-                let limited_latest_batch = if let Some(ChatHistoryRequest {
+                if let Some(ChatHistoryRequest {
                     subcommand,
                     message_reference,
                     limit,
-                }) = &channel.chathistory_request
+                }) = channel.chathistory_request.clone()
                 {
-                    let subcommand_is_latest_since = *subcommand == ChatHistorySubcommand::Latest
-                        && !matches!(message_reference, MessageReference::None);
-
-                    if matches!(*subcommand, ChatHistorySubcommand::After)
-                        || subcommand_is_latest_since
+                    if matches!(subcommand, ChatHistorySubcommand::Latest(_))
+                        && !matches!(message_reference, MessageReference::None)
                     {
                         batch.events.reverse();
+
+                        if batch.events.len() == limit as usize {
+                            let message_reference_type =
+                                self.get_chathistory_message_reference_type().clone();
+                            let limit = self.get_chathistory_limit();
+
+                            batch.events.push(Event::ChatHistoryCommand(
+                                subcommand.clone(),
+                                chathistory_target.to_string(),
+                                message_reference_type,
+                                limit,
+                            ));
+
+                            return;
+                        }
                     }
-
-                    subcommand_is_latest_since && batch.events.len() == *limit as usize
-                } else {
-                    false
-                };
-
-                if limited_latest_batch {
-                    let message_reference_type =
-                        self.get_chathistory_message_reference_type().clone();
-                    let limit = self.get_chathistory_limit();
-
-                    batch.events.push(Event::ChatHistoryCommand(
-                        ChatHistorySubcommand::Latest,
-                        chathistory_target.to_string(),
-                        message_reference_type,
-                        limit,
-                    ));
-                } else {
-                    channel.chathistory_request = None;
                 }
+
+                channel.chathistory_request = None;
             }
         }
     }
@@ -1729,11 +1724,11 @@ impl Map {
             if let Some(channel) = client.chanmap.get_mut(target) {
                 if client.supports_chathistory
                     && (channel.chathistory_request.is_none()
-                        || (subcommand == ChatHistorySubcommand::Latest
+                        || (matches!(subcommand, ChatHistorySubcommand::Latest(_))
                             && matches!(
                                 channel.chathistory_request,
                                 Some(ChatHistoryRequest {
-                                    subcommand: ChatHistorySubcommand::Latest,
+                                    subcommand: ChatHistorySubcommand::Latest(_),
                                     ..
                                 })
                             )))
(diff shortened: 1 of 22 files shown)


tarkah:
@andymandias I'll look to review this very soon :eyes:

andymandias:
Thanks @tarkah!  It's definitely still a work in progress, in particular query buffers don't have `chathistory` support yet.  There are also a number of smaller issues to work out (will try to add these to the list at the top of the PR).  I am currently using it with `soju` (timestamp-based) and ircs://ergo.chat/#ergo (msgid-based) and it ~works.

Things I think particularly need reviewing:
- I'm committing some crimes against asynchronicity with `load_history_now` and `make_partial_now` (starting in `main.rs`, going into `manager.rs`).  I believe that functionality should instead be implemented by a chain of asynchronous commands (load history → get latest message), but I wasn't quite sure how to set up the chain.  I set it aside to deal with later, but I suspect you may have an even better solution.
- It's useful to have a `Full` history in order to have a clear reference message when adding messages to the history.  I've been working on not making that necessary, but in the meantime I've messed with the `make_partial` to allow from converting from a `Full`-with-unread-messages to a `Partial` history.  Done my best to respect the existing structure, but it probably warrants review.

andymandias:
Brief addendum to note it looks like my next push will involve a moderate restructuring, so probably best to review after (hopefully tonight).

andymandias:
I'll be chipping away at this still, but it's probably in a good spot for a WIP review now.

src/screen/dashboard.rs
```diff
@@ -128,105 +128,177 @@ impl Dashboard {
...
+                                }
+                            }
+                            Some(buffer::Event::ChatHistoryBeforeRequest) => {
+                                if let Some(buffer) = pane.buffer.data() {
+                                    let server = buffer.server();
+
+                                    if clients.get_server_supports_chathistory(server) {
+                                        if let Some(target) = buffer.target() {
+                                            let message_reference_type = clients
+                                                .get_server_chathistory_message_reference_type(
```
tarkah, line 280: We can consider `infinite_scroll` within the `scroll_view` and conditionally return the event or not. Then we can define a single event variant `Event::RequestMoreChatHistory` that's used from both the button message & scroll to top message and dedupe all this code.

src/screen/dashboard.rs
```diff
@@ -128,105 +128,177 @@ impl Dashboard {
...
+                                                ChatHistorySubcommand::Before,
+                                                server,
+                                                &target,
+                                                oldest_message_reference,
+                                            );
+                                        }
+                                    }
+                                }
+                            }
+                            _ => (),
```
tarkah, line 301: `_` is dangerous as the compiler won't notify us when new variants are added. Match on `None` instead to fully exhaust the match

src/screen/dashboard.rs
```diff
@@ -741,14 +813,44 @@ impl Dashboard {
...
+                                        ChatHistorySubcommand::Before,
+                                        server,
+                                        &target,
+                                        oldest_message_reference,
+                                    );
+                                }
+                            }
+                        }
+                    }
+                }
```
tarkah, line 844: 3rd place this logic is defined. Time to refactor it into a helper method :)

src/screen/dashboard.rs
andymandias: Hadn't thought about it that way before, thanks for the suggestion/tip :)

tarkah:
I'm a bit unsure about all the API changes to `history` to support this feature. I need to internalize it all more, but it seems to me we shouldn't have so many new code paths. All we're doing is feeding messages to `history` like we would any other message that comes in, the only difference is we may want to splice it into place vs appending it.

So it seems to me the only API change of `history` needed would be the following:

- Did the message come in via history?
- If yes, splice it into place using message reference

It'd be great to sync up some time to discuss this over IRC so I can better understand the model at a high level as there's a lot to internalize. I'm also seeing async functions getting called in from `main` but w/out awaiting which is a `no-op`.

data/src/client.rs
```diff
@@ -312,6 +471,25 @@ impl Client {
...
+                        // We require batch for our chathistory support
+                        let requesting_chathistory = if contains_starting_with("chathistory") {
+                            requested.push("chathistory");
+                            true
+                        } else if contains_starting_with("draft/chathistory") {
+                            requested.push("draft/chathistory");
+                            true
+                        } else {
+                            false
+                        };
```
tarkah, line 484: Why `contains_starting_with` when we then insert is as the full value? We need to ensure we're requesting the exact same capability that we checked existed. Why doesn't just `contains ` work here since we check both w/ and w/out `draft` prefix?

data/src/client.rs
```diff
@@ -243,7 +334,72 @@ impl Client {
...
+                    .as_ref()
+                    .and_then(|batch| self.batches.get(batch))
+                    .and_then(|batch| batch.chathistory_target.clone())
+                {
+                    if let Some(ChatHistoryRequest {
+                        subcommand,
+                        message_reference,
+                        ..
+                    }) = self.chathistory_requests.get(&target)
+                    {
```
tarkah, line 347: You can collapse these two if let branches together by calling `self.chathistory_requests.get` in an `and_then` and pattern matching the `Some(ChatHistoryRequest` in the root if let

data/src/client.rs
```diff
@@ -243,7 +334,72 @@ impl Client {
...
+                                        subcommand.clone(),
+                                        message_reference.clone(),
+                                    )]
+                                }
+                                _ => vec![Event::ChatHistorySingle(
+                                    message,
+                                    self.nickname().to_owned(),
+                                    subcommand.clone(),
+                                    message_reference.clone(),
+                                )],
```
tarkah, line 394: I assume we're just forwarding all messages within here to history and not calling `handle` as we don't want to update local state based on history? That makes sense.

data/src/isupport.rs
```diff
@@ -630,9 +649,43 @@ pub struct CommandTargetLimit {
 }
 
 #[derive(Clone, Debug)]
+pub enum MessageReference {
+    Timestamp(DateTime<Utc>, String),
```
tarkah, line 653: What's the 2 field / string value of the `Timestamp` variant here? I don't see that documented anywhere here:

https://ircv3.net/specs/extensions/chathistory

This seems to be our own internal representation of a message reference _value_. I don't think this should live in `isupport`, but probably `message::Reference`

data/src/history/manager.rs
```diff
@@ -188,6 +190,98 @@ impl Manager {
...
+            history::Kind::Channel(target.to_string())
+        } else {
+            history::Kind::Query(target.to_string().into())
+        };
+
+        self.data.map.get_mut(&server).and_then(|map| {
+            map.get_mut(&kind)
+                .and_then(|history| history.make_partial(message_reference))
+        });
+    }
```
tarkah, line 244: What are these `_now` methods used for? They're async but we're calling them from a blocking thread so we aren't actually doing anything.

data/src/history.rs
```diff
@@ -238,9 +396,30 @@ impl History {
...
+                            std::cmp::min(earliest_received_at, unread_message.received_at)
+                        },
+                    )
+                } else {
+                    Posix::now()
+                };
                 let messages = std::mem::take(messages);
 
-                *self = Self::partial(server.clone(), kind.clone(), Posix::now());
+                *self = Self::partial(server.clone(), kind.clone(), unread_messages, opened_at);
```
tarkah, line 422: I'm not sure I follow what `unread_messages` and `opened_at` means here when we convert from a full "opened" buffer history to a partial "closed" buffer history. 

Why are we calling `make_partial` from a chat history event? The idea behind our history API and full vs partial is that these are full / partial based on if the underlying buffer is open / closed.

- Partial exists as a way to store messages received on a closed buffer which are then fully flushed to disk.
- Full for open buffers has all messages in it and we append new messages. When a buffer is closed, we flush this entire history to disk.

This lifecycle is managed by `Manager::track` which is called after processing any `Dashboard` messages, where we see which set of buffers are open.

Shouldn't chathistory messages just feed in to the existing full / partial history that exists for a buffer? I don't think chathistory should break this lifecycle.

data/src/client.rs
andymandias: I thought I had run into a server that listed the `chathistory` capability with a value attached (e.g. `draft/chathistory=1000`), and I used `contains_starting_with` to sidestep having an issue with it, but I can't find any server where that's the case now.  Will switch it back to `contains` since that's my reading of the specification.

data/src/client.rs
andymandias: Correct.  Particularly since we're requesting `draft/event-playback` when available there can be messages that would mess with local state.

data/src/history.rs
```diff
@@ -171,6 +188,144 @@ impl History {
         }
     }
 
+    fn add_chathistory_message(
+        &mut self,
+        message: Message,
+        subcommand: ChatHistorySubcommand,
+        message_reference: MessageReference,
+    ) {
+        match self {
```
tarkah, line 197: I haven't tested it, but I think we can simplify this and expose a single `add_message` API with the following. This is also non-dependent on `ChatHistorySubcommand`, not sure why that's needed here for splicing the message in.

We basically splice messages in `server_time` order, allowing us to `binary_search` on that time which should be pretty fast.

Binary search returns a `Result<usize, usize>` where the Ok value means an identical `Ordering::Equal` match was found, so we can swap out that value with the incoming value. 

This is 100% safe when we use the `Lookup::MessageId` since we return `Equal` if the id's match, otherwise the binary search continues based on server time.

For `Lookup::Timestamp` this means the message has an identical timestamp. Since `servertime` extension mandates milisecond accuracy, and if we don't have servertime, we use `Utc::now` which is nanosecond accuracy, I think this is a strong match. I doubt we will have too many false positives. 

The Err value means no match was found, but the position returns is the index where the element can be safely inserted to preserve ordering (server_time), so we can just safely insert it.

```rs
    fn add_message(&mut self, message: Message, reference: Option<MessageReference>) {
        if let History::Partial {
            unread_message_count,
            ..
        } = self
        {
            if message.triggers_unread() {
                *unread_message_count += 1;
            }
        }

        match self {
            History::Partial {
                messages,
                last_received_at,
                ..
            }
            | History::Full {
                messages,
                last_received_at,
                ..
            } => {
                enum Lookup {
                    Timestamp(DateTime<Utc>),
                    MessageId(String),
                }

                let lookup = match reference {
                    Some(MessageReference::Timestamp(timestamp, _)) => Lookup::Timestamp(timestamp),
                    Some(MessageReference::MessageId(id)) => Lookup::MessageId(id),
                    Some(MessageReference::None) | None => {
                        if let Some(id) = message.id.clone() {
                            Lookup::MessageId(id)
                        } else {
                            Lookup::Timestamp(message.server_time)
                        }
                    }
                };

                let search = match lookup {
                    Lookup::Timestamp(timestamp) => {
                        messages.binary_search_by(|stored| stored.server_time.cmp(&timestamp))
                    }
                    Lookup::MessageId(id) => messages.binary_search_by(|stored| {
                        if stored.id.as_ref() == Some(&id) {
                            Ordering::Equal
                        } else {
                            stored.server_time.cmp(&message.server_time)
                        }
                    }),
                };

                match search {
                    // Identical value found, replace it
                    //
                    // Message Id is 100% match
                    // Timestamp (server time) is milisecond accuracy so
                    // a very strong match
                    Ok(pos) => {
                        messages.splice(pos..pos + 1, Some(message));
                    }
                    // No match found, insert it at pos (server time sorted order)
                    Err(pos) => messages.insert(pos, message),
                }

                *last_received_at = Some(Instant::now());
            }
        }
    }
```

data/src/isupport.rs
andymandias: I think this warrants discussion, particularly since I'm not 100% convinced the way I've implemented this is what we want long term.  Difficulties arise with `Timestamp` based `chathistory` since the `DateTime<Utc>` is not unique identifier.  All messages within the same second will have the same timestamp, and while that doesn't happen all the time it happens enough (especially since we're getting `JOIN`/`QUIT`/`PART`/etc messages as part of our requests).  Also, because of that lumping together, we need to request messages with a server time before the last message we saw, otherwise we might miss messages that happened in the same interval but the client didn't receive.  That means we will basically have at least one duplicate message per chathistory request when using `Timestamp`.  In order to perform deduplication and proper stitching together of message histories we need to be able to distinguish between messages with the same timestamp, which is where the `String` comes in.

Originally I had the `String` be the `text` of the message, but that has obvious issues when the message text depends on the current state (i.e. for `JOIN`/`QUIT`/etc messages where the user in `text` will be represented by what we currently know about them).  So, I switched the second bit of information to be a `client_id`, which is a seahash of the `Encoded` that comes in.  I don't love this really, since changing the structure of `Encoded` will cause ids to shift, but it works better than using `text`.

Client ids are distinguished from server ids by starting with `:`, which is an invalid character in a message tag (which is how server tags are delivered).

data/src/history/manager.rs
andymandias: My understanding is that the [`#[tokio::main]`](https://docs.rs/tokio/latest/tokio/attr.main.html) allows them to run ~synchronously.  The histories are loaded, that much I am certain of.  Regardless, this is the crime against asynchronicity I was referring to earlier, and is a hack that's not really intended to stay.  The ultimate goal of this is to load the history such that the history can be used to (a) find a reference message to base the chathistory request on and (b) perform any necessary deduplication on received messages.  I've been working on reducing the need for (b) with the aim of ultimately removing that need entirely, but I haven't tested it lately to see if the need is gone yet.  I presume that (a) could be done more elegantly than I'm doing here, but I wanted to hash out some of the basic chathistory functionality before trying to work that out.

data/src/history.rs
andymandias: This is mainly a continuation of the hack above, and is the second thing I was trying to highlight needed review.  I don't disagree with anything you're saying here, crimes are being committed which need to be rectified.  But let me lay out the functionality that the crimes provide because I'm not yet sure how else to get at it.

When joining a channel after launch, we want to get all messages since we last closed Halloy.  To do that properly, we need to know what the last message we received for that channel was, but we can't do that with having loaded the message history for the channel.  We can request the `LATEST` however many messages, but there will often be duplicates in there we don't really want to process and we won't know if they're duplicates until we can compare them to the message history.  That messes with our ability to tell the user whether the channel has any new messages or not, and I wasn't able to see any way around it.  The process of loading a channel would also become more complicated in that scenario (deduplication would need to be done on channel load, along with any additional requests needed to fill in chat history that wasn't covered by the initial request), but that part is doable.

So, the hack that's being performed now is to load the history for every channel joined, use the loaded history to form the chathistory request for the latest messages, position received messages, and deduplicate, then make the loaded history partial.  All this junk right here is a way to convert a loaded history with all the chathistory request messages added into a proper partial history that we'd see if the chathistory messages had been loaded into a proper, fresh partial history.

data/src/isupport.rs
andymandias: I *swear* I was seeing second rounded server times, but I must have screwed up a debug print somewhere because I see proper millisecond times looking at it now.

data/src/history.rs
andymandias: I had considered something like this (was thinking of using a `BTreeMap`), but I was worried about the overhead it would put on the non-chathistory variant of `add_message`.

I'll put this in and test it.

tarkah: Binary search on 10k messages will be pretty quick, that's like 13-14 iterations before you've exhausted the search. 

However, I think I need to make it more robust. I have some ideas and will dig in more tomorrow.

andymandias: I believe we'd need to sort the messages by `server_time` or `id` for this to work.  Sorting by `server_time` should be okay, that's more or less how we expect messages to be already, but we have no guarantee that sorting by `id` will match the ordering given by `server_time`.

data/src/isupport.rs
andymandias: Pulling out the `client_id` business for now.  Will push that along with other changes to match our discussion tonight.

data/src/history.rs
tarkah: Correct, the binary search will insert things in server time sorted order.

I plan to do 2 binary search lookups +- N duration from current message server time.

I can then iterator through those messages and check for an id match. If an id match occurs, replace. If no id match, I can insert after the last position where server time <= current message.

This allows us to handle checking messages with duplicate server times, which the binary search function on its own doesn't handle.

We can also check user + content as well if no id exists for duplicate.

andymandias: Another consideration the specification brings up is that messages do not have to be ordered by server time in the server message log.  This is a reason supporting their suggestion to use a "fuzz interval" when requesting new messages via timestamp.  This means client-side message ordering *might* not match server side.  I don't know that we want to do anything to accommodate that (beyond adding the fuzz interval to our `Timestamp` requests, which we currently do).

andymandias: Content is unfortunately tricky to check for matching for event messages, in my experience, but works well otherwise in my experience.

tarkah: If we just blindly append like previously, then we're back to doing linear search which is much worse performance wise than binary search. Plus we'd have to sort at presentation time if we wanted to enforce that strictly. 

I think the +- binary search to get a surrounding slice and then linear search through that slice approach I mentioned will work really well and give us both performance and robustness.

andymandias: Oh, absolutely, I didn't mean to suggest otherwise.  I was just accepting worse chathistory performance in order to avoid messing with the non-chathistory `add_message`.

tarkah: For sure. I'm thinking this add_message can be completely unaware of chat history since we already will have the id and server time in the incoming message, so we can just treat all incoming messages the same for how we insert them.

andymandias: Sounds good to me.  I will leave it to you :)

tarkah:
@andymandias how does this feature interact w/ bouncers that send a replay buffer on connect? Are the bouncers programmed to only send one or the other, or will both get sent? If both are sent, do we just rely on our dedupe strategy to eliminate dupes, or do we have some other mechanism to handle this?

data/src/history.rs
andymandias: Looks good to me!  I agree that checking all messages for a clash is not necessary (and potentially not desirable, since a non-duplicate clash means something odd has happened;  let's cross that bridge when we come to it).

I incorporated it further into the existing PR and made a couple small tweaks:
1.  I think we don't want to mark a message as incrementing the unread count when it is truly a duplicate of an existing message.  Timestamp-based requests are fuzzed, so we will frequently get a message back that the user has already seen.  So I added a bit of logic to not increment the unread message count in that scenario.
2.  `JOIN`/`PART`/`QUIT` messages frequently, in my experience, do not match well on text due to how the user attributes resolve.  E.g. when a new user joins a channel it's not uncommon for their user attributes to not resolve at all if they weren't in another channel with the Halloy user, but if they're still in the channel with the Halloy user when the `chathistory` batch is received then their user attributes will resolve when the `chathistory` copy of the join is received.  So I relaxed the match requirements for those messages to just match on the target, which contains the nick of the user involved in the message.

andymandias:
@tarkah the ircv3 description of the specification says that the replay buffer should not be sent automatically when a client has negotiated `chathistory`.  In my experience with soju that is the case (ZNC, which I've switched off of, does not support `chathistory` to my knowledge).  I think we should be prepared to potentially get dupe messages around join time (even if they follow ircv3 specification, I think we could end up with some dupes if - for example - a message is sent to the channel right after we join but before the server receives our associated `chathistory` request).  But, my feeling at the moment is that our dedupe strategy is sufficient for that purpose.

data/src/history.rs
tarkah: Sounds good to me!

andymandias:
Another restructuring, to better allow for using `LATEST` then repeated `BETWEEN`s to update when joining a channel.  The old scheme using repeated `AFTER` would fail to receive any messages when the reference message was no longer in the server's available message history.  The restructuring should also make it easier to utilize the `TARGETS` subcommand, which I plan to work on next.

Should be a bit clearer in intent and use than before, but it doesn't change any of the problem areas of the PR (`History` and asynchronicity).

data/src/isupport.rs
andymandias: @tarkah I figured out why I thought I was seeing server times rounded to the nearest second;  Halloy's public soju bouncer is returning message server times rounded to the nearest second.  The instance of soju I fired up for personal use was built from commit `6d0f7330e37360d791266bc31d026c4f2cfaad96` (the at the time latest commit on the master branch), and it's returning millisecond server times with messages.  Maybe the public bouncer is running an older version?

andymandias:
Still testing, but at the moment this is what I consider feature complete.  There are a couple additions that probably warrant explanation.

- I added `read_marker` to message histories to store the timestamp of the last read message in a channel/query.  These operate very similar to `opened_at`, but the intention is to have a RFC 3339 timestamp that can be saved and loaded separate of the messages in a history.  Then, when messages arrive we can know whether they trigger unread state without loading the message history and looking for a duplicate.  I'm not aiming to implement [`draft/read-marker`](https://ircv3.net/specs/extensions/read-marker) here, but the intention is to be usable for that feature.  I have been a bit lazy in reading these synchronously using `std::fs`;  I'm hoping they are small enough reads that that's acceptable.  (As a side benefit, `read_marker` allows unread state to persist across application close/open.)
- Since I was adding to the files stored in the history directory (to store the `read_marker` separate from the message history), I took this as an opportunity to tweak the message storage scheme.  I've only done this because I recalled discussion about making the stored messages a bit easier for users to access;  if it's out of scope for this I'm happy to revert it.  (I would probably use a hash for the `read_marker` filename in that case.)

tarkah:
@andymandias @casperstorm Let's find some time to discuss the scope and desired UX of this feature. The PR is now ~2k LOC and makes a lot of API changes and I don't want to dive in and start making suggestions or changes until we are all aligned on scope & UX.

casperstorm:
> @andymandias @casperstorm Let's find some time to discuss the scope and desired UX of this feature. The PR is now ~2k LOC and makes a lot of API changes and I don't want to dive in and start making suggestions or changes until we are all aligned on scope & UX.

We almost needed a RFC for this PR 😅
Perhaps, @andymandias, you could do a small writeup of some of this PR including some decisions you have made along the way. I would love to read something like that before digging into this monster.

andymandias:
@casperstorm @tarkah I may not be able to commit much code to this PR for a bit, but I will do my best to explain the intended features along with an overview of the significant implementation decisions made in service of those features.  I'm going to try and keep it relatively high level to avoid getting lost in the weeds, but will be available to answer any follow-up questions you may have.

As the PR currently stands, the main features are to use `chathistory` to do the following:
1. When connecting to a server, get all new messages since the last login.  This is intended to, as much as possible, "just work" (no user interaction necessary).  This is currently implemented via three mechanisms:
   - A [`TARGETS`](https://ircv3.net/specs/extensions/chathistory#targets) request is made when `chathistory` support is acknowledged by the server.  `TARGETS` is used to discover any queries that were made while the client was not connected, and then `chathistory` requests for the latest messages in those queries are made.  
   - Any time a channel is joined, a request is made via `chathistory` for the latest messages in that channel.  
   - A request for the latest message in any queries that are open on launch.
   - For all three mechanisms above, "latest messages" means all messages since the most recent message in the query/channel history (potentially requiring multiple `chathistory` requests).  If there are no messages in the history, then a single [`LATEST`](https://ircv3.net/specs/extensions/chathistory#latest) `chathistory` request is made.  The maximum number of messages of a request is set by the server, but I also set a client-side maximum at 500 messages (not attached to that particular number).  `TARGETS` uses a `targets_marker` to set its request boundaries, more on that later.
2. When the user reaches the end of a channel/query history, they can request older messages from that channel/query.  Currently this is either done automatically (whenever the user scrolls to the very end of the history; there is an option to turn off this functionality), or manually (by clicking a button that has been placed at the very end of the history).  This submits one `chathistory` request for messages in the channel/query before the earliest message that exists in its client-side history.

I'm not opposed to adding further features, but I wanted to keep the scope of this PR as minimal as possible.

There are two major changes to `History` to support these features:

1. Messages are no longer stored in the order they are received.  Instead the messages list is sorted based on the message's `server_time`, and new messages are inserted accordingly.  Sorting is primarily done to enable deduplication, since the possibility of duplicate messages are an expected part of messages requested via `chathistory`.  New messages are checked to be duplicates against messages with similar `server_time`, based on the message's `id` (if available) or the message's `server_time` and contents (otherwise, excepting some special cases).  @tarkah deserves essentially all of the credit for this, and none of the blame for any aspect that may be broken.
2. A `History` now has a `read_marker` instead of an `opened_at`.  `read_marker` is essentially an `opened_at` that is written to the filesystem, and it serves a similar purpose.  So, the `backlog` marker is placed via `read_marker` in a history in nearly the exactly manner as it was done via `opened_at`.  But a `read_marker` allows for the determination of whether a message is "new" at the time it is inserted into a `History::Partial`.  Since a `History::Partial` cannot be expected to have messages available for deduplication, duplication detection cannot be used as a proxy for message newness.  (I don't think we want to load message history every time a new message is received, in order to check message newness.)  Instead, the `read_marker` is checked against any arriving message's `server_time` in order to only `trigger_unread` when the `server_time` is newer than the `read_marker`.  The `read_marker` is updated the exact same manner that `opened_at` was, except that it is also saved to disk.  (This has the side benefit of persisting that information across program close/open.)
   - I added a new file associated with each server/channel/query history, in which the `read_marker` is written.  I didn't want to store with the message history, since their main purpose is to avoid loading that history and I wasn't sure how to partially load a compressed history (or partially compress a history file).  I took this as an opportunity to make the stored history naming less opaque (i.e. to name history files based on `server/buffer.json.gz` rather than a hash).  Renaming the histories does have one function use;  we can expect histories with the old naming schemes will not have been written sorted (so we should sort them on load), but histories with the new naming scheme will presumably be stored sorted (and won't need to be sorted on load).  We could just sort everything though, and continue with hashes for all of the files.

Going back briefly to the first main feature: I mentioned `TARGETS` uses a `targets_marker` to build its request.  That is, it requests queries/channels that have a new message since the time specified in the `targets_marker`.  The `targets_marker` is the same data as a `read_marker`, but it is updated whenever:
- A `TARGETS` response is received in full to the `server_time` of the last message in the batch
- A `read_marker` is updated on disk to the same value as the `read_marker`.

The goal here is to be fairly conservative.  Don't update the `targets_marker` too readily, in order to reduce the likelihood of missing a message.  But also, don't be extremely conservative, otherwise we'll end up requesting messages for old queries (which results in reopening them in the client, even though nothing new has been sent).  If no `targets_marker` exists, then the start of Unix epoch is used.

That's everything that comes mind at the moment, so I think it might be best to stop here and field questions.

andymandias:
Going to take a crack at converting the `read_marker` (and `targets_marker`) filesystem reads to be async.  I have a better idea of how to implement it now, and doing so will put us in a good place to implement `draft/read-marker` after `chathistory` support is complete.  May be a little bit before I can get to it though.

andymandias:
Conversions of `read_marker` and `targets_marker` to use asynchronous filesystem reads are nominally complete.  I was hoping to have fewer places where `read_marker`s are loaded throughout the code, but ultimately I wasn't able to find a way around having them spread out.

casperstorm:
Sorry this has taken so long, @andymandias.
I am coordinating with @tarkah how we attack this the best. It is quite big and I feel like we should have sat down and discussed how to attack this before jumping into it. It is touching some areas which is quite critical so its hard to just quickly do a review.
Going forward, I think we should do a small RFC when attacking such a big feature. Anyway, know that it is not supposed to die in silence, but we are just trying to carve some time to figure out how we attack it.

andymandias:
I tried to make the minimal set of features to add `chathistory`, but in the end it was quite large. Bigger than I was expecting. More planning/coordination certainly could've helped it. Will be happy to break down the thought process (or lack thereof) for any section of changes that needs more clarity.

As it is, I've been using/testing it as my daily driver (whenever not testing another PR) and it's been working well for me. That said, I'm not attached to any of its particulars. Let's carve it up and reconstruct it into something better 😁

andymandias:
Closing since it is getting broken up into smaller, easier-to-review PRs.

PR #371, opened 2024-05-07
Add support for Windows binary meta information and fix WiX installer…

KaiKorla:
This PR add support for getting the Windows binary meta information in line with the release version and some other information. Also it updates the WiX installer description to have to correct release version also.

To achieve this i added two build dependencies:
- windows_exe_info = "0.4"
- cargo-edit = "0.12"

Tell me what you think.

image

### DIFF
--- a/scripts/build-windows.sh
+++ b/scripts/build-windows.sh
@@ -4,10 +4,13 @@
 #!/bin/bash
 EXE_NAME="halloy.exe"
 TARGET="x86_64-pc-windows-msvc"
-VERSION=$(cat VERSION).0
+HALLOY_VERSION=$(cat VERSION).0
+
+# update package version on Cargo.toml
+cargo install cargo-edit
+cargo set-version $HALLOY_VERSION
 
 # build binary
 rustup target add $TARGET
-cargo-set-version set-version $VERSION
 cargo build --release --target=$TARGET
 cp -fp target/$TARGET/release/$EXE_NAME target/release
--- a/scripts/build-windows.sh
+++ b/scripts/build-windows.sh
@@ -4,8 +4,10 @@
 #!/bin/bash
 EXE_NAME="halloy.exe"
 TARGET="x86_64-pc-windows-msvc"
+VERSION=$(cat VERSION).0
 
 # build binary
 rustup target add $TARGET
+cargo-set-version set-version $VERSION
 cargo build --release --target=$TARGET
 cp -fp target/$TARGET/release/$EXE_NAME target/release
--- a/wix/main.wxs
+++ b/wix/main.wxs
@@ -21,7 +21,7 @@
         <StandardDirectory Id="ProgramFiles6432Folder">
             <Directory
                 Id="INSTALLFOLDER"
-                Name="!(bind.Property.Manufacturer)\!(bind.Property.ProductName)"
+                Name="!(bind.Property.ProductName)"
             />
         </StandardDirectory>
 
--- a/wix/main.wxs
+++ b/wix/main.wxs
@@ -16,7 +16,7 @@
         Manufacturer='Squidowl'
         Language='1033'
         Codepage='1252'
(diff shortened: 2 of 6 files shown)


wix/main.wxs
```diff
@@ -16,7 +16,7 @@
         Manufacturer='Squidowl'
         Language='1033'
         Codepage='1252'
-        Version='0.0.{{ VERSION }}'>
+        Version='{{ VERSION }}.0.0'>
```
casperstorm, line 19: I am not the big Windows user, but noticed that I have apps have different version notations:
Surely there must be a way for us to do yyyy.version like we do everywhere else? I just recall adding `0.0.{{ VERSION }}` because otherwise WiX would fail to compile 🤔 

image

scripts/build-windows.sh
```diff
@@ -4,8 +4,10 @@
 #!/bin/bash
 EXE_NAME="halloy.exe"
 TARGET="x86_64-pc-windows-msvc"
+VERSION=$(cat VERSION).0
```
casperstorm, line 7: We should keep `Product version` and `File version` equal imho.
And preferably both stick to our normal schematics.

wix/main.wxs
KaiKorla: Yea, well it is an Windows Installer SDK limitation (which is, IMHO, bad). That means that the Windows version will be always differ from the version in the VERSION file, as the major version must be less than 256. :-(

I will change the PR.

KaiKorla: To be more precise it seems a WiX 3 limitation. The Windows Installer SDK itself defines version as documented [here](https://learn.microsoft.com/en-us/windows/win32/msi/version). I made a brief attempt with WiX 5 and it seems that there is a warning if the version is "2004.7" and no error. Is it an option to rewrite the WiX description and migrate to WiX 5? I could make this.

casperstorm:
Note @KaiKorla that i just modified the Wix file slightly after #363 was merged.

casperstorm:
@KaiKorla still working on this?

KaiKorla:
Yes.

scripts/build-windows.sh
KaiKorla: Cannot remove the trailing 0. If i removing it, then `cargo-set-version set-version $VERSION` is failing.

tarkah:
@KaiKorla I've rebased all your commits onto `main`. Please use `git rebase` instead of `git merge` so we can keep a clean git history, thanks!

KaiKorla:
> @KaiKorla @casperstorm I noticed we were pulling down windows deps / using them in `build.rs` for non windows targets. Can you both confirm that [791b6ce](https://github.com/squidowl/halloy/commit/791b6ce42a111b4b8c6f5c20d72591431cf0ae08) doesn't break the windows build?

Working locally. Binary and installer looking good in the Windows sandbox. IMO everything is working as expected. tyvm!

casperstorm (changes requested):
As discussed on IRC, final two missings:
* Install dir should be `C:\Program Files\Halloy` instead of `C:\Program Files\Squidowl\Halloy`
* When opening a url scheme the Halloy icon shows description rather than just title.

KaiKorla:
> As discussed on IRC, final two missings:
> 
> * Install dir should be `C:\Program Files\Halloy` instead of `C:\Program Files\Squidowl\Halloy`
> * When opening a url scheme the Halloy icon shows description rather than just title.

- Install dir is fixed
- Regarding the url sheme and the description:
It seems that Windows is showing the "File Description" metadata and then cache it somehow (i try to figure it out in the microsoft documentation). The "File Description" is currently set by "windows_exe_info::versioninfo::link_cargo_env();" in the build.rs and takes the value from the description field in the [package] section in the Cargo.toml file. IMHO we shoudln't change this.

casperstorm (approved):
Tested and works. Thank you so much for this.

PR #375, opened 2024-05-08
feat: overwrite nick color

casperstorm:
Ability to overwrite nickname colors by providing a hex string.
Note that I didn't add any breaking changes since we are still able to deserde from a string, instead of a map:

The new format looks like
```toml
[buffer.nickname.color]
kind = "unique"
hex = "#ff0000" # optional
```

But we still correctly parse
```toml
[buffer.nickname]
color = "unique"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+Added:
+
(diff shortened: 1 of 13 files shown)


tarkah (approved):
LGTM. I made one change to simplify the backwards compatible deserialization implementation. Can you please confirm it still works before merging?

PR #377, opened 2024-05-09
feat: overwrite server & internal messages color

### LINKED ISSUE #369: Feature request: Allow customising join/part/quit message colours
I've noticed in the buffer.server_messages section of the configuration you can differentiate between message types of join, part and quit. Would it be possible to have a colour option added to specify the colour the of the corresponding message please? That way when you have join, part and quit enabled you can see at a glance what happened based on colour. It makes readability better. As an example I like joins to be green and parts, quits and kicks to be red.

casperstorm:
fixes #369


tarkah (approved):
LGTM

I think any more future customization around how messages are colored or formatted warrants us to reconsider how to support more complex customization via some new formatting config / API

PR #378, opened 2024-05-14
User Context Menu Status Icon Tweak

andymandias:
The original idea for the user context menu status icon was to use a unicode character so that it would line up precisely with the last character.  However, playing around with font settings recently I've found that most fonts do not have that character, and depending on the font the fallback font the circle ends up being way too big/small.  So I switched it to use the same method as the sidebar's unread messages icon.  Looks about the same to me when using default font settings.


casperstorm (approved):
Tested, and works.

PR #381, opened 2024-05-17
Shell Expansion Documentation

andymandias:
Minor update to the documentation to note the lack of shell expansion in configuration path strings, as discussed in `#halloy` earlier today.


casperstorm (approved):
Ah great, make sense.

PR #382, opened 2024-05-17
Support expanded nick parsing

tarkah:
Usecase here is that matrix bridge requires full user string as `foo:matrix.org!foo@matrix.org` where nick is `foo:matrix.org`.

We need to support `:` and `.` in nick without accidentally parsing a nick as an IP address or hostname. To do this, we require that both `:` and `.` need to be present and the nick must be followed by `!` / username. The `:` prevents a hostname from being parsed and the trailing `!` requirement prevents it from parsing against an IP address

### DIFF
--- a/irc/proto/src/parse.rs
+++ b/irc/proto/src/parse.rs
@@ -3,7 +3,7 @@ use std::string::FromUtf8Error;
 use nom::branch::alt;
 use nom::bytes::complete::tag;
 use nom::character::complete::{alpha1, char, crlf, none_of, one_of, satisfy};
-use nom::combinator::{cut, map, opt, peek, recognize, value};
+use nom::combinator::{cut, map, opt, peek, recognize, value, verify};
 use nom::multi::{many0, many0_count, many1, many1_count, many_m_n, separated_list1};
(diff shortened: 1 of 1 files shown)


irc/proto/src/parse.rs
```diff
@@ -127,12 +127,28 @@ fn user(input: &str) -> IResult<&str, User> {
...
     // "-", "[", "]", "\", "`", "_", "^", "{", "|", "}", "*", "/", "@"
-    let special = one_of("-[]\\`_^{|}*/@");
+    let special = |input| one_of("-[]\\`_^{|}*/@")(input);
     // *( <letter> | <number> | <special> )
-    let nickname = recognize(many1_count(alt((
+    let strict_nick = recognize(many1_count(alt((
         satisfy(|c| c.is_ascii_alphanumeric()),
         special,
     ))));
+    // Use by things like matrix bridge
```
casperstorm, line 136: Nit

```suggestion
    // Used by things like matrix bridge
```

casperstorm (approved):
LGTM, haven't tested.

PR #383, opened 2024-05-22
Mention Flathub and Snap Store in README.md

Flimm:
The Repology image does not list Flathub and Snap entries. The developers of Repology have stated that they never will support listing Flathub or Snap entries.

I think many users would be interested to know that Flathub or Snap packages are available, without having to navigate to further documentation, so I added one sentence to the README.md file.


casperstorm:
I noticed that they dont want to list it, so I am ok with this addition. Thank you!

casperstorm (approved):

PR #384, opened 2024-05-22
Theme Configuration Parsing Tweak

andymandias:
A minor tweak to accept `"<theme_name>.toml"` as well as `"<theme_name>"` in the configuration `.toml`.  The documentation specifies it should be the name of the file, but does not say whether or not to include the extension;  we could make the documentation more specific/restricted, but since it's fairly simple to allow both formats that seems like the more user-friendly option to me.


casperstorm:
Good call, @andymandias. I like this.

PR #387, opened 2024-06-03
"Leave Buffer" and "Toggle Sidebar" Shortcuts

andymandias:
Added shortcuts for "leave buffer" (i.e. leave channel or close query as appropriate) and "toggle sidebar".  Picked "leave buffer" as a three-key shortcut by default to reduce the chance of accidental input, and "toggle sidebar" to use `b` in order to leave `s` open for other, future shortcuts.  But I'm not particularly attached to either.


casperstorm (approved):
I like this, always welcome more shortcuts.
Could you add this to CHANGELOG? :)

andymandias:
Ah, yes, of course.  Updated :sweat_smile:

PR #390, opened 2024-06-06
UTF-8 Channel Name Rendering Fix

andymandias:
Channels with UTF-8 name can include glyphs that are not commonly covered by fonts (e.g. `#☃` on `irc.hackint.org`).  I've turned on advanced shaping for sidebar items and the pane title bar text to ensure that those channel names render properly via font fallback.


casperstorm (approved):
Make sense to me! Good addition.

PR #391, opened 2024-06-06
feat/persist window position and size

### LINKED ISSUE #379: Feature request: Remember window size and position between launches
Fairly self-explanatory: rather than open at a default size centered on the primary screen's workspace [confirmed on MacOS, v 2024.7], remember the last used size and position and restore there if possible. If not possible, consult the Configuration for a suggested size and position. If still not possible, fallback to centered as it is now.

casperstorm:
fixes #379


src/window.rs
```diff
@@ -59,3 +63,152 @@ pub fn settings() -> Settings {
...
+
+    fn hash(&self, state: &mut subscription::Hasher) {
+        use std::hash::Hash;
+
+        std::any::TypeId::of::<Self>().hash(state);
+    }
+
+    fn stream(self: Box<Self>, events: subscription::EventStream) -> BoxStream<'static, Self::Output> {
+        use futures::stream;
+        const TIMEOUT: std::time::Duration = std::time::Duration::from_secs(3);
```
tarkah, line 103: 3 seconds feels a bit long? For instance, a double click has to register typically within 500ms. Maybe 1 second?

tarkah (changes requested):
A couple small things, otherwise this looks great!

src/window.rs
casperstorm: Changing to 500ms.

src/main.rs
```diff
@@ -178,6 +183,7 @@ impl Halloy {
                 servers: config.servers.clone(),
                 config,
                 modal: None,
+                window: Window::load().unwrap_or_default(),
             },
```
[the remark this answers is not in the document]
tarkah: Disregard, `load_from_state` is called in other places

tarkah (approved):

darienm:
Version 2024.8 MacOS does not return the window to the last used position or size. It's in a similar region on-screen (no longer centered) but not where it was left, which doesn't achieve the goal. Do you need more info? Primary screen is 3840x2160 and secondary screen is 2560x1600. (1) Window does not return to proper screen (2) Window seems to get smaller with each launch. (3) Window seems to disregard positions greater than around 1400x1400.

PR #392, opened 2024-06-09
Input Error Overlay Size 

andymandias:
In my testing the error overlay for user input fills the full height of the pane.  E.g. typing then attempting to submit `/msg BouncerServer` (without a message argument) results in a full height overlay that looks like:
Screenshot from 2024-06-09 14-43-36
This minor tweak is to instead shrink the size of the container to the text (not worthy of a changelog entry, I think).


casperstorm:
I don't actually have this error on macOS, below is me running `/msg BouncerServer` on main branch.
However I get the same result from this branch, so im good merging if it solves the problem on Linux.

image

@tarkah you mind verifying?

casperstorm (approved):

PR #395, opened 2024-06-17
Gruvbox theme

seth0xd:
I couldn't find one [here](https://halloy.squidowl.org/configuration/themes/community.html) so I hope you'd like this one.
image


casperstorm (changes requested):
Thanks for taking the interest in Halloy!
To add a community theme, please add the toml, and image to` book/src/configuration/themes` and then reference it in the `book/src/configuration/themes/community.md` file!

You can see all community themes here: https://halloy.squidowl.org/configuration/themes/community.html

seth0xd:
My bad should've known it wasn't the right directory, there was only one theme there :skull:

casperstorm (approved):

PR #407, opened 2024-07-03
custom notification sounds

casperstorm:


data/src/audio.rs
```diff
@@ -0,0 +1,60 @@
...
+use crate::Config;
+
+#[derive(Debug, Clone, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub enum Sound {
+    Internal(Internal),
+    External(String),
+    None,
+}
+
```
tarkah, line 14: Instead of encoding None here, we can do `Option<Sound>` which is more idiomatic.

Also, using Option<T> with serde works ootb and you won't need to define a custom default definition.

src/notification/audio.rs
```diff
@@ -0,0 +1,67 @@
...
+                audio::Sound::Internal(internal) => {
+                    let bytes = internal.bytes();
+                    StaticSoundData::from_cursor(Cursor::new(bytes))?
+                }
+                audio::Sound::External(sound_name) => {
+                    let Some(sound_path) = data::audio::find_external_sound(sound_name) else {
+                        return Err(PlayError::NoSoundFound);
+                    };
+
+                    StaticSoundData::from_file(sound_path)?
```
tarkah, line 39: We're doing blocking IO here during the main thread / `update`.

We either need to:

- load sound data into memory after we parse the config. We can hook this via Config::load like we do with themes and password files
- fire Tasks to load the data then play it once loaded. This is a bit more complex.

I vote for the first option

tarkah (commented):
Really clean work! One minor suggestion and then one change we need to prevent blocking IO on the UI thread

casperstorm:
> Really clean work! One minor suggestion and then one change we need to prevent blocking IO on the UI thread

Perfect good review. Can you perhaps look into the Linux dependency?

tarkah (approved):
Awesome stuff!

PR #413, opened 2024-07-14
Support for CTCP queries `CLIENTINFO`, `PING`, `SOURCE`, and `VERSION`

andymandias:
Using <https://raw.githubusercontent.com/DanielOaks/irc-rfcs/master/dist/draft-oakley-irc-ctcp-latest.txt> as reference for CTCP, this PR implements three changes:
1.  Responds to `CLIENTINFO`, `PING`, `SOURCE`, and `VERSION` queries.  These queries seem benign to reply to.  `FINGER`, `TIME`, and `USERINFO`, on the other hand, give information about the user that I don't think should be provided automatically.
2.  Does not pass any CTCP queries on to the user (whether responded to or not).
3.  Permits incoming CTCP queries to not be properly terminated by `\u{1}` (as prescribed by the reference document).

Primarily creating this to stop having `\u{1}VERSION\u{1}` messages from OFTC's `CTCPServ` show up in the GUI.

Didn't add anything to the changelog since there is little to no user-facing functionality provided as far as I can tell.

### DIFF
--- a/data/src/ctcp.rs
+++ b/data/src/ctcp.rs
@@ -4,21 +4,20 @@ pub struct Query<'a> {
     pub params: &'a str,
 }
 
-pub fn is_ctcp_query(text: &str) -> bool {
+pub fn is_query(text: &str) -> bool {
     text.starts_with('\u{1}')
 }
 
-pub fn parse_ctcp_query(text: &str) -> Result<Query, &'static str> {
+pub fn parse_query(text: &str) -> Option<Query> {
     let query = text
         .strip_suffix('\u{1}')
         .unwrap_or(text)
-        .strip_prefix('\u{1}')
-        .ok_or("text is not a CTCP query")?;
+        .strip_prefix('\u{1}')?;
 
     if let Some((command, params)) = query.split_once(char::is_whitespace) {
-        Ok(Query { command, params })
+        Some(Query { command, params })
     } else {
-        Ok(Query {
+        Some(Query {
             command: query,
             params: "",
         })
--- a/data/src/ctcp.rs
+++ b/data/src/ctcp.rs
@@ -1,0 +1,26 @@
+#[derive(Debug)]
+pub struct Query<'a> {
+    pub command: &'a str,
+    pub params: &'a str,
+}
+
+pub fn is_ctcp_query(text: &str) -> bool {
+    text.starts_with('\u{1}')
+}
+
+pub fn parse_ctcp_query(text: &str) -> Result<Query, &'static str> {
+    let query = text
+        .strip_suffix('\u{1}')
+        .unwrap_or(text)
+        .strip_prefix('\u{1}')
+        .ok_or("text is not a CTCP query")?;
+
+    if let Some((command, params)) = query.split_once(char::is_whitespace) {
+        Ok(Query { command, params })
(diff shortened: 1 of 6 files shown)


casperstorm (approved):
Great, thanks @andymandias. @tarkah you have anything?

data/src/message.rs
```diff
@@ -544,12 +544,24 @@ impl Limit {
...
-fn is_action(text: &str) -> bool {
-    text.starts_with("\u{1}ACTION ") && text.ends_with('\u{1}')
+pub fn ctcp_command(text: &str) -> Option<&str> {
+    let command = text
+        .strip_suffix('\u{1}')
+        .unwrap_or(text)
+        .strip_prefix('\u{1}')?;
+
+    command.split_whitespace().next()
+}
```
tarkah, line 554: I think we have some logic in the dcc module for parsing ctcp payload as well. Can we reference this from that module so we don't have duplicated logic for doing this?

data/src/message.rs
```diff
@@ -544,12 +544,24 @@ impl Limit {
     }
 }
 
-fn is_action(text: &str) -> bool {
-    text.starts_with("\u{1}ACTION ") && text.ends_with('\u{1}')
+pub fn ctcp_command(text: &str) -> Option<&str> {
+    let command = text
+        .strip_suffix('\u{1}')
+        .unwrap_or(text)
```
tarkah, line 550: Is the suffix optional? Why do we not require it?

andymandias: I was going by [this CTCP draft](https://raw.githubusercontent.com/DanielOaks/irc-rfcs/master/dist/draft-oakley-irc-ctcp-latest.txt), which is the most recent draft I found linked from <https://modern.ircdocs.horse/ctcp>.  It says,
```
   The final CTCP delimiter SHOULD be sent on outgoing messages for
   compatibility reasons, but software MUST accept incoming messages
   which lack it.
```
But I'm not particularly attached to adhering to the draft, it's just the closest thing I could find to a standard.

tarkah: Makes sense 👍

data/src/message.rs
andymandias: Sounds good.  I'll consolidate everything in a CTCP module.

data/src/ctcp.rs
```diff
@@ -0,0 +1,26 @@
...
+pub struct Query<'a> {
+    pub command: &'a str,
+    pub params: &'a str,
+}
+
+pub fn is_ctcp_query(text: &str) -> bool {
+    text.starts_with('\u{1}')
+}
+
+pub fn parse_ctcp_query(text: &str) -> Result<Query, &'static str> {
```
tarkah, line 11: Nit: ctcp::is_query / parse_query. We already namespace it via ctcp module

tarkah, line 11: Nit: I don't think Result gives us anything, plus we always .ok it. Let's make this an Option

tarkah (approved):
LGTM, just a couple nits

andymandias:
Thanks for the reviews!  I appreciate the attention to detail.

tarkah (approved):

PR #419, opened 2024-07-17
feat: description and a few alias to commands

### LINKED ISSUE #372: aliasing commands
Hi, could you please add a config option to alias some commands, e.g. /join to /j?

casperstorm:
Added a small description above commands, as well as a few aliases.
Further down the road we should perhaps consider making it a configuration to define your own aliases, but for now its not needed imho.

This fixes #372.


book/src/commands.md
```diff
@@ -0,0 +1,25 @@
...
+
+Halloy will first try to run below commands, and lastly send it directly to the server.
+
+| Command | Alias      | Description                                                   |
+|---------|------------|---------------------------------------------------------------|
+| `away`  |            | Mark yourself as away. If already away, the status is removed |
+| `join`  | `j`        | Join channel(s) with optional key(s)                          |
+| `me`    | `describe` | Send an action message to the channel                         |
+| `mode`  | `m`        | Set mode(s) on a channel or retrieve the current mode(s) set  |
+| `msg`   | `query`    | Open a query with a nickname and send an optional message     |
```
tarkah, line 19: This description doesn't seem entirely accurate, since `query` is only an alias of `/msg` and `msg` targets can be channels or nicknames

tarkah (approved):
Looks good, this is a nice change! Just the one minor comment

PR #420, opened 2024-07-21
Fix DCC Send: Allow filenames with spaces

### LINKED ISSUE #318: DCC Send fails on filenames with whitespace
These lines linked below, if the filename contains a space such as:
"DCC SEND Example file name.txt (127.0.0.1)"
The iterator spits out "Example" as the filename, then attempts to parse the rest as an IP.

https://github.com/squidowl/halloy/blob/698eef88b17832881987f2da16faad7e4469a71d/data/src/dcc.rs#L69-L70

Also on [line 11 ](https://github.com/squidowl/halloy/blob/698eef88b17832881987f2da16faad7e4469a71d/data/src/dcc.rs#L11) I'm unsure but you may want to throw a `.trim()` to strip white space before the iterator call?

I may take a stab at putting a PR in for this, but I'm a little unclear of all the surrounding logic still. No promises.

Good stuff though, nicely written!

ameknite:
Fixes #318

Thanks to this comment I was able to fix the bug: https://github.com/squidowl/halloy/issues/318#issuecomment-2037129873

### DIFF
--- a/data/src/dcc.rs
+++ b/data/src/dcc.rs
@@ -66,26 +66,28 @@ impl Send {
         }
     }
 
-    fn decode<'a>(args: impl Iterator<Item = &'a str>) -> Option<Self> {
-        let mut args: Vec<&str> = args.collect();
-        args.reverse();
+    fn decode<'a>(
+        args: impl std::iter::DoubleEndedIterator<Item = &'a str> + std::clone::Clone,
+    ) -> Option<Self> {
+        let mut args = args.rev();
+        let mut args_token = args.clone();
 
         // if token doesn't exist
         let mut token = None;
-        let mut size = args[0];
-        let mut port = args[1];
-        let mut host = args[2];
-        let mut filename = args.iter().skip(3);
+        let mut size = args.next()?;
+        let mut port = args.next()?;
+        let mut host = args.next()?;
+        let mut filename = args;
 
         // If token exists, port == 0
         // args[1] == port, if token doesn't exists.
         // args[2] == port, if token exists.
-        if args[1].parse::<u16>() == Ok(0) || args[2].parse::<u16>() == Ok(0) {
-            token = Some(args[0].parse::<NonZeroU16>().ok()?);
-            size = args[1];
-            port = args[2];
-            host = args[3];
-            filename = args.iter().skip(4);
+        if size.parse::<u16>() == Ok(0) || port.parse::<u16>() == Ok(0) {
+            token = Some(args_token.next()?.parse::<NonZeroU16>().ok()?);
+            size = args_token.next()?;
+            port = args_token.next()?;
+            host = args_token.next()?;
+            filename = args_token;
         }
 
         // Parse values
@@ -94,7 +96,7 @@ impl Send {
         let size = size.parse().ok()?;
         let filename = filename
             .rev()
-            .map(|&s| s.trim_matches('\"'))
+            .map(|s| s.trim_matches('\"'))
             .collect::<Vec<_>>()
             .join(" ");
 
--- a/data/src/dcc.rs
+++ b/data/src/dcc.rs
@@ -66,12 +66,33 @@ impl Send {
         }
     }
 
-    fn decode<'a>(mut args: impl Iterator<Item = &'a str>) -> Option<Self> {
-        let filename = args.next()?.to_string();
-        let host = args.next().and_then(decode_host)?;
-        let port = NonZeroU16::new(args.next()?.parse().ok()?);
-        let size = args.next()?.parse().ok()?;
-        let token = args.next();
+    fn decode<'a>(args: impl Iterator<Item = &'a str>) -> Option<Self> {
+        let mut args: Vec<&str> = args.collect();
+        args.reverse();
+
+        // if token doesn't exist
+        let mut token = None;
+        let mut size = args[0];
+        let mut port = args[1];
+        let mut host = args[2];
+        let mut filename = args.iter().skip(3);
+
+        // If token exists, port == 0
+        // args[1] == port, if token doesn't exists.
+        // args[2] == port, if token exists.
+        if args[1].parse::<u16>() == Ok(0) || args[2].parse::<u16>() == Ok(0) {
+            token = Some(args[0].parse::<NonZeroU16>().ok()?);
+            size = args[1];
+            port = args[2];
+            host = args[3];
+            filename = args.iter().skip(4);
+        }
+
+        // Parse values
+        let host = decode_host(host)?;
+        let port = NonZeroU16::new(port.parse().ok()?);
+        let size = size.parse().ok()?;
+        let filename = filename.rev().copied().collect::<Vec<_>>().join(" ");
 
         match (port, token) {
             (_, Some(token)) => Some(Self::Reverse {
--- a/data/src/dcc.rs
+++ b/data/src/dcc.rs
@@ -82,8 +82,8 @@ impl Send {
         // If token exists, port == 0
         // args[1] == port, if token doesn't exists.
         // args[2] == port, if token exists.
-        if size.parse::<u16>() == Ok(0) || port.parse::<u16>() == Ok(0) {
-            token = Some(args_token.next()?.parse::<NonZeroU16>().ok()?);
+        if port.parse::<u16>() == Ok(0) || host.parse::<u16>() == Ok(0) {
+            token = args_token.next();
             size = args_token.next()?;
             port = args_token.next()?;
             host = args_token.next()?;
--- a/data/src/dcc.rs
+++ b/data/src/dcc.rs
@@ -167,3 +167,131 @@ fn encode_host(host: IpAddr) -> String {
         IpAddr::V6(v6) => v6.to_string(),
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn send_decode() {
+        let args = "my_file_name 1402301083 12350 1453953495";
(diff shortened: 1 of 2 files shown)


data/src/dcc.rs
```diff
@@ -66,12 +66,37 @@ impl Send {
...
 
-    fn decode<'a>(mut args: impl Iterator<Item = &'a str>) -> Option<Self> {
-        let filename = args.next()?.to_string();
-        let host = args.next().and_then(decode_host)?;
-        let port = NonZeroU16::new(args.next()?.parse().ok()?);
-        let size = args.next()?.parse().ok()?;
-        let token = args.next();
+    fn decode<'a>(args: impl Iterator<Item = &'a str>) -> Option<Self> {
+        let mut args: Vec<&str> = args.collect();
+        args.reverse();
```
tarkah, line 71: We can make this bounds a double ended iterator I believe. Though it'd impact the logic below so maybe not.

data/src/dcc.rs
```diff
@@ -66,12 +66,37 @@ impl Send {
...
+        let mut token = None;
+        let mut size = args[0];
+        let mut port = args[1];
+        let mut host = args[2];
+        let mut filename = args.iter().skip(3);
+
+        // If token exists, port == 0
+        // args[1] == port, if token doesn't exists.
+        // args[2] == port, if token exists.
+        if args[1].parse::<u16>() == Ok(0) || args[2].parse::<u16>() == Ok(0) {
```
tarkah, line 83: Slicing is unsafe as we haven't validated args >= 3 so these can panic / crash the application

data/src/dcc.rs
```diff
@@ -66,12 +66,37 @@ impl Send {
...
+            token = Some(args[0].parse::<NonZeroU16>().ok()?);
+            size = args[1];
+            port = args[2];
+            host = args[3];
+            filename = args.iter().skip(4);
```
tarkah, line 88: The logic doesn't seem right. If args[1] == 0 then we don't have token, but then you parse args[0] as token instead of size.

Having port 0 without token shouldn't be allowed, we may need to return None in that case

tarkah (changes requested):
Thanks for tackling this! I've left a few comments. I think we should also add some unit tests since the logic is a little tricky.

data/src/dcc.rs
ameknite: the problem is that there is not a easy way to determine if token exists, so I don't the exact position of port.
This could be easier if instead of an iterator of strings. it was already a struct or if the token was marked as None if doesn't exists.
I am not familiar with irc or dcc, is there a reason why args is a iterator of strings or the position of the values?

data/src/dcc.rs
ameknite: Good idea. Added.

data/src/dcc.rs
ameknite: after changing the logic to use iterators, this doesn't apply anymore.

data/src/dcc.rs
tarkah: I think we can just check `args.len > 4 and args[len - 3] == 0` to be positive the last arg is a token w/ port 0 provided and enough args to include all the other required parameters.

ameknite: With that logic:

args = "my file name" 1402301083 12350 1453953495
args.len = 6
args[len-3] == 1402301083 == host
Here you are checking the host not the port

args = "my file name" 1402301083 12350 1453953495 token
args.len = 7
args[len-3] == 12350 == port
Here you are checking the port

not much different than this:
port.parse::<u16>() == Ok(0) || host.parse::<u16>() == Ok(0)

ameknite: What is the structure of token? can be modified to put a None in the String when it doesn't exist? 
This will make thinks easier.

ameknite:
Here are some test, but without knowing if a token exist, it has some problems:

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn send_decode() {
        let args = "my_file_name 1402301083 12350 1453953495";
        let send = Send::decode(args.split_whitespace());
        assert_eq!(
            send,
            Some(Send::Direct {
                filename: "my_file_name".to_string(),
                host: IpAddr::V4(Ipv4Addr::from(1402301083)),
                port: NonZeroU16::new(12350).unwrap(),
                size: 1453953495
            })
        );
    }

    #[test]
    fn send_decode_whitespace() {
        let args = "my file name 1402301083 12350 1453953495";
        let send = Send::decode(args.split_whitespace());
        assert_eq!(
            send,
            Some(Send::Direct {
                filename: "my file name".to_string(),
                host: IpAddr::V4(Ipv4Addr::from(1402301083)),
                port: NonZeroU16::new(12350).unwrap(),
                size: 1453953495
            })
        );
    }

    #[test]
    fn send_decode_quotation_marks() {
        let args = "\"my file name\" 1402301083 12350 1453953495";
        let send = Send::decode(args.split_whitespace());
        assert_eq!(
            send,
            Some(Send::Direct {
                filename: "my file name".to_string(),
                host: IpAddr::V4(Ipv4Addr::from(1402301083)),
                port: NonZeroU16::new(12350).unwrap(),
                size: 1453953495
            })
        );
    }

    #[test]
    fn send_decode_token() {
        let args = "\"my file name\" 1402301083 0 1453953495 token";
        let send = Send::decode(args.split_whitespace());
        assert_eq!(
            send,
            Some(Send::Reverse {
                filename: "my file name".to_string(),
                host: IpAddr::V4(Ipv4Addr::from(1402301083)),
                port: None,
                size: 1453953495,
                token: "token".to_string()
            })
        );
    }

    #[test]
    fn send_decode_port_zero() {
        // Works because host is now the last word of filename
        // and that is not an number, host = name\"
        // so host fails to parse and return None.
        let args = "\"my file name\" 1402301083 0 1453953495";
        let send = Send::decode(args.split_whitespace());
        assert_eq!(send, None);

        // Works because host is now the last word of filename
        // and that is not an number, host = 2\"
        // due to the quotation marks,
        // so host fails to parse and return None.
        let args = "\"my file name 2\" 1402301083 0 1453953495";
        let send = Send::decode(args.split_whitespace());
        assert_eq!(send, None);

        // Works because port is now host and is too large to fit in NonZeroU16
        // port = host = 1402301083
        // so port fails to parse and return None.
        let args = "my file name 2 1402301083 0 1453953495";
        let send = Send::decode(args.split_whitespace());
        assert_eq!(send, None);

        // This will fail because
        // file name doesn't have "" and last word is a number
        // host == last == 2
        // and host is small enough to fit in NonZeroU16
        // port = host = 16532
        // so this will not return None when parsing host or port
        // let args = "my file name 2 16532 0 1453953495";
        // let send = Send::decode(args.split_whitespace());
        // assert_eq!(send, None);
    }

    #[test]
    fn send_decode_token_with_port() {
        // Works because size is now token, and token is not a number
        // size = token
        // so size fails to parse and return None.
        let args = "\"my file name\" 1402301083 12350 1453953495 token";
        let send = Send::decode(args.split_whitespace());
        assert_eq!(send, None);

        // Works because port is now size,
        // and size is too big to fit in NonZeroU16
        // port = size
        // so port fails to parse and return None.
        let args = "\"my file name\" 1402301083 12350 1453953495 1111";
        let send = Send::decode(args.split_whitespace());
        assert_eq!(send, None);

        // This will fail because
        // token is a number (token = 1111) and
        // port = size = 1000, but is small enough to fit in NonZeroU16
        // so this will not Return None
        // let args = "\"my file name\" 1402301083 12350 1000 1111";
        // let send = Send::decode(args.split_whitespace());
        // assert_eq!(send, None);
    }
}

data/src/dcc.rs
tarkah: I forgot that port isn't always 0 when token is provided, for example the person responding to the reverse request will put their host / port in along with the token.

Thinking through this again, I think the most robust way is to iterate in reverse until we find the arg that successfully parses as IpAddr. Since this won't be token, size or port, this should always be reliable. We can then check if 2 or 3 args exist AFTER that arg to know 100% if token exists. And everything BEFORE that arg must be the filename.

ameknite: The problem is that IpAdrr just check if the string can be parsed to an u32.
```rust
fn decode_host(host: &str) -> Option<IpAddr> {
    match host.parse::<u32>() {
        Ok(n) => Some(IpAddr::V4(Ipv4Addr::from(n))),
        Err(_) => host.parse().ok(),
    }
}
```
So you can actually parse `size` and `port` as `IpAdrr`. Also if the `filename` include only numbers in the last word, it will be parsed as `IpAdrr`.

tarkah: Ahh right, ipv4 in this "spec" is sent as u32 BE encoding for the 4 bytes. Bummer! 

We could assume the _last_ successful IpAddr parsed with decode_host in rev iteration?

tarkah: @ameknite Checkout a46eec9 and let me know what you think.

The only potential gap I can think of is if token isn't set and filename has spaces + ends in an integer. But I think this is an extreme edge case that shouldn't be accounted for, as it's probably impossible to account for. So I think this is sufficient.

tarkah:
@ameknite Thanks a ton for this PR! I've tested this final implementation and it works great. I've also added most of your tests and confirmed the only edge case that wont parse is: ">=5 args w/out token and filename ends in numeric", which I think is acceptable. Most files will have a file extension and so I can't see this happening often or ever. Plus halloy as the sender will never run into this since we always sanitize file names before sending requests.

tarkah (approved):

Issue #422, opened 2024-07-21
Halloy keeps audio device open on Linux


anarsoul:
Halloy built from git master (5c597db06cde887e27a7a320af35266ca8671449) constantly keeps audio device open on Linux which results in speaker icon next to Halloy icon in task bar and in increased power consumption.

![image](https://github.com/user-attachments/assets/e8177d16-ac52-4858-be03-0ef13bb48616)

You can also see that halloy plays audio in `pw-top` output.

![image](https://github.com/user-attachments/assets/333d6b16-71b5-4316-95ef-1b549d18879a)

PR #423, opened 2024-07-21
Only Open Audio Device When Playing Audio

andymandias:
This is intended to address issue #422.  Works for me on Fedora GNOME 46, but have been able to test other OSes/distros yet.

Only did a brief survey of audio crates, but didn't find anything that was certain to avoid issue #422 while also being as high-level as `kira`.  So I went the route of spinning up an audio manager for each notification sound, then letting it drop after it plays.  Seems lightweight enough in my testing (no difference vs previous implementation as far as I can hear).


tarkah:
If all we need to do is play a sound, I wonder how using cpal directly would go. I haven't looked at the API but I'm also not opposed to this.

andymandias:
Looked into `cpal` a bit, but it looked a little low-level for what I think we want.  In doing so came across `rodio` which RustAudio also provides, and I think it's a pretty good fit for what we want in terms of capabilities and ease-of-use.  But, the solution remains more-or-less the same;  that is, we have to open a new `OutputStream` whenever we play a sound, similar to how we needed to open a new `AudioManager` for each notification.  Seems plausible that an `OutputStream` is lighter weight than an `AudioManager`, but I don't know how to determine whether that's actually the case.

I've converted the notification sounds to use `rodio`, but I'm okay with going back to the `kira` solution.

tarkah:
I think in the end let's just choose whatever has fewer deps / less bloat. They both use cpal I believe so I'm not sure we will have much difference in compatibility.

andymandias:
Looks like `rodio` has fewer dependencies.  Both do use `cpal` (and both can be configured to use Symphonia for decoding audio formats), so I would also expect there will be little difference in compatibility.

casperstorm:
I looked at `Rodio` as well, and liked it. Its actively being developed, and has few deps.
I am good with choosing this over `kira`.

src/notification.rs
```diff
@@ -4,75 +4,54 @@ use data::{
...
 
     if let Some(sound) = &notification.sound {
-        audio.play(sound);
+        let sound = sound.clone();
+
+        tokio::spawn(async move {
+            if let Err(error) = sound.play() {
+                log::debug!("Unable to play notification sound: {error}");
+            }
+        });
```
tarkah, line 55: We should not be blocking in an async thread. We either need to just use thread::spawn or use spawn_blocking inside the spawned future.

tarkah: I've fixed in ada5870 and moved rodio dep to frontend so data isn't dependent on it

tarkah (approved):

tarkah:
I've tested and this approach works. One quirk is we open multiple output streams if multiple notifications are happening at once. I'm not sure this is a big deal, shouldn't happen often and it didn't cause any issues on my end. Definitely not worth the complexity to try and solve with some shared Sink that is only created / dropped after all sounds against it are done.

src/notification.rs
andymandias: Ahh, thank you for catching this!  I initially wrote it as `thread::spawn` but changed it to `tokio::spawn` since I wasn't sure if it was ok to mix `std::thread` with the `tokio` stuff.  And... I neglected to check whether they were actually equivalent.  Sorry about that!

tarkah: I don't think it'll be an issue, these are real threads so the OS is responsible over scheduling and if we have all cores blocked via tokio worker cores + this thread, the OS can handle that (I'm pretty sure)

Issue #429, opened 2024-07-24
Client should not reply to CTCP NOTICE


glguy:
CTCP NOTICE means reply. The client should only reply to queries (sent as PRIVMSG)

Issue #428, opened 2024-07-24
CTCP commands are case-insensitive


glguy:
CTCP commands are specified to be case insensitive. The current conditionals only check for capitalized versions. Most CTCP is unused by this can cause some *actions* to be missed in chat.

andymandias:
<https://modern.ircdocs.horse/ctcp.html> doesn't say anything one way or the other, but it looks like <https://rawgit.com/DanielOaks/irc-rfcs/master/dist/draft-oakley-irc-ctcp-latest.html> does specify case insensitivity (contrary to older specifications).

PR #432, opened 2024-07-24
CTCP Fixes

andymandias:
Fixes for #428, #429, and #431.  Also adds rich text parsing to CTCP `ACTION`s.

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -511,7 +511,7 @@ impl Client {
                                             let _ = self.handle.try_send(command!(
                                             "NOTICE",
                                             user.nickname().to_string(),
-                                            "\u{1}CLIENTINFO CLIENTINFO ACTION CLIENTINFO DCC PING SOURCE VERSION\u{1}"
+                                            "\u{1}CLIENTINFO ACTION CLIENTINFO DCC PING SOURCE VERSION\u{1}"
                                         ));
                                         }
                                         "PING" => {
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -506,42 +506,59 @@ impl Client {
                         {
                             if let Some(query) = ctcp::parse_query(text) {
                                 if matches!(&message.command, Command::PRIVMSG(_, _)) {
-                                    match query.command.as_ref() {
-                                        "CLIENTINFO" => {
-                                            let _ = self.handle.try_send(command!(
-                                            "NOTICE",
-                                            user.nickname().to_string(),
-                                            "\u{1}CLIENTINFO ACTION CLIENTINFO DCC PING SOURCE VERSION\u{1}"
-                                        ));
-                                        }
-                                        "PING" => {
+                                    match query.command {
+                                        ctcp::Command::Action => (),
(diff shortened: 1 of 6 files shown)


glguy (approved):
everything seems to be in order!

the client doesn't show CTCP queries that aren't actions when sent to channel. that's probably intentional but for a channel operator it's not great to not know what's being sent to your channel. If you think that's an issue that's a topic for a different PR, I think.

glguy:
Actually, I found a problem. The PING reply doesn't send the PING command; it just sends the parameters directly without the PING

andymandias:
Thank you for testing!  I really should find a secondary client to use for testing purposes.

I don't have any opinion on non-actions sent to channels.

glguy:
It's normal to have two copies of CLIENTINFO. The first is the response command and the second indicates that that is a supported command. This duplication shows up in the standard.

andymandias:
That was the one command where I had added the command name to response, so when I added another we had three `CLIENTINFO`s.  I alphabetized the response list, so the two `CLIENTINFO`s don't show up next to each other.

data/src/ctcp.rs
```diff
@@ -15,10 +15,13 @@ pub fn parse_query(text: &str) -> Option<Query> {
...
     if let Some((command, params)) = query.split_once(char::is_whitespace) {
-        Some(Query { command, params })
+        Some(Query {
+            command: command.to_uppercase(),
+            params,
+        })
     } else {
         Some(Query {
-            command: query,
+            command: query.to_uppercase(),
```
tarkah, line 24: We are relying on an internal implementation detail from another module, to know we should match on this "uppercase".

Instead, let's define a Command enum in this module and match on this uppercase guarantee _internally_ to produce the strongly typed enum. `client` module can then rely on the strongly typed Command guarantee, making this impossible to misuse.

data/src/client.rs
```diff
@@ -505,42 +505,44 @@ impl Client {
...
+                                            user.nickname().to_string(),
                                             "\u{1}CLIENTINFO ACTION CLIENTINFO DCC PING SOURCE VERSION\u{1}"
                                         ));
+                                        }
+                                        "PING" => {
+                                            let _ = self.handle.try_send(command!(
+                                                "NOTICE",
+                                                user.nickname().to_string(),
+                                                format!("\u{1}PING {}\u{1}", query.params)
+                                            ));
```
tarkah, line 522: I smell a `ctcp::format` method which takes a `ctcp::Command`, `User` and `&str` / params and formats these IRC messages so we don't have to define it here and accidentally miss `\u{1}` or like before, missing the `COMMAND` in the response

andymandias:
Thanks for the review and the instructive suggestions (as usual) @tarkah.  I think they're all implemented at this point.

tarkah (approved):
LGTM! Thanks for tackling these

andymandias:
@glguy if you want to make an issue for the re: showing CTCP queries sent to channels I can take a look at it when I've got some time.  Sounds like it might make sense to provide for operators if we gate it behind a setting.

PR #442, opened 2024-07-25
Minor URL Parsing Tweaks

andymandias:
Two minor tweaks to the URL parsing regex:

1.  Allow TLDs to be up to 63 characters long (that's my reading from [Wikipedia's DNS article](https://en.wikipedia.org/wiki/Domain_Name_System#Domain_name_syntax,_internationalization) which refers to a TLD as the "right-most label" then later states "A label may contain zero to 63 characters").
2.  I *think* there was a typo near the end of the regex (`//` instead of `\/`).  At least, when I entered it into <https://regexr.com/> it gave me an error until I made the change.

Added the URL that didn't parse to the tests.

### DIFF
--- a/data/src/message.rs
+++ b/data/src/message.rs
@@ -16,7 +16,7 @@ use crate::{ctcp, Config, User};
 
 static URL_REGEX: Lazy<Regex> = Lazy::new(|| {
     Regex::new(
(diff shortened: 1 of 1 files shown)


casperstorm (approved):
Looks good to me!

PR #457, opened 2024-08-01
chghost Support

andymandias:
Simple implementation for [chghost](https://ircv3.net/specs/extensions/chghost).  Appears to work in my testing, though it seems like Libera does not send `CHGHOST` messages when your own host changes (instead numeric 396 seems to be sent, which contains a notice that the cloak has been applied).

Would it make sense to add the a setting to mute change host messages? (And/or nickname change messages?)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Added:
 
 - Small icon in sidemenu when a new release is available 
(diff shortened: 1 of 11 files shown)


tarkah (approved):
Very clean, looks great!

PR #462, opened 2024-08-02
Minor URL Parsing Improvements

andymandias:
Updated the URL parsing regex to use the equivalent of the uric BNF from RFC 2396 for the path/query component of URL parsing.


casperstorm (approved):
Awesome!

Issue #466, opened 2024-08-04
Feature Request: Channel buttons in sidebar toggle panes


neutralparadoxdev:
I find myself occasionally clicking the sidebar channels wanting to close them out of same strange belief that would work. I believe this feature would be useful if a user accidently opens the wrong channel. 

If the feature is agreeable, there should be a pull request with this issue with some very minor changes to get it to be at least visually functional.

Thanks for your time,

paradox

#467

Issue #451, opened 2024-07-29
Feature request: Allow clicks on windows in sidebar to toggle restore/minimize


smakkerlak:
Tiny request: Clicking the button for a window that is currently shown and has focus should minimize the window. I'd like to be able to "toggle" the minimized state easily, to help look through channels that i normally keep out of the way.

neutralparadoxdev:
I presume when you say 'window', you mean the pane containing the buffer for server, channel, etc.

I am working on a feature to have the sidebar buttons toggle the panes. It will be a configuration option and pressing the button will open and close the pane. The key thing is that it properly closes the pane. Which means pressing again will open in the next position where a pane would normally be opened. It wont open to its original position.

I can add a feature where pressing the button will only toggle if the pane is focused and closes it. Opening it again will not put it in its original position.

Would this be a suitable implementation for this feature?

smakkerlak:
> I presume when you say 'window', you mean the pane containing the buffer for server, channel, etc.

Yup. General functionality for any pane that has a sidebar button/tab.

Your description seems fine - i can't think of a good way to make panes appear in an expected way consistently, outside of small optimizations like restoring the previous layout if a pane was added and removed without other changes.

andymandias:
Should be completed with #514. Thanks for the suggestion and the help with implementation.

PR #467, opened 2024-08-04
new feature: change behavior of sidebar channel buttons to toggle.

neutralparadoxdev:
clicking on a channel in the sidebar with an open pane closes the pane. Sidebar channel buttons are now effectively toggles.

#466

### DIFF
--- a/book/src/configuration/sidebar.md
+++ b/book/src/configuration/sidebar.md
@@ -4,18 +4,20 @@
 
 ```toml
 [sidebar]
-default_action = "new-pane" | "replace-pane"
+buffer_action = "new-pane" | "replace-pane"
+buffer_focused_action = "close-pane"
 unread_indicators = "dot" | "title" | "none"
 position = "left" | "right" | "top" | "bottom"
 width = <integer>
 ```
 
-| Key                 | Description                                                                            | Default      |
-| ------------------- | -------------------------------------------------------------------------------------- | ------------ |
-| `default_action`    | Action when selecting buffers in the sidebar. Can be `"new-pane"` or `"replace-pane"`. | `"new-pane"` |
-| `unread_indicators` | Unread buffer indicator style.                                                         | `"dot"`      |
-| `position`          | Sidebar position.                                                                      | `"left"`     |
-| `width`             | Specify sidebar width in pixels. Only used if `position` is `"left"` or `"right"`.     | `120`        |
+| Key                     | Description                                                                                                                | Default      |
+| ----------------------- | -------------------------------------------------------------------------------------------------------------------------- | ------------ |
+| `buffer_action`         | Action when pressing buffers in the sidebar. `"new-pane"` opens a new pane, and `"replace-pane"` replaces the focused pane | `"new-pane"` |
+| `buffer_focused_action` | Action when pressing a focused buffer in the sidebar. `"close-pane"` will close the focused pane.                          | `not set`    |
+| `unread_indicators`     | Unread buffer indicator style.                                                                                             | `"dot"`      |
+| `position`              | Sidebar position.                                                                                                          | `"left"`     |
+| `width`                 | Specify sidebar width in pixels. Only used if `position` is `"left"` or `"right"`.                                         | `120`        |
 
 ## `[sidebar.buttons]` Section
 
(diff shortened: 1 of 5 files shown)


andymandias:
I like the idea here, but I think it would be best to make it a configuration option.  I know some users of the `new-pane` and `replace-pane` settings for `default_action` would prefer not to have this behavior, so I suggest we add a `toggle-pane` setting that will cover this use case.

I'd also like to see if we can address #451 in this PR, which is similar but not exactly the same.  If the two ideas can be merged that'd be ideal, but if they're truly incompatible we could add a fourth `default_action`.

neutralparadoxdev:
- Added the associated config option.
- Moved hacked Open code to its own event sequence.
- Added documentation for the config option.\
- Did not tackle #451 as I do not know the required behavior to meet it.
- I carelessly merged...

andymandias (changes requested):
Sorry bitrie, I forgot to submit this earlier.

andymandias:
I removed the `pane_toggling` config since your update to `DefaultAction` means `default_action = "toggle-pane"` will set the configuration.  Otherwise I just reorganized it a bit.

Looks good to me, but let's see if we need to implement a "focus before close" setting.

casperstorm:
I think this is confusing.
1. `new-pane` opens a new pane.
2. `replace-pane` opens in the same pane.
3. `toggle-pane` opens in the new pane like `new-pane` but changes the behavior of focus.

I think we should consider refactoring a bit here now we have the chance.
Also because `default_action` is a terrible word (that is on me!).

What if we instead split it up:
```
[sidebar]
open_buffer = "new-pane" | "replace-pane"
focus_buffer = "close"
```

`open_buffer`: use serde rename to ensure backwards compatibility with `default_action`
`focus_buffer`: would be `Option<FocusBuffer>` which defaults to `None`.

Suggestions? Better names?

casperstorm (changes requested):
I think this is confusing.
1. `new-pane` opens a new pane.
2. `replace-pane` opens in the same pane.
3. `toggle-pane` opens in the new pane like `new-pane` but changes the behavior of focus.

I think we should consider refactoring a bit here now we have the chance.
Also because `default_action` is a terrible word (that is on me!).

What if we instead split it up:
```
[sidebar]
open_buffer = "new-pane" | "replace-pane"
focus_buffer = "close"
```

`open_buffer`: use serde rename to ensure backwards compatibility with `default_action`
`focus_buffer`: would be `Option<FocusBuffer>` which defaults to `None`.

I also think that no matter what we should first FOCUS the pane and then consider the action (`focus_buffer`).

Suggestions? Better names?

andymandias:
If I understand correctly, the first setting would describe the type of open action made when clicking, and the second would describe the action made when clicking on a buffer with the focused state. Since they describe slightly different concepts, maybe they should have distinct styling. Something like, `open_action` and `focused_buffer`? Or `focused_buffer_action`?

casperstorm:
> If I understand correctly, the first setting would describe the type of open action made when clicking, and the second would describe the action made when clicking on a buffer with the focused state. Since they describe slightly different concepts, maybe they should have distinct styling. Something like, `open_action` and `focused_buffer`? Or `focused_buffer_action`?

Yes! What about:
`buffer_action` and `focused_buffer_action` to keep it similar?

neutralparadoxdev:
thanks accepting my suggestion and the help. 

I completely missed the ability to minimize changes by reusing messages regarding opening and close buffers.

andymandias:
> Yes! What about: `buffer_action` and `focused_buffer_action` to keep it similar?

Sounds good to me!  Did you want to take a crack at the refactor @neutralparadoxdev?  Happy to help, of course, but it's all yours if you want.

casperstorm:
@neutralparadoxdev,
I rebased and updated this PR to make it ready for review. Here's what I did:

* Renamed `default_action` to `buffer_action`.
* Added a `buffer_focused_action` which is a optional enum which has `close-pane` variant. By default this is not set.
  * When you set `buffer_focused_action = "close-pane"` it will close a focused buffer, when pressed.

casperstorm (approved):

casperstorm:
@andymandias take a look if this make sense to you - I kinda like this new setting.

andymandias (approved):
LGTM!

casperstorm:
Sorry, this was a mistake. I'll reopen.

PR #469, opened 2024-08-04
`account-notify` and `extended-join` Support

### LINKED ISSUE #465: Failed to auto-join OFTC channels that require nick_password
I'm on 2024.9, which is supposed to contain #346 `Handle ERR_NEEDREGGEDNICK (aka ERR_NOCHANMODES)`. Yet, at least on OFTC, it's not working.

```
#tor-relays Cannot join channel (Need to be identified and verified to join this channel, '/msg NickServ help' to learn how to register and verify.)
```

```
[servers.oftc]
nickname = "removed"
nick_password = "removed"
nick_identify_syntax = "password-nick"
server = "irc.oftc.net"
channels = ["#tor-relays"]
```

I see the PR was only tested on Libera. As I use SASL to identify myself there, I did not test it.

___
EDIT 1:
~I would also propose to let us opt-in (in config.toml) to "delayed join" (`nick_identify_first`) - only ever join after nickserv identified - for a specific server. I don't mind delaying it for all the other channels.~

___
EDIT 2:
I've switched to [SASL (External)](https://www.oftc.net/NickServ/CertFP/), with which auto-join works as expected.

andymandias:
Adds support for IRCv3 [`account-notify`](https://ircv3.net/specs/extensions/account-notify) and [`extended-join`](https://ircv3.net/specs/extensions/extended-join) by adding an `accountname` to users.  Currently the `accountname` is only utilized to determine whether Halloy's user has been registered/logged-in.  Does not fix #465 by itself, since OFTC does not support `account-notify` (or `extended-join`), but I think we should check for these capabilities first before parsing user modes that are not standardized (which afaict is the only notice OFTC gives for that a user has logged in).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 Added:
 
 - Small icon in sidemenu when a new release is available 
-- Enable support for IRCv3 `chghost`
+- Enable support for IRCv3 `chghost`, `account-notify`, and `extended-join`
 
 Removed:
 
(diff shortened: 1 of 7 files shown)


casperstorm (approved):
LGTM!

PR #471, opened 2024-08-05
Re-Join Channels Requiring Registration After Receiving Registered User Mode 

andymandias:
This finishes addressing #465 by rejoining registered-only channels after the user has been granted the `'r'` or `'R'` user mode.  According to the [reference table](https://defs.ircdocs.horse/defs/usermodes) OFTC's registered user mode (`'R'`) is not consistently used that way across servers, so this method is only attempted if `account-notify` is not available.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+Fixed:
+
+- Expanded recognized login notifications (used to join channels that report themselves as requiring registration after logging in)
+
 # 2024.10 (2024-08-04)
 
 Added:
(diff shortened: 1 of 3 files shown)


casperstorm (approved):
Not tested, but looks good.
Do we want CHANGELOG?

andymandias:
Added changelog.  I've only tested with `#tor-relays` on OFTC.  Libera has `account-notify` and doesn't need this patch (and <https://ircv3.net/support/networks> suggests `account-notify` should be fairly common).  Thanks for the review!

PR #474, opened 2024-08-05
Ability to position `Sidebar` top, left, bottom and right.

### LINKED ISSUE #450: Allow docking sidebar to top/bottom
When the sidebar is docked to the side, it allows for a large amount of windows to be shown, but depending on the overall window size, it can be a lot of unused space.

I'd love to be able to dock the sidebar to the top or bottom, since for me this would allow more space for actual channel windows and give an overall neater presentation.

There might be some issues implementing this:
There needs to be a min/max width for each element, possibly configurable, to help handle overflow.
The overflow behavior with the list in a horizontal position should probably be configurable between scrollable and adding multiple rows of buttons.
It might be difficult to distinguish what channels belong to what server, especially if there is multiple rows of buttons. It could be great with a visual indication of relationship. In a perfect world, related windows would be bunched together across multiple rows and not just treated as text, where buttons are lined up with a linebreak here and there.

My setup and why I'd love to be able to dock to the top:
image

Example from a different irc client:
image

For anyone wondering, my secondary monitor is rotated and the remaining part of the screen is the perfect size for a browser with a video :)

casperstorm:
Fixes #450 

Screenshot 2024-08-05 at 23 32 13

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@ Fixed:
 - Expanded recognized login notifications (used to join channels that report themselves as requiring registration after logging in)
 - Messages with multiple targets are correctly recorded into multiple buffers (and/or multiple times into the same buffer) client-side.
 - Messages sent with a STATUSMSG prefix are recorded and indicated in the corresponding channel.
+- Ability to position the sidebar at the top, bottom, right, or left. See [Sidebar configuration](https://halloy.squidowl.org/configuration/sidebar.html).
+
+Changed:
+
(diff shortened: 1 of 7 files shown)


tarkah (approved):
LGTM! Great feature, I'll roll with this myself since I only have a few channels

dmd:
This doesn't appear to be documented.

dmd:
Have you tested this with "right" and "bottom"?

"top" seems to work for me, but "right" and "bottom" do not.

casperstorm:
> Have you tested this with "right" and "bottom"?
> 
> "top" seems to work for me, but "right" and "bottom" do not.

Thank you. It seems like a regression. I'll fix this right now.

PR #486, opened 2024-08-10
Feat/right aligned nick

### LINKED ISSUE #477: Feature request: Allow nicks to be right-aligned in the main channel buffer
As discussed on the #halloy channel, it would be nice if halloy had the ability to right align nicks as I find this improves readability.

Here's a script which does this for irssi: https://github.com/irssi/scripts.irssi.org/blob/master/scripts/nm2.pl

I think it uses the logic of calculating the longest nick then padding each line so that each message starts at the same point.

Here's an example of how it looks in other clients:

weechat2

tarkah:
Resolves #477

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 Added:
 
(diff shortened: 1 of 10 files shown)


casperstorm (approved):
Awesome addition, looks great.

PR #488, opened 2024-08-11
Fix/437 mode requires too many parameters

### LINKED ISSUE #437: /mode requires too many parameters
`/mode #channel` can be used to check the modes of a channel (even if you're not in that channel) and `/mode user` can be used to check modes of a user. The <mode> parameter should be optional.

KaiKorla:
Try to fix #437

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -1519,7 +1519,7 @@ fn remove_tag(key: &str, tags: &mut Vec<irc::proto::Tag>) -> Option<String> {
 fn start_reroute(command: &Command) -> bool {
     use Command::*;
 
-    matches!(command, WHO(..) | WHOIS(..) | WHOWAS(..))
+    matches!(command, WHO(..) | WHOIS(..) | WHOWAS(..) | MODE(..))
 }
 
 fn stop_reroute(command: &Command) -> bool {
@@ -1535,7 +1535,10 @@ fn stop_reroute(command: &Command) -> bool {
                 | ERR_NOSUCHSERVER
                 | ERR_NONICKNAMEGIVEN
                 | ERR_WASNOSUCHNICK
-                | ERR_NEEDMOREPARAMS,
+                | ERR_NEEDMOREPARAMS
+                | ERR_USERSDONTMATCH
+                | RPL_UMODEIS
+                | ERR_UMODEUNKNOWNFLAG,
             _
         )
     )
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -1518,7 +1518,7 @@ fn remove_tag(key: &str, tags: &mut Vec<irc::proto::Tag>) -> Option<String> {
 
 fn start_reroute(command: &Command) -> bool {
     use Command::*;
-    
+
     if let MODE(target, _, _) = command {
         !proto::is_channel(target)
     } else {
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -949,7 +949,7 @@ impl Client {
                     }
                 }
             }
-            Command::MODE(target, Some(modes), args) => {
+            Command::MODE(target, Some(modes), Some(args)) => {
                 if proto::is_channel(target) {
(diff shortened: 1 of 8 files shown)


andymandias (changes requested):
This is a good start!  I think we should handle at least a few more of the return messages while we're at it.  In particular, let's add some contextual info for `RPL_UMODEIS` to `data/src/message.rs` (similar to `RPL_CHANNELMODEIS`).  And the mode command tooltips should be updated in `src/buffer/input_view/completion.rs` to reflect the new capabilities.

KaiKorla:
@andymandias, i updated the PR. Tell me if the nested `if` is ok. Otherwise i would try with `match`.

andymandias (changes requested):
Some minor consistency changes and a small bug fix, but overall looks good.  If structure and regex look good to me.

Testing it out, there is one more change I think we should make before requesting final approval.  When requesting for a user (i.e. `/mode <ourself>`) the output is currently sent to the server buffer, which could be confusing if you're not already in the server buffer (since there isn't a visual indication of where the message went).  So, let's reroute the user mode replies to the current buffer (similar to how `WHOIS` is handled).

To do that, I believe we just need two changes in `client.rs`:
1.  `start_reroute` should return `true` if `command` is `MODE` and  `proto::parse_channel_from_target(target).is_none()` (i.e. `target` is not a channel).
2.  `stop_reroute` should return true for every possible response to `MODE` when `target` is not a channel (`ERR_NOSUCHNICK`, `ERR_USERSDONTMATCH`, `RPL_UMODEIS`, or `ERR_UMODEUNKNOWNFLAG`).

And, a couple style requests.  `rustfmt` will help keep style consistent throughout, and I find it quite helpful (even if I don't *always* agree with its choices).  I believe the preferred method for updating a Halloy fork is `rebase` (over `merge`), just to keep the commit history as clean as possible.

KaiKorla:
Could need some help with the rerouting. :-/

data/src/client.rs
```diff
@@ -1519,7 +1519,7 @@ fn remove_tag(key: &str, tags: &mut Vec<irc::proto::Tag>) -> Option<String> {
 fn start_reroute(command: &Command) -> bool {
     use Command::*;
 
-    matches!(command, WHO(..) | WHOIS(..) | WHOWAS(..))
+    matches!(command, WHO(..) | WHOIS(..) | WHOWAS(..) | MODE(..))
```
andymandias, line 1522: For this to only apply to user `MODE` commands, we should look at the `target` like 
```rust
    if let MODE(target, _, _) = command {
        !proto::is_channel(target)
    } else {
        matches!(command, WHO(..) | WHOIS(..) | WHOWAS(..))
    }
```

data/src/message.rs
```diff
@@ -769,6 +770,16 @@ fn content(
...
         }
+        Command::Numeric(RPL_UMODEIS, params) => {
+            let mode = params
+                .iter()
+                .skip(1)
+                .map(|s| s.as_str())
+                .collect::<Vec<_>>()
+                .join(" ");
+
+            Some(parse_fragments(format!(" ∙ User mode is {mode}")))
```
andymandias, line 781: Due to PR #496 this should now be
```rust
            Some(parse_fragments(format!("User mode is {mode}")))
```

andymandias:
Looks good to me!  After those last couple small changes I think it'll be ready for a final review. :slightly_smiling_face:

data/src/client.rs
KaiKorla: Understood, done.

data/src/message.rs
KaiKorla: Fixed.

andymandias (approved):
Looks good to me.  Anything to add, @casperstorm?

casperstorm (changes requested):
Please don't back merge, but rather rebase. It is currently not in a great state, and impossible to review.

tarkah (approved):

casperstorm (approved):

Issue #438, opened 2024-07-24
Route STATUSMSG to the correct buffer


glguy:
Messages sent with a STATUSMSG prefix are presented as normal PMs instead of as channel messages.

- A message sent to `+#libera` should show up in channel but indicate that it's only readable to users with +v
- A message sent to `@#libera` should show up in channel but indicate that it's only readable to users with +o

Note that this can apply to PRIVMSG, NOTICE, and even CTCP ACTIONs

https://modern.ircdocs.horse/#statusmsg-parameter

I happen to render these messages with a special prefix, but obviously you'd want to pick whatever makes the most sense for your UI so that the reader knows that it's a restricted message

<img width="392" alt="Screenshot 2024-07-24 at 1 23 20 PM" src="https://github.com/user-attachments/assets/08fc5b5d-85ec-4b29-922a-36cbdb114715">

PR #490, opened 2024-08-12
STATUSMSG Support

andymandias:
Shows messages with a STATUSMSG prefix in the corresponding channel, and adds an indicator for those messages to the left of the nickname.  I put the indicator to the left of the nickname so that it couldn't be spoofed by message text.  I believe this addresses #438.

### DIFF
--- a/data/src/message.rs
+++ b/data/src/message.rs
@@ -90,11 +90,11 @@ pub enum Target {
 }
 
 impl Target {
-    pub fn prefix(&self) -> &Option<char> {
+    pub fn prefix(&self) -> Option<&char> {
         match self {
-            Target::Server { .. } => &None,
-            Target::Channel { prefix, .. } => prefix,
-            Target::Query { .. } => &None,
+            Target::Server { .. } => None,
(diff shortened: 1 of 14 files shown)


casperstorm (approved):
LGTM, @tarkah any feedback?

data/src/message.rs
```diff
@@ -75,12 +75,29 @@ impl From<Encoded> for proto::Message {
...
 }
 
 impl Target {
+    pub fn prefix(&self) -> &Option<char> {
+        match self {
+            Target::Server { .. } => &None,
+            Target::Channel { prefix, .. } => prefix,
+            Target::Query { .. } => &None,
+        }
+    }
```
tarkah, line 99: Nit: Typically you'd return Option<&T>, which you can do by prefix.as_ref()

tarkah (approved):

tarkah:
LGTM, I haven't tested to see how this looks visually but code looks good  👍

data/src/message.rs
andymandias: Makes perfect sense!  Thanks for the tip :)

PR #492, opened 2024-08-13
Exclude Trailing Punctuation from URLs

andymandias:
My understanding is that URLs technically can end in punctuation, but in my experience it is extremely rare.  So I think we should err on the side of assuming punctuation at the end of a URL is not part of the URL, which is what this PR does.

The regex got so long and unwieldy I broke it up into smaller components to make it more legible, but the only way I could find to do that involved adding the [`const_format`](https://docs.rs/const_format/latest/const_format/) crate.  The names `RESERVED` and `UNRESERVED` come from [RFC 2396](https://www.ietf.org/rfc/rfc2396.txt).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -749,6 +749,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "const_format"
+version = "0.2.32"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


casperstorm (approved):
I agree with this, and think its a good solution

PR #496, opened 2024-08-17
Right Align Miscellaneous Messages

andymandias:
This tweaks how server messages, internal messages, and actions are displayed in order to line up with right-aligned nicknames.  

screenshot

Messages already in history will get an extra `∙` as this is now.  History processing could be added to avoid this, but I'm not sure if that juice is worth the squeeze.


casperstorm (approved):
LGTM!

Issue #498, opened 2024-08-24
Autoconnecting to #cinemageddon behind ZNC causes crash


trfunk:
## What happens
When I launch halloy it connect to everything and then the program closes without any error message or so.  This happens whens when I autoconnect to the #cinemageddon channel on irc.tv-vault.me via znc. Joining the channel after being already connected for a bit to cinemageddon works just fine. This only happens if the connection happens via znc. All other networks from that znc work without any issues. Hexchat connects to znc just fine.

## Configuration
**Server:**
**OS:** debian 12
**ZNC:** ZNC 1.8.2+deb3.1+deb12u1
**Network:** irc.tv-vault.me +6697
- ssl enabled
- sha256 fingerprint of ssl certificate: e2:18:87:bf:88:37:6b:6c:7b:3f:39:3c:c3:59:f4:36:e3:33:18:5b:e4:32:73:d0:7e:9a:67:8b:fa:6a:92:b3
- logging into nickserv via cert module

Image of the server settings, in case i missed something: https://i.postimg.cc/cxB9wQLt/znc-config.png

**Client:**
**OS:** Windows 11
**Version:** Halloy 2024.10
**Config:**
```
[servers.TVVault]
nickname = "<name>/TVVault"
server = "<znc domain>"
port = 7001
password = "<name>/TVVault:<password>"
```

## Reproduction
1. Setup ZNC to autoconnect to #cinemageddon on irc.tv-vault.net
2. #cinemageddon requires users to be registered to services before joining
3. add configuration for znc bouncer in config.toml for halloy
4. start halloy
5. observe crash

## Log
( I stripped everything that was related to other networks)
```
23:38:42.639:INFO -- halloy 2024.10 (4ff56fd-dirty) has started
23:38:42.640:INFO -- config dir: "C:\\Users\\<name>\\AppData\\Roaming\\halloy"
23:38:42.640:INFO -- data dir: "C:\\Users\\<name>\\AppData\\Roaming\\halloy"
23:38:42.827:INFO -- Settings {
    present_mode: AutoVsync,
    backends: Backends(
        VULKAN | GL | METAL | DX12 | BROWSER_WEBGPU,
    ),
    default_font: Font {
        family: Name(
            "Iosevka Term",
        ),
        weight: Normal,
        stretch: Normal,
        style: Normal,
    },
    default_text_size: Pixels(
        13.0,
    ),
    antialiasing: None,
}
23:38:42.915:INFO -- Available adapters: [
[...]

PR #503, opened 2024-08-26
Issue #498 Fix

andymandias:
When a server has an unrecognized access level character (in the case of #498, `$`) then access level parsing could end prematurely (leaving unparsed access levels at the beginning of the rest of the string).  Since `@` is a valid access level, it could then be found in the remaining string before `!` leading to a panic from `rest[i + 1..j].to_string()`.  This switches to looking for a valid first nickname character, then parsing everything before it as an access level and everything after as rest.  As a fallback, if `i` is ever found to be greater than or equal to `j` then an error is produced instead of evaluating `rest[i + 1..j].to_string()`.

Edit: criteria for valid first character was taken from [RFC 2812](https://www.rfc-editor.org/rfc/rfc2812.txt).

### DIFF
--- a/data/src/user.rs
+++ b/data/src/user.rs
@@ -69,25 +69,33 @@ impl<'a> TryFrom<&'a str> for User {
             return Err("nickname can't be empty");
         }
 
-        let access_levels = value
-            .chars()
-            .map(|c| AccessLevel::try_from(c).ok())
-            .take_while(Option::is_some)
-            .flatten()
-            .collect::<HashSet<_>>();
+        let Some(index) = value.find(|c: char| c.is_alphabetic() || "[\\]^_`{|}".find(c).is_some())
+        else {
+            return Err("nickname must start with alphabetic or [ \\ ] ^ _ ` { | }");
+        };
 
-        // Safe as access levels are just ASCII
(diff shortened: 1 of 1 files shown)


casperstorm:
Would it make sense to add some tests to this?

tarkah:
Changes make sense. Would we rather return _something_ as the nickname even if it's somehow parsed wrong than returning errors? Error == message discarded which is much harder for the end user to know something is incorrect vs showing a potentially invalid (includes unknown access chars) nickname?

andymandias:
Makes sense to me.  Returns my reckoning of best guess for nickname/username if it winds up finding `@` before `!`.

Tests with stock user `nick!user@host` as well as the one I came across when testing the issue.  As far as I can tell `assert_eq!` uses the default equality operator for the struct, so `assert_eq!(user, expected)` would only test their `nickname` components.  That's why it's testing components individually.

casperstorm:
Do you want this reviewed now, @andymandias?

andymandias:
Ah! Sorry, yes. I think this is ready for review now @casperstorm  😅

casperstorm (approved):

PR #507, opened 2024-08-30
Show Errors from Loading Sounds

andymandias:
When configuring the `sound` setting for a notification to a non-existent sound, 

```toml
[notifications.highlight]
show_toast = true
sound = "no-such-sound-file"
```
no error message is shown to the user.  Instead the welcome screen is shown.  These changes aim to show the error screen, and to make the error actionable.

### DIFF
--- a/data/src/audio.rs
+++ b/data/src/audio.rs
@@ -21,7 +21,7 @@ impl Sound {
             internal.bytes()
         } else {
             let Some(sound_path) = find_external_sound(name) else {
-                return Err(LoadError::NoSoundFound);
+                return Err(LoadError::NoSoundFound(name.to_string()));
             };
 
             read(sound_path)?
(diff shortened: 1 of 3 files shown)


casperstorm (changes requested):
https://github.com/squidowl/halloy/blob/70833a29bc6f97a145365a66bd2d5a68b9016bac/data/src/config.rs#L308-L310

I say we change this to `#[error(transparent)]` so we get
```
sound "foobar.ogg" was not found
```

instead of 

```
error loading sound: sound "foobar.ogg" was not found
```

andymandias:
> I say we change this to `#[error(transparent)]`

Sounds like a good idea to me!

I also added some more detail to the error message (where the sounds directory is located) to help reduce potential confusion.  In testing I had tried to set an absolute path, and took me a minute to realize why it didn't work.

casperstorm (approved):

casperstorm:
Awesome man!

PR #509, opened 2024-08-31
Context sensitive auto-complete

### LINKED ISSUE #489: context-sensitive nickname tab completion
Weechat does something smart when nicknames are tab-completed which I think halloy should copy.

- if a nick is completed at the beginning of the input line, `: ` (colon space) is appended to the completion
- otherwise, a space is appended to the completion.

I don't think this even needs to be an option; I honestly can't think of any reason not to just make it the sole behavior.

It looks you already thought of this (except putting a space after the nick in the 2nd cases):

https://github.com/squidowl/halloy/blob/890da2d8c2db931639dec45a0a38b5f75002eb6f/src/buffer/input_view.rs#L301-L307

... but it is only for what happens when someone clicks on a nick in the nicklist, which ... does anyone do that?

I looked into implementing this myself (even though I don't know Rust) but got stymied because by the time `tab` is called, it no longer knows whether `filtered` was populated by `process_users` or `process_channels`. I added such a flag to the Text struct but then realized `tab` doesn't have the `input` context and at that point I realized if you want to do this you could do it in 30 seconds because you actually know how this code works (I guess you'd check if input == prompt to see if we're at the beginning of the line??)

casperstorm:
Fixes #489.

Behaviour copied from Weechat.

> If a nick is completed at the beginning of the input line, :  (colon space) is appended to the completion. Otherwise, a space is appended to the completion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,12 +23,13 @@ Changed:
   - **NOTE** Old theme files are not compatibile with the new format. However all the themes in the [theme community](https://halloy.squidowl.org/configuration/themes/community.html) has been updated to the new format.
 - Unread indicator has changed from a boolean value to a enum. See [Sidebar configuration](https://halloy.squidowl.org/configuration/sidebar.html).
 - Renamed `sidebar.default_action` to `sidebar.buffer_action`.
-
-
+- Auto-completing (with tab) a nickname at the beginning of the input line will append ': ' (colon space). Otherwise, a space is appended to the completion.
+ 
 Removed:
 
 - Removed `hex` configuration option for server messages and nicknames.
   - `hex` was previously used to overwrite a color value. This is now done through the new theme format.
+
 
 
 # 2024.10 (2024-08-04)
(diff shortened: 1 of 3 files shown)


andymandias:
Looks good to me!  Two minor QoL changes I think would be nice to add:
1.  Let shift-tab cycle backwards through completions.  If you overshoot the name you wanted to autocomplete to it might take a while to cycle all the way back around.
2.  I think it would nice if after cycling through all of the possible completions it, tab would revert to the prompt the user entered.  E.g. `cas` (tab)→ `casperstorm` (tab)→`cassandra` (tab)→ `cas`.

casperstorm:
> Looks good to me! Two minor QoL changes I think would be nice to add:
> 
> 1. Let shift-tab cycle backwards through completions.  If you overshoot the name you wanted to autocomplete to it might take a while to cycle all the way back around.
> 2. I think it would nice if after cycling through all of the possible completions it, tab would revert to the prompt the user entered.  E.g. `cas` (tab)→ `casperstorm` (tab)→`cassandra` (tab)→ `cas`.

1. I like this 👍🏻 
2. I initially removed this as Weechat didn't do it. I am ok reverting.

andymandias (approved):
Looks good to me!

Issue #491, opened 2024-08-13
Trigger notifications on direct message


cvengler:
Currently, there is no way to trigger a notification when a direct message is received.
I would like to have a feature where a notification gets triggered when I receive a direct message, either as a `[notification.direct]` option or as an additional feature of `[notification.highlight]`

casperstorm:
This has been released in 2024.11. See configuration here: https://halloy.squidowl.org/configuration/notifications.html.

PR #510, opened 2024-08-31
notifications for direct messages

hugegulp:
Addresses feature request: 
[Trigger notifications on direct message](https://github.com/squidowl/halloy/issues/491)

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -1315,6 +1315,10 @@ impl HighlightBlackout {
 pub struct Map(BTreeMap<Server, State>);
 
 impl Map {
+    pub fn len(&self) -> usize {
+        self.0.len()
+    }
+
     pub fn disconnected(&mut self, server: Server) {
         self.0.insert(server, State::Disconnected);
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@ Fixed:
 - Expanded recognized login notifications (used to join channels that report themselves as requiring registration after logging in)
 - Messages with multiple targets are correctly recorded into multiple buffers (and/or multiple times into the same buffer) client-side.
 - Messages sent with a STATUSMSG prefix are recorded and indicated in the corresponding channel.
+- Ability to position the sidebar at the top, bottom, right, or left. See [Sidebar configuration](https://halloy.squidowl.org/configuration/sidebar.html).
+
+Changed:
+
+- Unread indicator has changed from a boolean value to a enum. See [Sidebar configuration](https://halloy.squidowl.org/configuration/sidebar.html).
+
 
 # 2024.10 (2024-08-04)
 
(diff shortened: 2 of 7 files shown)


data/src/client.rs
```diff
@@ -636,6 +637,16 @@ impl Client {
                             // If we sent (echo) & context exists (we sent from this client), ignore
                             return None;
                         }
+
+                        // send direct message notification
+                        // TODO: check if the active buffer is the same as who sent the message
```
qauff, line 642: @tarkah @casperstorm @andymandias is there existing code/example that checks/references the active/focused buffer's information?

casperstorm: Take a look here @hugegulp: https://github.com/squidowl/halloy/blob/dbac8aeb4910095760d93ef19eab109eb3d7864c/src/screen/dashboard/sidebar.rs#L281-L290

`panes` and `focus` is in the `Dashboard` struct. I think what you could do is make a helper `fn` on the `Dashboard` struct which returns the current selected `Pane` (essentially refactoring that piece of code i linked into a `fn`). You can then check the `buffer` of that pane to see if its the one you got a message for.

Once you have such a method you can call it on the `dashboard` which you can get like this:
https://github.com/squidowl/halloy/blob/f0a4e8e4baeea72dfbe1c48f93e29ba1eb03dc70/src/main.rs#L408-L410

casperstorm: Alternative, and potentially better, take a look at:

https://github.com/squidowl/halloy/blob/10fde389a92edf4b1f29e2a132abbb771fc88fa0/data/src/history/manager.rs#L250-L265

hugegulp: This is a great suggestion. I went with the history manager method! I added the conditional check in `src/main.rs` let me know what you think!

hugegulp:
The only scenario I haven't covered for receiving a notification sound is when halloy itself is not in focus. I've noticed when I have a DM/query open and I navigate away from halloy, when I get a message, the notification doesn't play. So next stage of this feature would be to figure out if halloy is in focus. Is that possible currently, @casperstorm?

hugegulp:
> So next stage of this feature would be to figure out if halloy is in focus.

I guess another way to look at this is, to make it so the history manager counts messages as unread if halloy (and panes) are not in focus.

casperstorm:
https://github.com/iced-rs/iced/blob/master/core/src/window/event.rs#L39-L43

We can detect if a window has focus with these events.

hugegulp:
> https://github.com/iced-rs/iced/blob/master/core/src/window/event.rs#L39-L43
> 
> We can detect if a window has focus with these events.

Thanks for this @casperstorm I've pushed up an implementation of tracking the window's focus, and triggering notifications for DMs if the window is out of focus! We should be good to go now! :pray:

hugegulp:
Apologies for all the little refactors! _Now_ I'm ready for review :pray:

casperstorm (approved):
This is really well done, @tarkah you have any feedback before merge?

tarkah (approved):
Great addition, thanks! Added a couple small changes.

tarkah:
Fuck I messed up pushing my changes... will re-open

PR #513, opened 2024-09-02
IRCv3 Monitor

andymandias:
An initial implementation for [Monitor](https://ircv3.net/specs/extensions/monitor).  Negotiates the Monitor capability, adds `MONITOR` to the command completion, a monitor configuration setting to monitor users on connection, customization of Monitor messages, and configuration options to get notifications when a monitored user goes offline/online.  Since Monitor messages appear in the server buffer, I made it so the server buffer's globe icon gets the unread indicator color when they are added to the buffer.  Works in my testing, but I would like to get thoughts on whether it should operate differently.


data/src/config/server.rs
```diff
@@ -86,6 +86,9 @@ pub struct Server {
         deserialize_with = "deserialize_duration_from_u64"
     )]
     pub who_retry_interval: Duration,
+    /// A list of nicknames to monitor (iff MONITOR is supported by the server).
```
casperstorm, line 89:
```suggestion
    /// A list of nicknames to monitor (if MONITOR is supported by the server).
```

casperstorm:
I like this, @andymandias - not sure how I feel about the new subcommand added tho 🤔 I would like @tarkah feedback here.

andymandias:
Originally I had set up all of the monitor commands as individual completions, but having a space in the middle of the command messed with the existing completion implementation.  The implementation I went with was an attempt to provide a tooltip in between `MONITOR` and each of the individual subcommands (i.e. the user could type in `/MONITOR` then from the tooltip learn what each of the available `MONITOR` commands are).  Not particularly attached to it, nor the specifics of its implementation.

tarkah (approved):
This looks great, very clean and simple to follow!

As for `subcommand` in completions, I don't have too much opinion. Current implementation doesn't handle nested subcommands but I don't think that's a big deal, we can cross that bridge if something needs it.

I think we can probably handle this more gracefully with a bigger completions refactor, but I think that should happen outside this unit of work.

For now, I'm fine w/ the change since it is a nice QOL improvement for updating args / tooltips based on the first arg value.

casperstorm (approved):

PR #517, opened 2024-09-02
port v1 dracula theme

hugegulp:
the new themes engine works great. the new dracula theme is awesome too, but there was a couple of things I missed from v1, like the green border. I figured I'd share it here too in case anyone wants that theme too.

I took the new dracula theme, and just changed a few colours from v1.


casperstorm (changes requested):
As im not a regular Dracula user myself i am happy to change yours with my edit. I think its better than having multiple edits of a theme. Could you replace what you have instead of naming it "v1"?

hugegulp:
> As im not a regular Dracula user myself i am happy to change yours with my edit. I think its better than having multiple edits of a theme. Could you replace what you have instead of naming it "v1"?
:pray: okay! I've replaced the existing dracula theme with the one I ported and based off the existing one :saluting_face:

casperstorm (approved):

PR #531, opened 2024-09-03
Allow Zero-Arguments Commands without Trailing Space

andymandias:
When attempting to submit a command that does not require any arguments (e.g. `/list`), if the user attempts to submit the command without a trailing space then it is rejected with the error message "missing args".  The command submits as expected if a space is added after the command name (e.g. `/list `).  This PR is just to remove the need for a trailing space after the command.


casperstorm (approved):
Tested. Works great.

tarkah (approved):

Issue #528, opened 2024-09-03
message to *status ends up in a pane for status


dmd:
If I message `*status` (part of the znc bouncer):

1. the actual privmsg goes to the right place
2. the outgoing message appears in a pane for `status`
3. the incoming replies appear in a pane for `*status`

dmd:
Yes, this fixes it. (I don't know what your policy is on closing issues - do I close it when I see my issue is fixed, or do you close it when you do the release?)

PR #540, opened 2024-09-05
ZNC Nickname Support for Outgoing Messages

andymandias:
This allows nicknames to start with `*` in `User::try_from`, which addresses #528 in my testing.


casperstorm (approved):
Deserves a small CHANGELOG entry, then its good to go.

PR #541, opened 2024-09-06
Feat/theme editor

tarkah:
Adds an in-app theme editor with the ability to save the theme and share it via a `halloy://` URL. Clicking those links will cause that theme to open in the theme editor, where users can subsequently save it.

Oh yeah, adds multi window support to show theme editor :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: 

https://github.com/user-attachments/assets/723e1ed0-a758-4f92-8330-f60322db8751


casperstorm (approved):
Adjust UI slightly and tested on macOS and Windows. Works perfectly.
This is SUCH a nice feature. Before we merge I'd like to add a entry to the book.

PR #542, opened 2024-09-06
Exclude and include channels on Server Messages

casperstorm:
This PR solves the following problems:

1. You want to see `join` in all channels except `#linux` and `#halloy`.
2. You dont want to see `part` in any channels except `#linux` and `#halloy`.

This is now solved by `exclude` and `include`

```toml
# 1. see join messages in all channels except #linux and #halloy.
[buffer.server_messages.join]
exclude = ["#linux", "#halloy"]
```

```toml
# 2. don't see part messages in any channels except #linux and #halloy.
# The "*" (or "all") is a wildcard which means all channels.
[buffer.server_messages.join]
exclude = ["*"]
include = ["#linux", "#halloy"]
```

If the channel is `included`, it has precedence over `excluded` and `enabled` is above both so the following wont show any message at all:

```toml
# dont see any quit messages in any channel
[buffer.server_messages.quit]
enabled = false
include = ["*"]
```

- [x] Update CHANGELOG
- [x] Update book

### DIFF
--- a/data/src/config/buffer.rs
+++ b/data/src/config/buffer.rs
@@ -65,6 +65,10 @@ pub struct ServerMessage {
     pub smart: Option<i64>,
     #[serde(default)]
     pub username_format: UsernameFormat,
+    #[serde(default)]
+    pub excluded: Vec<String>,
+    #[serde(default)]
+    pub included: Vec<String>,
 }
 
 impl Default for ServerMessage {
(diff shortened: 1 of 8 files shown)


data/src/config/buffer.rs
```diff
@@ -73,7 +77,36 @@ impl Default for ServerMessage {
...
             smart: Default::default(),
             username_format: UsernameFormat::default(),
+            exclude: Default::default(),
+            include: Default::default(),
+        }
+    }
+}
+
+impl ServerMessage {
+    pub fn should_send_message(&self, channel: Option<&str>) -> bool {
```
tarkah, line 87: Nit: Remove Option here and change how it's handled at call site. The option has no need within this funciton

data/src/config/buffer.rs
```diff
@@ -73,7 +77,36 @@ impl Default for ServerMessage {
...
+
+        let channel_included = is_channel_filtered(&self.include, channel);
+        let channel_excluded = is_channel_filtered(&self.exclude, channel);
+
+        // If the channel is included, it has precedence over excluded.
+        if channel_included || !channel_excluded {
+            return true;
+        }
+
+        false
```
tarkah, line 109: Nit: Return `channel_included || !channel_excluded` directly as it resolving to `true / false` is returning `true / false`

casperstorm: Doh, of cause. Thanks.

tarkah (approved):

Issue #545, opened 2024-09-09
URL parser choking on Umlauts


trfunk:
Hello,
the URL parser doesn't seem to handle umlauts at all very well. For example `https://www.reddit.com/r/witze/comments/1fcoz5a/ein_vampir_auf_einem_tandem_gerät_in_eine/` breaks just infront of the ä and sends you to `https://www.reddit.com/r/witze/comments/1fcoz5a/ein_vampir_auf_einem_tandem_ger` when you click it.

If the part infront of the tld contains an Umlaut, the URL is not parsed at all, see `http://öbb.at` (not detected as link),  `http://xn--bb-eka.at/` works, `http://oebb.at/` also works.

Thank you!

andymandias:
I can look at this tonight, yeah.

PR #546, opened 2024-09-10
Unicode URL Parsing

andymandias:
Updates the regex to use the Unicode letter and number character classes instead of ASCII letter and number ranges.  Also adds the examples provided in issue #545 as tests.

The conversion from `Regex::new` to `RegexBuilder::new` is because `Regex::new` runs into the [`CompiledTooBig` error](https://docs.rs/regex/latest/regex/enum.Error.html#variant.CompiledTooBig) because Unicode letter and number character classes are much larger than their ASCII counterparts.  The documentation warns that this error is a potential sign of a slow regex.  For now I bumped the `size_limit`, but an alternative could be to simplify the regex.  For example, if we switch the domain parsing from `{1,256}` and `{1,63}` to `+` and `+`, then the regex would again fit within the default `size_limit` (the URL parsing would be less strict as a result, treating potential non-URLs as URLs).

### DIFF
--- a/data/src/message.rs
+++ b/data/src/message.rs
@@ -6,7 +6,7 @@ use irc::proto;
 use irc::proto::Command;
 use itertools::Itertools;
 use once_cell::sync::Lazy;
-use regex::Regex;
+use regex::{Regex, RegexBuilder};
 use serde::{Deserialize, Deserializer, Serialize};
 use url::Url;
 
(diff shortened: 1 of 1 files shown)


casperstorm (approved):

tarkah (approved):
Thanks!

Issue #551, opened 2024-09-11
Load the configuration file in an async fashion


4e554c4c:
Currently loading the configuration file can take some time, which will block the UI of halloy. If features that involve running commands in the configuration file are added, such as #424, then the UI can bock for even longer. Therefore loads should be async.

PR #552, opened 2024-09-11
Add `password_command` config option

### LINKED ISSUE #424: password-cmd
Would be nice addition if passwords could be supplied with a command given by `password-cmd` like in many other applications. E.g. I use https://www.passwordstore.org/ as my password manager.

4e554c4c:
closes #424

related to #551  but I don't think that fix should be done in this PR, since it the changes involved are largely orthogonal to this one.

### DIFF
--- a/data/src/server.rs
+++ b/data/src/server.rs
@@ -78,7 +78,8 @@ impl Map {
             if let Some(pass_file) = &config.password_file {
                 if config.password.is_some() || config.password_command.is_some() {
                     return Err(Error::Parse(
-                        "Only one of password, password_file and password_command can be set.".to_string(),
+                        "Only one of password, password_file and password_command can be set."
+                            .to_string(),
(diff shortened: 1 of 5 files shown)


data/src/server.rs
```diff
@@ -72,17 +73,37 @@ impl Map {
...
+                if config.password.is_some() {
+                    return Err(Error::Parse(
+                        "Only one of password, password_file and password_command can be set.".to_string(),
+                    ));
+                }
+                let output = if cfg!(target_os = "windows") {
+                    Command::new("cmd")
+                        .args(["/C", pass_command])
+                        .output()
+                        .expect("failed to execute process")
```
tarkah, line 97: We shouldn't panic here, instead add it to `Error` and throw `?` it

tarkah (approved):

PR #554, opened 2024-09-12
Fix panic when MOTD set by user w/o username

ljrk0:
Some servers use bots without a valid IRC username (but a nickname) for MOTD. This results in a panic:

	thread 'main' panicked at data/src/client.rs:1047:53:
	called `Option::unwrap()` on a `None` value
	thread 'main' panicked at /Users/janis.koenig/.cargo/git/checkouts/iced-f01cba4d5e61fd0a/9628dc2/winit/src/program.rs:467:43:
	Send event: SendError { kind: Disconnected }

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -1044,7 +1044,7 @@ impl Client {
 
                     channel.topic.who = message
                         .user()
-                        .map(|user| user.username().unwrap().to_string());
+                        .map(|user| user.nickname().to_string());
                     channel.topic.time = Some(server_time(&message));
                 }
             }


casperstorm:
Hmm this seems like a bug we were using `username` (and `unwrapping`!).
Why don't we just use `nickname`, or `formatted` if we really want username and host if available?

Good find, @ljrk0 and thanks for picking it up.

ljrk0:
I have no preferences with respect to that and I think using `nickname` or suchlike would work just as well -- I've simply tried to retain old behavior for those MOTDs that provide a valid `username` :)

I can adjust the PR to simply always use `.formatted()` or `.nickname()` if that'd be preferred?

Thank you for building my favorite IRC client! :3

ljrk0:
Great!  I've changed the patch and tested on my servers and everything seems fine :)

Always happy to be able to help a little bit!

casperstorm (approved):

PR #556, opened 2024-09-12
Remove irc stream task from iced stream which has backpressure

tarkah:


data/src/stream.rs
```diff
@@ -75,7 +75,12 @@ pub fn run(
 ) -> impl futures::Stream<Item = Update> {
     let (sender, receiver) = mpsc::unbounded();
 
-    let runner = stream::once(_run(server, proxy, sender)).map(|_| unreachable!());
+    let runner = stream::once(async {
+        // Spawn to unblock backend from iced stream which
+        // has backpressure
+        tokio::spawn(_run(server, proxy, sender));
+        future::pending().await
```
hecrj, line 82: I believe we can simply `await` the `spawn` result here.

hecrj (changes requested):

data/src/stream.rs
tarkah: Won't that tie that future back to the stream?

tarkah: We need it to stay detached

hecrj: No, the handle returned is a future that resolves once the spawned task finishes. It is not the task itself.

tarkah: Yup makes sense, I just read the docs further. Thanks!

PR #561, opened 2024-09-13
Reload config in an async manner

### LINKED ISSUE #551: Load the configuration file in an async fashion

4e554c4c:
Fixes #551
Needed for #552, #424

Test plan:
1) open halloy
2) test reloading config
3) apply patches in #552
4) observe that UI operates while long-running command is blocking the
   config from reloading


src/screen/dashboard/sidebar.rs
```diff
@@ -61,39 +60,176 @@ impl Sidebar {
...
-                self.timestamp = Some(Instant::now());
-                Event::ReloadConfigFile
+                self.reloading_config = true;
+                Some(Event::ReloadConfigFile)
+            },
+            Message::OpenReleaseWebsite => Some(Event::OpenReleaseWebsite),
+            Message::ReloadComplete => {
+                self.reloading_config = false;
+                None
             }
```
tarkah, line 93: Since we have some "complex" interaction here w/ our widget state being tied to the config being loaded, I think it actually makes more sense to call `Task::perform(Config::load(), Message)` at this level, then return an `Event::ConfigReloaded(config)`.

That will allow us to:

- Directly handle the icon state of loaded / error -> reset back to normal icon as we do in `theme_editor` (checkout how `Message::Saved(_)` is handled and how we can chain a `time::sleep` to automatically reset the icon back to normal)
- We won't have to "know" about calling `dashboard.config_reloaded() -> sidebar.config_reloaded()` to handle this state if we're instead reloading from `Hallow / main.rs`

src/screen/dashboard/sidebar.rs
```diff
@@ -61,39 +60,176 @@ impl Sidebar {
...
+fn menu_buttons<'a>(
+    &'a self,
+    main_window: window::Id,
+    panes: &Panes,
+    config: data::config::Sidebar,
+    show_tooltips: bool,
+    file_transfers: &'a file_transfer::Manager,
+    version: &'a Version,
+    theme_editor_open: bool,
+) -> Element<'a, Message> {
```
tarkah, line 106: Nit: Just pass `reloading_config` as a param instead of making this a method of `Sidebar`. This will also prevent this entire code-block from showing up as `+` diff which makes it very hard to review for other changes.

tarkah (changes requested):
@4e554c4c really nice work! This is definitely overdue since we shouldn't be blocking on the UI thread, so thanks for tackling this.

Let me know if you have any questions, I've left a few nits but then some high level design feedback which I think will help w/ dealing w/ state in Sidebar.

src/main.rs
```diff
@@ -33,7 +33,8 @@ use self::theme::Theme;
 use self::widget::Element;
 use self::window::Window;
 
-pub fn main() -> iced::Result {
+#[tokio::main]
+pub async fn main() -> iced::Result {
```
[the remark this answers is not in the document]
4e554c4c: I was on the fence about that.

src/screen/dashboard/sidebar.rs
4e554c4c: I was thinking that this function would be better off as a method that takes &self instead of a reference to a few of its fields

but that can be a different PR (it definitely crowds up this one)

src/main.rs
```diff
@@ -291,41 +326,21 @@ impl Halloy {
...
-                            self.clients.quit(&server, None);
-                        }
-                    }
-                }
+                let event_task = match event {
+                    Some(dashboard::Event::ReloadConfiguration) => Task::future(async  {
+                        Message::UpdateConfig(Config::load().await)
+                    }).chain(Task::done(Message::Dashboard(dashboard::Message::reload_complete()))),
+                    Some(dashboard::Event::ReloadThemes) => Task::future(Config::load())
+                        .and_then(|config| Task::done(Message::ReloadThemes(config))),
```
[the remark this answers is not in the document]
4e554c4c: this isn't completely equivalent since `Config::load()` produces a result. `and_then` only calls the closure if the result is `Ok()`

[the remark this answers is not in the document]
tarkah: Ahh right :+1:

src/screen/dashboard/sidebar.rs
tarkah: Yeah that's definitely fair. Another thing that can be refactored within `menu_buttons` is we can have a button helper closure / function since all the various buttons are basically the same shape with different data.

src/screen/dashboard/sidebar.rs
4e554c4c: let me know what you think about the current iteration. All state manipulation for the sidebar is done in `sidebar.rs` (which is much cleaner, thank you for the suggestion). However, config reloading from commands and shortcuts happens in `dashboard.rs`.

tarkah (approved):
Great work, the changes look great! I've cleaned up a couple minor nits and tested, everything works :+1:

Issue #560, opened 2024-09-13
Clickable channels


4e554c4c:
Usually irc channels are mentioned by their names without any additional formatting (such as hyperlinks)

If a user sees the string `#halloy` in chat, they should be able to click it in order to go to that channel on the current server

Halloy already has `CHANTYPES` support, so this shouldn't be too difficult to add.

tarkah:
Merged in #570

PR #563, opened 2024-09-13
Make user fragment clickable

tarkah:
Allows single left click functionality on user spans similar to the nicklist and nicknames in the buffer.

Adding right click / context menu to these is much more complex since it'll require some overlay approach, so that'll be a TODO item in the future.


casperstorm (approved):
This is a good little QOL. Tested and works.
With this merged I think it would make sense to address #560 as well. Do you want to take it now that you've touched this area a bit now?

tarkah:
> This is a good little QOL. Tested and works.
> With this merged I think it would make sense to address #560 as well. Do you want to take it now that you've touched this area a bit now?

Yup! It'll be very easy to add now.

PR #565, opened 2024-09-14
Optimize context menu, defer building menu

tarkah:
Defers building the menu used for overlay until the menu is opened. This vastly reduces the number of elements we build during `view` and on a channel like `#linux` where the nicklist is huge,  it reduces view times in debug mode from 41ms -> 30ms.


casperstorm (approved):
A clear difference when testing on Windows. Great win!

PR #571, opened 2024-09-16
`msgid` Support

andymandias:
Support for [message-ids](https://ircv3.net/specs/extensions/message-ids).

First step in restructuring PR #370.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Added:
   - Ability to include or exclude channels for server messages (join, part, etc.). See [configuration](https://halloy.squidowl.org/configuration/buffer/server_messages/index.html).
   - Ability to color nicknames within channel messages. See [configuration](https://halloy.squidowl.org/configuration/buffer/channel/message.html#nickname_color)
   - Ability to define a shell command for loading a server password. See [configuration](https://halloy.squidowl.org/configuration/servers/index.html#password_command)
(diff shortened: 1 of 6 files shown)


tarkah (approved):
Simple enough! 

We need to eventually handle outbound messages more gracefully, especially for servers that have [echo-message](https://ircv3.net/specs/extensions/echo-message) so when supported we can:

- Show messages as failed to send if they aren't echo'd within a timeout
- Update our stored outbound message w/ the server chosen message id

andymandias:
Agreed re: outbound messages!  The `chathistory` PR has changes to history that make the updating messages with their id clean & simple, so my inclination would be to wait for the history changes before implementing an id update for stored outbound messages.  Detecting messages that fail to send, however, is not addressed at all by the `chathistory` PR.

4e554c4c:
> We can use `labeled-response` as a way to match echo'd messages back to the original message we sent -> update it w/ server message id. If server doesn't support `labeled-response` but does support `echo-message`, we can use some timestamp / content comparison method I guess.

keeping track of messages with echo-message but not labeled-response is a bit tricky. a good reference IMO is how goguma does it: https://codeberg.org/emersion/goguma/src/commit/d26acea8416739cc962a3b3f05c4c7b148f4bb5d/lib/client.dart#L786
That is, keep track of how many messages are sent vs received. If both are the same then you're generally in sync.

PR #572, opened 2024-09-17
`draft/read-marker` Support

andymandias:
An initial implementation for [`read-marker`](https://ircv3.net/specs/extensions/read-marker) support.  To facilitate the support, a new file is stored with the local read marker information (last read non-internal message).  It's a separate from the message history file since I wanted it to be quick/simple to access without loading the channel's message history.

As a side benefit, this allows unread message state for channels to persist across application close→open.

Second step in cutting up PR #370 into more digestible pieces.

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -6,7 +6,7 @@ use std::collections::{BTreeMap, HashMap, HashSet};
 use std::fmt;
 use std::time::{Duration, Instant};
 
-use crate::history::metadata::read_marker_to_string;
+use crate::history::ReadMarker;
 use crate::message::server_time;
 use crate::time::Posix;
 use crate::user::{Nick, NickRef};
@@ -80,8 +80,7 @@ pub enum Event {
     Broadcast(Broadcast),
     Notification(message::Encoded, Nick, Notification),
     FileTransferRequest(file_transfer::ReceiveRequest),
-    LoadHistoryMetadata(String),
-    UpdateReadMarker(String, Option<DateTime<Utc>>),
+    UpdateReadMarker(String, ReadMarker),
 }
 
 pub struct Client {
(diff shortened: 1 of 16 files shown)


data/src/client.rs
```diff
@@ -367,6 +372,9 @@ impl Client {
                     if contains("multi-prefix") {
                         requested.push("multi-prefix");
                     }
+                    if contains("draft/read-marker") {
+                        requested.push("draft/read-marker");
+                    }
```
tarkah, line 378: Do we need to think about chunking this request? We've quite a few requested capabilities now

data/src/client.rs
```diff
@@ -1225,12 +1242,32 @@ impl Client {
...
+            }
             _ => {}
         }
 
         Some(vec![Event::Single(message, self.nickname().to_owned())])
     }
 
+    pub fn send_markread(&mut self, target: &str, read_marker: Option<DateTime<Utc>>) {
+        if self.supports_read_marker {
+            if let Some(read_marker) = read_marker {
```
tarkah, line 1261: This API should never be called if there's no read marker, since it doesn't do anything with None. Let's change the API to always take a value and handle the option at the call site.

data/src/history.rs
```diff
@@ -1,3 +1,5 @@
+use chrono::{format::SecondsFormat, DateTime, Utc};
+use irc::proto;
 use std::path::PathBuf;
```
tarkah, line 3: Nit: import grouping Std, External, Internal

data/src/history.rs
```diff
@@ -49,69 +50,136 @@ impl From<message::Target> for Kind {
...
+pub async fn load_messages(server: &server::Server, kind: &Kind) -> Result<Vec<Message>, Error> {
+    let path = messages_path(server, kind).await?;
+
+    Ok(read_messages(&path).await.unwrap_or_default())
+}
+
+pub async fn load_read_marker(
+    server: server::Server,
+    kind: Kind,
+) -> Result<Option<DateTime<Utc>>, Error> {
```
tarkah, line 78: I can imagine a scenario where we have additional metadata related to a buffers history we will want to persist 

Maybe we instead have a struct called Metadata and the first field will be this read marker, but it gives us room to expand it in the future without having to add new files.

tarkah (commented):
My review got cut short tonight, but I've left some initial comments! I'll try to wrap it up tomorrow. Looking good so far!

data/src/history.rs
```diff
@@ -253,16 +354,34 @@ impl History {
...
                 ..
-            } => append(&server, &kind, messages).await,
+            } => append(&server, &kind, messages, &metadata).await,
             History::Full {
                 server,
                 kind,
                 messages,
+                metadata,
                 ..
-            } => overwrite(&server, &kind, &messages).await,
```
tarkah, line 263: I need to call metadata.updated here to ensure we save updates metadata when we close

PR #573, opened 2024-09-17
Update user fragments

casperstorm:


data/src/message.rs
```diff
@@ -395,11 +395,10 @@ fn parse_user_fragments(text: &str, channel_users: &[User]) -> Vec<Fragment> {
             let text = chars.collect::<String>();
 
             if is_word {
-                if let Some(user) = channel_users
-                    .iter()
-                    .find(|user| user.nickname().as_ref() == text)
-                {
-                    return Fragment::User(user.clone());
+                if let Some(user) = channel_users.iter().find(|user| {
+                    user.nickname().as_ref().to_ascii_lowercase() == text.to_ascii_lowercase()
```
casperstorm, line 399: After some more testing this isn't quite good enough. We should probably use https://modern.ircdocs.horse/#casemapping-parameter instead.

Also there is still some problems eg if someone is named `eternal_bdkei\!` since `\` is valid: 

> all alphanumerical characters, square and curly brackets ([]{}), backslashes (\), and pipe (|) are valid chars.

tarkah (approved):

PR #583, opened 2024-09-20
Add password_command for sasl and `nick_password`

### LINKED ISSUE #580: Password command not working for SASL
According to the documentation, it should be possible to configure `password_command` for `servers.<servername>.sasl.plain`:
https://halloy.squidowl.org/configuration/servers/sasl/plain.html

When I try to do so, I get an error when starting the client that there is no such configuration option, and also looking at the source code I cannot see any such option in the configuration struct.

4e554c4c:
This commit builds off of #552 and allows the user to supply a command instead of a password file anywhere that a password file is used. If the password command ends unsuccessfully, then an error is displayed.

Trailing whitespace is now trimmed from password commands, as it is commonplace for commands to leave a trailing newline after the password they output.

Resolves #580

### DIFF
--- a/data/src/config.rs
+++ b/data/src/config.rs
@@ -159,6 +159,9 @@ impl Config {
         }
 
         let path = Self::path();
+        if !path.try_exists()? {
+            return Err(Error::ConfigMissing { has_yaml_config: has_yaml_config()? });
+        }
         let content = fs::read_to_string(path)
             .await
             .map_err(|e| Error::LoadConfigFile(e.to_string()))?;
@@ -294,8 +297,8 @@ pub fn random_nickname_with_seed<R: Rng>(rng: &mut R) -> String {
 }
 
 /// Has YAML configuration file.
-pub fn has_yaml_config() -> bool {
-    config_dir().join("config.yaml").exists()
+fn has_yaml_config() -> Result<bool, Error> {
+    Ok(config_dir().join("config.yaml").try_exists()?)
 }
 
 fn default_tooltip() -> bool {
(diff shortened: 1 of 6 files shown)


casperstorm:
I think this deserves a entry in CHANGELOG. I'll do proper review tomorrow.

data/src/config.rs
```diff
@@ -306,16 +306,30 @@ fn default_tooltip() -> bool {
...
 
+impl Error {
+    pub fn is_expected(&self) -> bool {
+        match self {
+            // If a user doesn't have a config when we start up, then we end up with a read error
+            Error::Read(_) => true,
+            _ => false,
+        }
+    }
+}
```
casperstorm, line 331: `Error::Read(_)` isn't the best name anymore i'd say. Lets rename it now that you are touching these files, and while doing so rename this `fn` as well to make it more clear just by reading it.

Perhaps the error could be something like
`Error::Config(_)` or `Error::LoadConfigFile(_)` to keep it similar to `LoadSounds`.
*

casperstorm (changes requested):
A single nit regarding naming, but overall it looks great.

casperstorm (approved):
Looks good to me. @tarkah do you have anything for this?

data/src/server.rs
```diff
@@ -77,62 +103,62 @@ impl Map {
         for (_, config) in self.0.iter_mut() {
             if let Some(pass_file) = &config.password_file {
                 if config.password.is_some() || config.password_command.is_some() {
-                    return Err(Error::Parse(
-                        "Only one of password, password_file and password_command can be set."
-                            .to_string(),
-                    ));
+                    return Err(Error::Parse(DUP_PASS_MSG.to_string()));
```
tarkah, line 106: Nit: Just make these separate Error variants so we don't need to use `Parse` plus some const string for each error type.

src/main.rs
```diff
@@ -151,29 +151,28 @@ impl Halloy {
...
             }
-            Err(error) => match &error {
-                config::Error::Parse(_) | config::Error::LoadSounds(_) => (
-                    Screen::Help(screen::Help::new(error)),
-                    Config::default(),
-                    Task::none(),
-                ),
-                _ => {
+            Err(error) => {
+                if config::has_yaml_config() {
```
tarkah, line 155: We really should add a new `config::Error` variant for `config::Error::ConfigMissing { has_yaml_config: bool }` where we check if `path.exists` and if not, we can populate the error with `config::has_yaml_config`.

Then under this branch, we can simply check if error is `Error::ConfigMissing { has_yaml_config: true }` for `Screen::Migration`.

The `ConfigMissing { has_yaml_config: false }` can then route to `Screen::Welcome`. We only want to route here if the config is missing & no yaml config exits. Any parsing errors should truly route to `Screen::Help` w/ the error

tarkah (changes requested):

tarkah (approved):
Changes look great! I think that cleaned it up very nicely. Thanks for tackling this

casperstorm:
Thanks for yet another good PR.

PR #595, opened 2024-09-25
re-add zenburn

Tea23:
Zenburn went missing with the new theme schema. I recreated it by eye (actually before the theme editor got added in - so some things might need tweaks). Here it is, back again!


casperstorm (approved):

casperstorm:
Looks good thanks!

PR #597, opened 2024-09-25
Only highlight if user nick isn't part of another word

### LINKED ISSUE #590: Feature request: Avoid highlighting lines where the user's nickname is part of another word
If the user's nickname is "thing", and someone uses the word "something" in a channel, that line will be highlighted. It would be nice for Halloy to highlight a line only when an entire word matches the user's nickname.

casperstorm:
Fixes #590.

This fixes so that we only highlight if nickname isn't part of another word.

### DIFF
--- a/data/src/message.rs
+++ b/data/src/message.rs
@@ -1061,11 +1061,16 @@ fn monitored_targets_text(targets: Vec<String>) -> Option<String> {
 }
 
 pub fn reference_user(sender: NickRef, own_nick: NickRef, message: &Message) -> bool {
+    let contains_nick = |text: &str, nick: &str| {
+        text.split(|c: char| !c.is_alphanumeric())
+            .any(|word| word == nick)
+    };
+
     let has_nick = match &message.content {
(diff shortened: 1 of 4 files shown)


data/src/message.rs
```diff
@@ -1061,11 +1061,16 @@ fn monitored_targets_text(targets: Vec<String>) -> Option<String> {
 }
 
 pub fn reference_user(sender: NickRef, own_nick: NickRef, message: &Message) -> bool {
+    let contains_nick = |text: &str, nick: &str| {
+        text.split(|c: char| !c.is_alphanumeric())
+            .any(|word| word == nick)
+    };
```
tarkah, line 1067: We should use the same logic as when we parse fragments, which is non whitespace and then compare word and trimmed word, both normal and lowercase with incoming nick

casperstorm:
Take a look if you like the change @tarkah.

tarkah (approved):
LGTM! Minor nit, but all the function names could probably be `s/reference/references/` since they're "conditional" checks, "if text reference**s** nickname"

PR #598, opened 2024-09-26
Fix memory leak in rust-dark-light crate

casperstorm:
The original crate (https://github.com/frewsxcv/rust-dark-light) has a memory leak which causes macOS to eventually crash. This has been fixed by @madsmtm in a [PR ](https://github.com/frewsxcv/rust-dark-light/pull/37) where he updates to `objc2` - however this hasn't been merged yet and have been sitting since may.

I've forked the fork and will keep an eye on when it merges.
Thanks for the fix, @madsmtm and thanks for spotting it @englut.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1165,14 +1165,16 @@ dependencies = [
 [[package]]
 name = "dark-light"
 version = "1.1.1"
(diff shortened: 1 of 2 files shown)


englut (approved):
confirmed, this fixes the crashes I was experiencing on macOS

tarkah (approved):

PR #599, opened 2024-09-27
Ability to disable who polling

casperstorm:
This adds an configuration option to disable WHO polling all together.

### DIFF
--- a/book/src/configuration/servers/README.md
+++ b/book/src/configuration/servers/README.md
@@ -220,6 +220,14 @@ Example. `["/msg NickServ IDENTIFY foo bar"]`.
 - **type**: array of string
 - **values**: array of any strings
 - **default**: not set
+  
+## `who_poll_enabled`
+
+Whether or not to WHO polling is enabled.
+
+- **type**: boolean
+- **values**: `true`, `false`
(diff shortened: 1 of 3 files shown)


book/src/configuration/servers/README.md
```diff
@@ -220,6 +220,14 @@ Example. `["/msg NickServ IDENTIFY foo bar"]`.
 - **type**: array of string
 - **values**: array of any strings
 - **default**: not set
+  
+## `who_poll_enabled`
+
+Whether or not to WHO polling is enabled.
```
andymandias, line 226: I know it's a bit redundant, but users may tunnel vision on the configuration option they're looking for, so it might be good to tack on the "for servers without away-notify" here as well.

andymandias (approved):
I think if WHO polling is turned off via a configuration reload, then users will be stuck in whatever away state they have at the time.  But even if I'm right about that, it's a harmless edge case that doesn't need to be handled in this PR imo.

book/src/configuration/servers/README.md
casperstorm: Correct but I think that is an edge case I'm willing to accept.

PR #601, opened 2024-09-30
Feat/logging buffer

tarkah:


casperstorm (approved):
Updated with icon (maybe not 100% yet on it, but good enough for now i think).
Also added command bar entry.

PR #610, opened 2024-10-06
Remove iced `auto-detect-theme` feature

bbb651:
A different version of `dark-light` is depended on directly, meaning a bunch of unused dependencies were compiled and bundled. 
This is extra bad on linux because of `zbus`, see: https://github.com/iced-rs/iced/pull/2624#issuecomment-2389512677 (to entirely remove `zbus 3.x`, it's blocked on https://github.com/frewsxcv/rust-dark-light/pull/41 + upgrading `dark-light`, it now already includes the `objc2` patch)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -41,17 +41,6 @@ checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
 
 [[package]]
 name = "ahash"
-version = "0.7.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


casperstorm:
@tarkah could you take a look at this. Seems reasonable to me.

casperstorm (approved):

bbb651:
They merged the `objc2` to main (and `zbus` was upgraded in a different PR), should upgrading to it (still not on crates.io but better than depending on a fork) be a separate PR or part of this one? (it also completely dropped reading toolkit specific configuration files in favor of portals, I'm a bit worried it'll make people on window manager setups with reconfigured portals experience the 30 seconds portal timeout, hopefully the situation on e.g. hyprland has improved since I last tried it).

tarkah (approved):
Makes sense to me as well

PR #619, opened 2024-10-18
`chathistory` Support

andymandias:
A refactoring of the previous [`chathistory`](https://ircv3.net/specs/extensions/chathistory) PR.  It has all of the same features as before (plus a new configuration option), but the plumbing has been restructured to better match the paradigm set by the `draft/read-marker` PR.

Features (when `chathistory` is supported by the server):
- Uses `chathistory` to request new history for all joined channels from the server (limited to one request if no history for the channel that exists, 500 messages or the maximum number of messages allowed by the server).
- Adds button at the start of  channel/query buffers to request older history via `chathistory`.
- Uses `chathistory` to determine whether queries were received since last connection to the server, then request the new history for those queries from the server.
- Configuration option to automatically request older history when scrolling to the top of a channel/query buffer (`infinite_scroll`).
- Configuration option to disable `chathistory` for a server.

I haven't had much time to test it since refactoring, so I may make fixes in this PR as testing continues.  But so far it is working as intended for me, has the entirety of my intended functionality for initial support, and is ready for review.

### DIFF
--- a/data/src/history.rs
+++ b/data/src/history.rs
@@ -9,11 +9,12 @@ use irc::proto;
 use tokio::fs;
 use tokio::time::Instant;
 
+use crate::message::{self, MessageReferences};
 use crate::user::Nick;
-use crate::{buffer, compression, environment, message, Buffer, Message, Server};
+use crate::{buffer, compression, environment, Buffer, Message, Server};
 
 pub use self::manager::{Manager, Resource};
-pub use self::metadata::{MessageReferences, Metadata, ReadMarker};
+pub use self::metadata::{Metadata, ReadMarker};
 
 pub mod manager;
 pub mod metadata;
--- a/data/src/history.rs
+++ b/data/src/history.rs
@@ -479,14 +479,16 @@ impl History {
 /// of the incoming message. Either message IDs match, or server times
 /// have an exact match + target & content.
 pub fn insert_message(messages: &mut Vec<Message>, message: Message) {
+    const FUZZ_SECONDS: chrono::Duration = chrono::Duration::seconds(1);
+
     if messages.is_empty() {
         messages.push(message);
 
         return;
     }
 
-    let start = message::fuzz_start_server_time(message.server_time);
-    let end = message::fuzz_end_server_time(message.server_time);
+    let start = message.server_time - FUZZ_SECONDS;
+    let end = message.server_time + FUZZ_SECONDS;
 
     let start_index = match messages.binary_search_by(|stored| stored.server_time.cmp(&start)) {
         Ok(match_index) => match_index,
--- a/data/src/history.rs
+++ b/data/src/history.rs
@@ -13,7 +13,7 @@ use crate::user::Nick;
 use crate::{buffer, compression, environment, message, Buffer, Message, Server};
 
 pub use self::manager::{Manager, Resource};
-pub use self::metadata::{Metadata, ReadMarker};
+pub use self::metadata::{MessageReferences, Metadata, ReadMarker};
 
 pub mod manager;
 pub mod metadata;
@@ -157,7 +157,9 @@ pub async fn append(
     let loaded = load(kind.clone()).await?;
 
     let mut all_messages = loaded.messages;
-    all_messages.extend(messages);
+    messages.into_iter().for_each(|message| {
+        insert_message(&mut all_messages, message);
+    });
 
     overwrite(kind, &all_messages, read_marker).await
 }
@@ -203,6 +205,7 @@ pub enum History {
         last_updated_at: Option<Instant>,
         max_triggers_unread: Option<DateTime<Utc>>,
         read_marker: Option<ReadMarker>,
+        chathistory_references: Option<MessageReferences>,
     },
     Full {
         kind: Kind,
@@ -220,6 +223,7 @@ impl History {
             last_updated_at: None,
             max_triggers_unread: None,
             read_marker: None,
+            chathistory_references: None,
         }
     }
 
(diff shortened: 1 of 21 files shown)


data/src/history.rs
```diff
@@ -397,6 +449,90 @@ impl History {
     }
 }
 
+/// Insert the incoming message into the provided vector, sorted
+/// on server time
+///
+/// Deduplication is only checked +/- 1 second around the server time
+/// of the incoming message. Either message IDs match, or server times
+/// have an exact match + target & content.
+pub fn insert_message(messages: &mut Vec<Message>, message: Message) {
```
tarkah, line 458: Didn't we have a commit which used a binary search implementation for inserting messages?

andymandias: Yes, this is the `insert_message` from that work we did together, completely unchanged.  I was lazy about persisting the commit history while refactoring, so it's all been squashed into one commit.

The one thing this doesn't keep related to that work is the sorting of saved histories on load.  In that version of the `chathistory` PR, I had changed the history storage scheme to no longer be obfuscated (no longer use hashes in naming the files).  If histories were not found with the non-hashed storage scheme (where histories can be expected to be sorted already), then it would attempt to load & sort a history from the old, hashed storage scheme.  I still like that idea, but it seemed like this PR might have too many changes already, so I omitted it this time around.

data/src/history.rs
```diff
@@ -397,6 +449,90 @@ impl History {
...
+    if messages.is_empty() {
+        messages.push(message);
+
+        return;
+    }
+
+    let start = message::fuzz_start_server_time(message.server_time);
+    let end = message::fuzz_end_server_time(message.server_time);
+
```
tarkah, line 467: I don't think these helpers are helpful or necessary. We can add / sub `chrono::Duration` from `DateTime<Utc>` and construct is `const`. So we can add `const FUZZ_SECONDS: chrono::Duration = chrono::Duration::seconds(1)` to the first line of this function and then `let start = message.server_time - FUZZ_SECONDS;`.

andymandias: Aha!  Yes, I think I added the functions primarily because I was only aware of the `TimeDelta` version of changing the time which is considerably more of a mouthful.  Secondarily because `FUZZ_SECONDS` was being shared with the `chathistory` request.  I don't remember why I did that, and while they're not completely unrelated I don't see any reason to keep them in lock-step.  I'll add `FUZZ_SECONDS` here as you suggest, and set a distinct `FUZZ_SECONDS` inside `isupport.rs`.

src/buffer/scroll_view.rs
```diff
@@ -22,20 +25,22 @@ pub enum Message {
...
+    RequestOlderChatHistory,
 }
 
 #[derive(Debug, Clone, Copy)]
 pub enum Kind<'a> {
     Server(&'a Server),
-    Channel(&'a Server, &'a str),
-    Query(&'a Server, &'a Nick),
+    Channel(&'a Server, &'a str, Option<ChatHistoryState>),
+    Query(&'a Server, &'a Nick, Option<ChatHistoryState>),
```
tarkah, line 43: Stuffing this into here seems to be causing us to require fetching this state when it's not even used. Checkout how we're querying for it when we construct `Kind` when calling `scroll_to_backlog`, this state isn't used anywhere there.

It seems we only need it during `view` so instead pass `Option<ChatHistoryState>` as a separate param there and it's just always `None` from server & query. This will clean up some code paths

tarkah (commented):
This is looking really good! A lot easier to grok without all the other work that you've now separately landed.

I think there's some opportunity to remove a lot of the chat history logic from the frontend and push it behind the `history` API on the data layer which I think would be a little nicer. Let me know what you think!

data/src/history/metadata.rs
```diff
@@ -57,6 +58,54 @@ impl fmt::Display for ReadMarker {
...
+                    if let Some(id) = &self.id {
+                        return isupport::MessageReference::MessageId(id.clone());
+                    }
+                }
+                isupport::MessageReferenceType::Timestamp => {
+                    return isupport::MessageReference::Timestamp(self.timestamp);
+                }
+            }
+        }
+
```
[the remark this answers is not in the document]
andymandias: The [chathistory spec](https://ircv3.net/specs/extensions/chathistory#isupport-tokens) says that they are to be provided by the server ordered in (the server's) decreasing preference.  I think the most polite thing to do is respect the server's preference, but I don't believe we are required to do so.  My personal opinion is that ID is preferable over timestamp, when available.

data/src/history/metadata.rs
```diff
@@ -57,6 +58,54 @@ impl fmt::Display for ReadMarker {
...
+        isupport::MessageReference::None
+    }
+}
+
+impl PartialEq for MessageReferences {
+    fn eq(&self, other: &Self) -> bool {
+        self.timestamp.eq(&other.timestamp)
```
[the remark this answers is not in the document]
andymandias: Currently this is only used for some syntactic sugar, in `update_partial` in `data/src/history.rs` this (via `Ord`) allows `.max` to be used.  `id` isn't considered because for the purpose it's currently being used we only want to know if we have a more recent set of references.

I originally implemented it this way because `Ord` was utilized elsewhere in the refactor, but that turned out to be unnecessary and was stripped back to just the one use rather than unraveled completely.  No particular attachment to it.

src/buffer/scroll_view.rs
```diff
@@ -245,6 +283,11 @@ impl State {
                         scrollable::scroll_to(self.scrollable.clone(), new_offset),
                         None,
                     );
+                } else if infinite_scroll
+                    && matches!(self.limit, Limit::Top(_))
+                    && relative_offset == 0.0
```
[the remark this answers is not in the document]
andymandias: Aha, thanks!  I can't remember why, but I thought it was necessary to check limit as well.  I'll switch this.

andymandias:
I think this is in a better place (that ~addresses the current review comments) and is ready for another look, when you have time @tarkah.

tarkah (commented):
This looks like it's in a really good spot @andymandias! I will try to test tomorrow but I think the code is looking good. 

@casperstorm did you have a chance to test at all?

casperstorm (approved):
I've tested, and it works great!

tarkah (approved):
Awesome! There's a couple minor bugs w/ inifinte scroll but nothing that should hold this up. @andymandias can you add CHANGELOG and we can then merge :rocket:

PR #621, opened 2024-10-18
Add `IRC` to linux desktop file keywords

bbb651:
This makes it show up when searching "irc" on GNOME.


andymandias (approved):
LGTM!  Confirmed works on GNOME, conforms to [the Freedesktop spec](https://specifications.freedesktop.org/desktop-entry-spec/latest/recognized-keys.html).

casperstorm (approved):

PR #623, opened 2024-10-18
Change sorting of channels in sidebar

4e554c4c:
This commit changes the sorting of channels to more resemble the weechat [autosort](https://weechat.org/scripts/source/autosort.py.html/) plugin, which I use. In the future more sort methods could be added.

This depends on determining the character that is used to denote a channel. Instead of using our existing logic, I implemented `CHANTYPES` following the [modern irc docs](https://modern.ircdocs.horse/#chantypes-parameter). For good measure, I also added support for the `STATUSMSG` isupport.

As discussed on the #ircdocs channel, there is a parsing ambiguity if the `STATUSMSG` and `CHANTYPES` isupports intersect (which they rarely do in practice). I do not define our behaviour in this case, but chose a fairly simple heuristic where the first `CHANTYPES` char always begins a halloy channel. 

I've run `rustfmt` on one file, and also fixed some odd assumptions that there is only one status prefix on a message. Currently a channel can have multiple status prefixes and I've changed the code to reflect this.

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -188,6 +188,17 @@ impl Client {
         }
     }
 
+    fn start_reroute(&self, command: &Command) -> bool {
+        use Command::*;
+
+        if let MODE(target, _, _) = command {
+            !self.is_channel(target)
+        } else {
+            matches!(command, WHO(..) | WHOIS(..) | WHOWAS(..))
+        }
+    }
+
+
     fn send(&mut self, buffer: &buffer::Upstream, mut message: message::Encoded) {
         if self.supports_labels {
             use proto::Tag;
@@ -204,7 +215,7 @@ impl Client {
             }];
         }
 
-        self.reroute_responses_to = start_reroute(&message.command).then(|| buffer.clone());
+        self.reroute_responses_to = self.start_reroute(&message.command).then(|| buffer.clone());
 
         if let Err(e) = self.handle.try_send(message.into()) {
             log::warn!("Error sending message: {e}");
@@ -896,7 +907,7 @@ impl Client {
             Command::Numeric(RPL_WHOREPLY, args) => {
                 let target = args.get(1)?;
 
-                if proto::is_channel(target) {
+                if self.is_channel(target) {
                     if let Some(channel) = self.chanmap.get_mut(target) {
                         channel.update_user_away(args.get(5)?, args.get(6)?);
 
@@ -915,7 +926,7 @@ impl Client {
             Command::Numeric(RPL_WHOSPCRPL, args) => {
                 let target = args.get(2)?;
 
-                if proto::is_channel(target) {
+                if self.is_channel(target) {
                     if let Some(channel) = self.chanmap.get_mut(target) {
                         channel.update_user_away(args.get(3)?, args.get(4)?);
 
@@ -947,7 +958,7 @@ impl Client {
             Command::Numeric(RPL_ENDOFWHO, args) => {
                 let target = args.get(1)?;
 
-                if proto::is_channel(target) {
+                if self.is_channel(target) {
                     if let Some(channel) = self.chanmap.get_mut(target) {
                         if matches!(channel.last_who, Some(WhoStatus::Receiving(_))) {
                             channel.last_who = Some(WhoStatus::Done(Instant::now()));
@@ -995,7 +1006,7 @@ impl Client {
                 }
             }
             Command::MODE(target, Some(modes), Some(args)) => {
-                if proto::is_channel(target) {
+                if self.is_channel(target) {
                     let modes = mode::parse::<mode::Channel>(modes, args);
 
                     if let Some(channel) = self.chanmap.get_mut(target) {
@@ -1053,7 +1064,7 @@ impl Client {
             Command::Numeric(RPL_ENDOFNAMES, args) => {
                 let target = args.get(1)?;
 
-                if proto::is_channel(target) {
+                if self.is_channel(target) {
                     if let Some(channel) = self.chanmap.get_mut(target) {
                         if !channel.names_init {
                             channel.names_init = true;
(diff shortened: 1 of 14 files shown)


4e554c4c:
hm, this PR seems to have performance issues. The sort code is being called a lot of the time

4e554c4c:
@tarkah I was asked to tag you for a review :smile_cat:

tarkah:
Hey @4e554c4c I just got back from vacation, I'll do some review this week. Thanks!

data/src/client.rs
```diff
@@ -1270,8 +1282,27 @@ impl Client {
...
+        };
+
+        if [a_chantype, b_chantype].iter().all(|c| self.chantypes().contains(c)) {
+            let ord = a.trim_start_matches(a_chantype).cmp(b.trim_start_matches(b_chantype));
+            if ord != Ordering::Equal {
+                return ord;
+            }
+        }
+        a.cmp(b)
+    }
```
tarkah, line 1302: I think we're long overdue for a `Channel` type in our data crate. We're exposing a ton of the complexities of parsing / dealing w/ channel prefix vs names across the codebase instead of encapsulating it within a `channel` module / type.

I'm not sure we should block this work on it as it'd be a much bigger refactor, but I think it'd be really nice to clean this up along w/ dealing w/ channels and "targets" (channel vs query "nick") throughout the codebase instead of just `String` like we've been doing.

data/src/isupport.rs
```diff
@@ -139,12 +140,11 @@ impl FromStr for Operation {
...
-                            if value.is_empty() {
+                            let chars = value.chars().collect::<Vec<_>>();
+                            if chars.is_empty() {
                                 Ok(Operation::Add(Parameter::CHANTYPES(None)))
-                            } else if value.chars().all(|c| proto::CHANNEL_PREFIXES.contains(&c)) {
-                                Ok(Operation::Add(Parameter::CHANTYPES(Some(
-                                    value.to_string(),
-                                ))))
+                            } else if chars.iter().all(|c| proto::CHANNEL_PREFIXES.contains(c)) {
+                                Ok(Operation::Add(Parameter::CHANTYPES(Some(chars))))
```
tarkah, line 147: Why do we need to validate what the server specifies as a channel prefix? In theory a server can choose something completely random and we should respect it? I think we can remove `proto::CHANNEL_PREFIXES` completely since we should just trust the server and if they don't specify ISUPPORT, we use the `DEFAULT_CHANNEL_PREFIXES`.

Same with `CHANNEL_MEMBERSHIP_PREFIXES` below.

data/src/history.rs
```diff
@@ -75,7 +75,8 @@ impl From<String> for Kind {
 
 impl From<&str> for Kind {
     fn from(target: &str) -> Self {
-        if proto::is_channel(target) {
+        // XXX not sure how to fix this in a way that will preserve on-disk history
+        if proto::is_channel(target, proto::DEFAULT_CHANNEL_PREFIXES) {
```
tarkah, line 79: We only construct via this when receiving read marker updates from the server, which will always be a valid channel or query.

I we remove this completely once we get around to refactoring as I mention in https://github.com/squidowl/halloy/pull/623/files#r1809548087 and we add some new `Channel` and `Target` types to use across the codebase

tarkah:
@andymandias would love your eyes on this as well since you've got great insight into ISUPPORT and these nuances now

data/src/client.rs
4e554c4c: one issue is that Ord (as in hashmap) and Ord as in channel sorting are (and should be!) different

we really want a locale-dependent sort on channels and i don't think this is suitable for `Ord`, which the BTreeMap is based on (for an example of this see my [gamja PR](https://codeberg.org/emersion/gamja/pulls/4))

I wasn't entirely sure on how to rectify this so i decided to not work on it in this PR... so maybe we should push it to another PR?

data/src/isupport.rs
andymandias: If we want to open the floodgates for any possible channel/ `STATUSMSG`, then it would probably be good for us to handle the `PREFIX` ISUPPORT parameter.  That gives the mapping between prefix characters and user access levels, and it's my reading that `STATUSMSG` prefixes should all come from the default/provided set of `PREFIX` characters.  Aside from `STATUSMSG`, I believe user access level parsing should be derived from the mapping provided by the `PREFIX` ISUPPORT parameter.

andymandias (commented):
It's definitely helpful to have someone look at this who actually uses these features.  I don't really use many of these more esoteric features, so things get missed or put off that wouldn't make sense to a frequent user.  These look like good additions/extensions to the ISUPPORT support to me.

data/src/client.rs
4e554c4c: I opened #626 to track this

irc/proto/src/lib.rs
```diff
@@ -49,33 +49,48 @@ pub fn command(command: &str, parameters: Vec<String>) -> Message {
...
+    // We parse the target by finding the first character in chantypes, and returing (even if that
+    // character is in statusmsg_prefixes)
+    // If the characters before the first chantypes character are all valid prefixes, then we have
+    // a valid channel name with those prefixes.    let chan_index = target.find(chantypes)?;
+    let chan_index = target.find(chantypes)?;
 
-    if is_channel(target) {
-        Some((None, target.to_string()))
+    // will not panic, since `find` always returns a valid codepoint index
+    let (prefix, chan) = target.split_at(chan_index);
```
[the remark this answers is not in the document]
4e554c4c: I changed this and my tests started failing... so now I remember why I did this :)

We want an inclusive split for this function, since the channel string includes its chantype. split_at is an inclusive split, and we don't want to split multiple times like `split_inclusive`.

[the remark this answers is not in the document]
andymandias: Ahh, right. There is no `split_once_inclusive`. Let's stick with the `split_at` then.

4e554c4c:
Fixed the merge conflicts! @tarkah, do you mind checking this out before it accumulates too many more?

tarkah (approved):
Sorry for taking so long to approve, thanks for this!

PR #632, opened 2024-10-26
Handle errors in `data::Client::handle`

### LINKED ISSUE #627: Handle errors in `data::Client::handle`
There are many errors that can occur in `data::Client::handle`, however they are almost all silenced with `let _ = ...`. In my opinion, these should be properly handled and displayed when necessary.

4e554c4c:
For simplicity's sake, the returned type is a `Result<T, anyhow::Error>` . Errors are displayed in a modal text box on top of the IRC client.

Test plan:
Added code which returns an error from `data::Clients::receive` with a 1% chance, and observed that the error shows up.
result: https://nyan.network/notice/AnPGIlp1et3YgBAkfw

Resolves #627

### DIFF
--- a/data/src/stream.rs
+++ b/data/src/stream.rs
@@ -280,12 +280,17 @@ async fn connect(
 
     let (sender, receiver) = mpsc::channel(100);
 
+    let mut client = Client::new(server, config, sender);
+    if let Err(e) = client.connect() {
+        log::error!("Error when connecting client: {:?}", e);
+    }
+
     Ok((
         Stream {
             connection,
             receiver,
         },
-        Client::new(server, config, sender),
+        client,
     ))
 }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -118,6 +118,10 @@ fn settings(config_load: &Result<Config, config::Error>) -> iced::Settings {
     }
 }
 
+fn handle_irc_error(e: anyhow::Error) {
+    log::error!("IRC error: {:?}", e);
+}
+
 struct Halloy {
     version: Version,
     screen: Screen,
@@ -345,6 +349,10 @@ impl Halloy {
                         self.clients.quit(&server, None);
                         Task::none()
                     }
+                    Some(dashboard::Event::IrcError(e)) => {
+                        handle_irc_error(e);
+                        Task::none()
+                    }
                     Some(dashboard::Event::Exit) => {
                         let pending_exit = self.clients.exit();
 
@@ -489,9 +497,17 @@ impl Halloy {
                     let commands = messages
                         .into_iter()
                         .flat_map(|message| {
+                            let events = match self.clients.receive(&server, message) {
+                                Ok(events) => events,
+                                Err(e) => {
+                                    handle_irc_error(e);
+                                    return vec![];
+                                }
+                            };
+
                             let mut commands = vec![];
 
-                            for event in self.clients.receive(&server, message) {
+                            for event in events {
                                 // Resolve a user using client state which stores attributes
                                 let resolve_user_attributes = |user: &User, channel: &str| {
                                     self.clients
@@ -816,9 +832,10 @@ impl Halloy {
                 Task::none()
             }
             Message::Tick(now) => {
-                self.clients.tick(now);
-
-                if let Screen::Dashboard(dashboard) = &mut self.screen {
+                if let Err(e) = self.clients.tick(now) {
+                    handle_irc_error(e);
+                    Task::none()
+                } else if let Screen::Dashboard(dashboard) = &mut self.screen {
                     dashboard.tick(now).map(Message::Dashboard)
                 } else {
                     Task::none()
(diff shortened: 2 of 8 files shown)


data/src/stream.rs
```diff
@@ -280,12 +280,17 @@ async fn connect(
 
     let (sender, receiver) = mpsc::channel(100);
 
+    let mut client = Client::new(server, config, sender);
+    if let Err(e) = client.connect() {
+        log::error!("Error when connecting client: {:?}", e);
+    }
```
tarkah, line 286: You can use "inspect" to log the error without needing to assign and check an intermediate variable

src/modal.rs
```diff
@@ -2,11 +2,12 @@ use crate::widget::Element;
 use data::{config, Server};
 
 pub mod connect_to_server;
-pub mod reload_configuration_error;
+pub mod error;
 
 #[derive(Debug)]
 pub enum Modal {
     ReloadConfigurationError(config::Error),
+    IrcError(anyhow::Error),
```
tarkah, line 10: I'm not sure I'm sold on putting errors in a modal. It seems very disruptive. We have a log buffer now, we should just make sure everything is logged and they can be viewed there. 

Are there certain types of errors that need to escalate to a modal? We can probably think of a different UI mechanism that indicates errors have occurred in the sidebar and clicking it opens the log buffer

4e554c4c: i was thinking that an error that opens up in a corner window would be the best, not one that takes up full screen like the current modal

casperstorm: I think we should use the logger buffer for this.

tarkah: So like a toast notification? I'll defer to @casperstorm on that. Anything that bleeds over the panes will be quite disruptive / noisy imo. 

I'm thinking we show an error icon in the sidebar, clicking it clears the icon and opens log buffer. New errors make icon reappear if log buffer is closed. And icon visibility is configurable.

4e554c4c: I was thinking a toast in the sidebar that you can click away

but a link that goes to the log buffer would also be good IMO.

For now I'll log an error and we can work on the UI element later

tarkah: Sounds good 👍

casperstorm:
@4e554c4c You still want to update this PR to move the irc errors to log instead of a modal?

4e554c4c:
yes, sorry I've been on vacation. I'm happy if somebody else wants to take over the PR though. Otherwise I can probably get to it this weekend

data/src/stream.rs
4e554c4c: I prefer using this construction, since `.inspect_err` returns the original result that needs to be used.

So at best this would be 
```rust
let _ = client.connect().inspect_err(|e| log::error!("Error when connecting client: {:?}", e)));
```
 which I don't think is cleaner

4e554c4c:
ok ready for review :smile:

casperstorm:
Gonna review this today after work 👏🏻

casperstorm (approved):
This looks good to me, awaiting final confirmation from @tarkah.

src/main.rs
```diff
@@ -489,9 +491,14 @@ impl Halloy {
                     let commands = messages
                         .into_iter()
                         .flat_map(|message| {
+                            let events = match self.clients.receive(&server, message) {
+                                Ok(events) => events,
+                                Err(e) => return vec![Task::done(Message::IrcError(e))],
```
tarkah, line 496: We shouldn't be returning a `Task` / `Message` here for simply logging the error, we can handle this in the current code path. Let's just instead have a function called `log_irc_error` or something and call that directly

src/main.rs
```diff
@@ -345,6 +346,7 @@ impl Halloy {
                         self.clients.quit(&server, None);
                         Task::none()
                     }
+                    Some(dashboard::Event::IrcError(e)) => Task::done(Message::IrcError(e)),
```
tarkah, line 349: Same here

src/main.rs
```diff
@@ -816,9 +823,9 @@ impl Halloy {
                 Task::none()
             }
             Message::Tick(now) => {
-                self.clients.tick(now);
-
-                if let Screen::Dashboard(dashboard) = &mut self.screen {
+                if let Err(e) = self.clients.tick(now) {
+                    Task::done(Message::IrcError(e))
```
tarkah, line 827: And here. We can remove `Message::IrcError` completely

tarkah (changes requested):
Looks great! One small fix needed around not needing to abuse `Task -> Message` for something we can just call directly. We should only be returning `Task` if we truly have actual tasks that need executing, not for abusing it to generate a message when we can just call a normal rust function instead.

src/main.rs
4e554c4c: I had a task initially for the modal, but IMO we should keep it just in case the behaviour changes in the future (like with a red dot getting shown)

tarkah: Let's not keep dead code around in anticipation of something that may never land. This current pattern is not one we want to use.

4e554c4c: sounds good :)

tarkah:
Awesome thank you, this is great!

PR #643, opened 2024-11-09
add cyberpunk theme

glaubermagal:
felt like it would be nice to have a darker/futuristic theme


casperstorm (approved):
That It looks great, thanks!

PR #646, opened 2024-11-10
Add users to userlist on PRIVMSG event

voidlily:
A certain irc server doesn't respond to WHO properly, and we might have missed the join. This means that we might never actually know that that user is in the channel even if they are sending messages.

The best fix I could come up with client-side for this is to just add the user to the channel's userlist on a PRIVMSG event.

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -675,6 +675,13 @@ impl Client {
                             }
                         }
 
+                        // add user to user list if we somehow missed the join
+                        if let Some(channel) = self.chanmap.get_mut(channel) {
+                            if channel.users.take(&user).is_none() {
+                                channel.users.insert(user.clone());
+                            }
+                        }
+
                         // Highlight notification
                         if message::references_user_text(user.nickname(), self.nickname(), text) {
                             return Ok(vec![Event::Notification(


4e554c4c:
Unless if a channel is [`+n`](https://modern.ircdocs.horse/#no-external-messages-mode) it is possible to send a privmsg to a channel that is not joined, so i don't think adding users directly to the channel list after privmsg is a great idea.

voidlily:
you know what, i honestly forgot that +n is a thing that you can turn off and that's a very good point to keep in mind.

4e554c4c:
i think this is a good thing to keep on the backlog (i know some ircds like twitch absolutely don't support WHO) but it would need some careful thinking

voidlily:
yeah the "some server" i was alluding to was twitch, wasn't sure how saying upfront that "yeah it's twitch" would be taken here, or if it would be "but it's not real irc we don't support that"

twitch does have a workaround with `/CAP REQ :twitch.tv/membership` but it's been weird in the exact formatting to get the server to accept it on this client (as compared to hexchat) - i've had the most success with `/raw CAP REQ :twitch.tv/membership` but it may have been without the colon, it's actually kind of hard to tell because i can't find like a raw response log that shows ping/pong and stuff like that anywhere to debug when i _do_ get it right (sometimes i've even had issue with it not taking the command if i put it in the `on_connect` command list)

voidlily:
i can reopen this since can't easily like "turn it back into an issue" on github so it doesn't disappear into the aether, since it seems like something that may be worth thinking about in any case

4e554c4c:
imo halloy should also have configurable cap requests like weechat

casperstorm:
> imo halloy should also have configurable cap requests like weechat

Yeah i agree, and should be a easy one to implement.

casperstorm:
@4e554c4c @voidlily: circling back here - is this still needed?

4e554c4c:
i think it's ok to track at #662, this can be closed

PR #651, opened 2024-11-11
Minor Fixes

andymandias:
A few minor things I found while working on `chathistory`.  A typo, a missed return value, and a tweak to the return value for `prefixes`.  If brackets are configured for status message prefixes, then brackets were being placed around the empty list of prefixes (e.g. with `(` and `)` configured as status message prefix brackets, `()` was being prepended to virtually every message in a channel).  I'm open to changing the prefix bracketing logic instead of the return value, but changing the return value for `prefixes` seemed cleaner to me.


casperstorm (approved):
Looks good!

PR #654, opened 2024-11-12
Ability to change away appearance for users 

casperstorm:
This PR introduces a `Away` struct with a single `appearance` enum inside to handle the the look of away nicknames. 
Down the road it could make sense to expand the `Away` struct with auto-nickname change after `x` minutes, etc.

```
[buffer.away]
appearance = "dimmed" | "solid"
```

Does naming and location make sense? Or should it be placed under `[buffer.channel.away]`?
I feel like it is a global buffer thing since buffer::query also has away state but let me know.


data/src/config/away.rs
```diff
@@ -0,0 +1,21 @@
+use serde::Deserialize;
+
+#[derive(Debug, Clone, Copy, Default, Deserialize)]
+pub struct Away {
```
tarkah, line 4: This should live in / under buffer since it's on buffer and not top level to Config

tarkah (commented):
I think placement under buffer config is fine  👍

casperstorm:
This is good for review then @tarkah - just a small one.

tarkah (approved):
Great change!

PR #660, opened 2024-11-21
Miscellaneous Chat History Fixes

andymandias:
Two minor fixes.

1.  After one of the refactors history requests were no longer being made for the active buffer (due to metadata not being loaded, so there was no command to chain off of for the chathistory request).
2.  Requesting older chat history messages would fail for buffers with no messages in their history (because they produced an invalid chathistory request; `MessageReference::None` is not allowed for `BEFORE` chathistory requests).

### DIFF
--- a/src/screen/dashboard.rs
+++ b/src/screen/dashboard.rs
@@ -1488,14 +1488,25 @@ impl Dashboard {
                     &message_reference_types,
                 );
 
-                clients.send_chathistory_request(
-                    server,
-                    ChatHistorySubcommand::Before(
-                        target.clone(),
-                        first_can_reference,
-                        clients.get_server_chathistory_limit(server),
-                    ),
-                );
+                if matches!(first_can_reference, MessageReference::None) {
+                    clients.send_chathistory_request(
+                        server,
+                        ChatHistorySubcommand::Latest(
+                            target.clone(),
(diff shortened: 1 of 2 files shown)


src/screen/dashboard.rs
```diff
@@ -1488,14 +1488,25 @@ impl Dashboard {
...
+                } else {
+                    clients.send_chathistory_request(
+                        server,
+                        ChatHistorySubcommand::Before(
+                            target.clone(),
+                            first_can_reference,
+                            clients.get_server_chathistory_limit(server),
+                        ),
+                    );
+                }
```
casperstorm, line 1509: Nit: i think i would if/else the subcommand, and leave a single `send_chathistory_request` for readability.

```rust
let subcommand = if matches!(first_can_reference, MessageReference::None) {
    ChatHistorySubcommand::Latest(
        target.clone(),
        first_can_reference,
        clients.get_server_chathistory_limit(server),
    )
} else {
    ChatHistorySubcommand::Before(
        target.clone(),
        first_can_reference,
        clients.get_server_chathistory_limit(server),
    )
};

clients.send_chathistory_request(server, subcommand);
```

casperstorm (approved):
A single nit, but i'll approve since its nothing big.

PR #664, opened 2024-11-28
Add documentation about the halloy:// URL scheme.

funkeleinhorn:
This commit adds a small section about the halloy:// URL scheme to the URL Schemes page of the book.


casperstorm:
Awesome! Thanks.
Perhaps we should link to community theme where they are used?

casperstorm (approved):

Issue #600, opened 2024-09-30
Timeout for notifications


cvengler:
I currently have a notification sound for incoming direct messages.
When I issue a command to a bot, such as `/msg NickServ help`, it floods me with a few dozen direct messages in a few miliseconds, because it formats line breaks using it.
This triggers the notification sounds several dozen times, which not only sounds a little bit weird, but also makes the application unresponsive, thereby more likely to crash.

My proposed solution would be, that a single actor (e.g. a certain user) may only add a notification to the queue once every 500ms(?).

PR #665, opened 2024-12-06
Timeout delay for notifications

englut:
Added timeout delay for ~~direct message~~ all notifications to prevent multiple simultaneous messages flooding with notification sounds.

Default is set to 500ms.
Addresses issue #600. 

EDIT: Refactored for all notifications.

EDIT 2: I had looked into maybe turning `Notifications` into a trait. But I don't think that's what traits are meant for. So I'll leave it for now. If anyone has any suggestions for refactors or improvements, I am open :pray:

### DIFF
--- a/book/src/configuration/notifications.md
+++ b/book/src/configuration/notifications.md
@@ -19,6 +19,8 @@ Following notifications are available:
 | `disconnected`          | Triggered when a server disconnects                |
 | `file_transfer_request` | Triggered when a file transfer request is received |
 | `highlight`             | Triggered when you were highlighted in a buffer    |
+| `monitored_online`      | Triggered when a user you're monitoring is online  |
+| `monitored_offline`     | Triggered when a user you're monitoring is offline |
 | `reconnected`           | Triggered when a server reconnects                 |
 
 
@@ -39,3 +41,11 @@ Notification should trigger a OS toast.
 - **type**: boolean
 - **values**: `true`, `false`
 - **default**: `false`
+
+## `delay`
+
+Delay in milliseconds before triggering the next notification.
+
+- **type**: integer
+- **values**: `0` or greater
+- **default**: `500`
--- a/src/notification.rs
+++ b/src/notification.rs
@@ -1,7 +1,12 @@
+use std::collections::HashMap;
+use std::time::Duration;
+
+use chrono::{DateTime, Utc};
+
 use data::{
     audio::Sound,
+    client::Notification,
     config::{self, notification},
-    user::{Nick, NickRef},
 };
 
 use crate::audio;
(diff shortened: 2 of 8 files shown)


englut:
I'm going to refactor this to include all of the other notifications. So consider this in DRAFT mode (can't seem to move to draft) for now.

englut:
Alrighty, refactor done! Now supports all notification messages. I've tested a bunch of the notification types, most seem to work. I haven't been able to test `reconnect` yet. 

There is an issue with `monitored_online` notifications not triggering when the user I'm monitoring is online. I tested this on the release build, and I'm seeing the same issue there, so it may be an existing bug. I poked around, but it seems to require IRC proto knowledge, which I am not brushed up on at the moment. But because it's an existing bug, I am hoping this PR can still go through and it's sliiightly not related. :pray:  Maybe someone else can test on their release build to be sure?

book/src/configuration/notifications.md
```diff
@@ -38,4 +40,12 @@ Notification should trigger a OS toast.
 
 - **type**: boolean
 - **values**: `true`, `false`
-- **default**: `false`
\ No newline at end of file
+- **default**: `false`
+
+## `smart`
```
casperstorm, line 45: i think we should call it delay instead, since smart sounds like... its smart :D

englut: Done! `delay` makes way more sense :)

src/notification.rs
```diff
@@ -10,68 +15,132 @@ pub use self::toast::prepare;
...
+        }
 
-pub fn highlight(config: &config::Notifications<Sound>, nick: NickRef, channel: String) {
-    show_notification(
-        &config.highlight,
-        "Highlight",
-        format!("{} highlighted you in {}", nick, channel),
-    );
-}
+        match notification {
```
casperstorm, line 93: I don't like we have two match statements, and they end with `_ => {}`. When we then add new notifications it could be hard to remember to add them here, instead of relying on the compiler telling us we are missing an arm.

Instead, perhaps do something like in a simple `match notification {}`

```rust
match notification {
        Notification::Connected => {
            if let Some(server) = server{
                self._execute(&config.connected, notification, "Connected", server.clone());
            }
        }
[....]
}
```

src/notification.rs
```diff
@@ -10,68 +15,132 @@ pub use self::toast::prepare;
...
+    notified_times: HashMap<Notification, DateTime<Utc>>,
 }
 
-pub fn disconnected(config: &config::Notifications<Sound>, server: impl ToString) {
-    show_notification(&config.disconnected, "Disconnected", server);
-}
+impl Notifications {
+    pub fn new() -> Self {
+        Self {
+            notified_times: HashMap::new(),
```
casperstorm, line 25: Nit: "notified_times" is a bit confusing imo 🤔 
Perhaps something like:
- notification_timestamps
- recent_notifications 
or something similar?

src/notification.rs
englut: That's a very good point. I modified this to your suggestion :pray:

src/notification.rs
```diff
@@ -10,68 +15,132 @@ pub use self::toast::prepare;
...
-    nick: Nick,
-    server: impl ToString,
-) {
-    show_notification(
-        &config.file_transfer_request,
-        &format!("File transfer from {}", nick),
-        server,
-    );
-}
+    fn _execute(
```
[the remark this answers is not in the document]
englut: Removed!

src/notification.rs
englut: Yeah I had went through a few iterations of what to name this. I am not attached to `notified_times`. I was going to use `notified_at` at some point, but I think maybe `recent_notifications` makes most sense.

I've went ahead and made the change. Thank you!

casperstorm (approved):

Issue #494, opened 2024-08-16
Channel name is case sensitive


bmillham:
If you join a channel, say #Rockit and a message is send to #rockit halloy does not display the message.

The channel should not be case sensitive.

casperstorm:
@bmillham, could you show me how to reproduce this, and on which network?
I tried the following on Libera:

Test 1
1. Halloy joins `#FOObaz`
2. Second client joins `#foobaz`
3. Second client sends "hi"
4. Halloy sees "hi"

Test 2
1. Second client joins `#ROCKit`
2. Halloy joins `#rockit`
3. Halloy sends "hi"
4. Second client sees "hi"

glguy:
Libera normalizes channel names. You can't reproduce the bug there.

casperstorm:
> Libera (solanum) normalizes channel names. You can't reproduce the bug there.

That explains. Thank you.

gAlleb:
Hi @casperstorm 

I've encountered this bug too on russian ircnet.ru network.

If you connect to it via ircnet.ru:6688 server and join #usue channel - the channel you've joined will be `#Usue` and you will be able to see only messages from those who are connected to `#Usue` somehow but you'll see full users list.

If you connect to it via irc.386.ru:6688 server and join `#usue` channel - the channel you've joined will be `#usue` and you will be able to see only messages from those who are connected to `#usue` somehow but you'll see full users list.

With all other clients I've tried - there is no such behavior: every message from every user is seen on the channel. Must say it has confused me a lot from the start. But I do think it's connected with how Halloy treats case sensitivity. It joins the same channel butt somehow ignores some messages (the are not visible anywhere).

casperstorm:
Will start looking into this next @gAlleb!

gAlleb:
Thank you!

ircnet.ru doesn't use `Solanum` but historically uses `PTlink` (PTlink7.1.0-rc2). I'm not an expert but I assume that depending on the linked server inside this network it (the server) sends some kind of raw channel data that differs in cases. Maybe not. 

Because when you are inside the network and you `/join #usue` (or `/join #uSUe`) - you end up either on `#usue` or `#Usue` depending on the server you are being connected to. But i'm glad that could catch this cause at the beginning I thought that I'm going mad when I've seen a person talking and talking to himself :)

gAlleb:
> We need to handle this properly which will fix this issue

True. Looking forward. Thank you!

[...]

PR #666, opened 2024-12-06
Fix bug - case-sensitive channel name handling

vk3dan:
This seems to fix the issue described in https://github.com/squidowl/halloy/issues/494 where unneeded case-sensitivity in handling channel names causes people's messages to not be displayed if those people joined a channel using different capitalisation than you (on servers that do not normalise case). 

IRC standard specifies case-insensitive channel names per [RFC 2812](https://www.rfc-editor.org/rfc/rfc2812.txt#:~:text=Channel%20names%20are%20case%20insensitive.). 

Apologies if the way I've changed it isn't great or introduces another bug that I did not find testing this, I am not fluent in rust.

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -492,7 +492,7 @@ impl Client {
                             .and_then(|chathistory| chathistory.target())
                     })
                     .and_then(|target| {
-                        if self.chathistory_requests.contains_key(&target) {
+                        if self.chathistory_requests.contains_key(&target.to_lowercase()) {
                             Some(target)
                         } else {
                             None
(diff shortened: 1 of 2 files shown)


vk3dan:
No worries!

I had seen that they had self-assigned working on it in the issue page, I just wasn't sure at the time how far down the road a fix was going to be before implementation. Since this bug was the only thing preventing me from being able to switch over to halloy as my daily driver and I was enjoying the client overall, I figured I may as well have a go at it so that I can use it in the meantime and have a little look at rust in the process, once it was working I thought I should share it incase somebody else was in the same boat while the full fix was on its way :)

andymandias:
Appreciate the PR, sharing your workaround with the community.  Since #668 is merged into main it should no longer be necessary, so I'm going to close the PR.  Thank you!

PR #667, opened 2024-12-06
No Sent Messages as `chathistory` Reference

andymandias:
Sent messages may not be have been registered by the server yet.  This prevents the following bug:
1.  User is disconnected.
2.  Chat continues in the channel(s) the user is in while they're disconnected.
3.  User sends a message while disconnected.
4.  User reconnects.  Message sent in 3 is used as the reference for requesting chat history, but it was sent (locally) after all the messages in 2 were sent.  As a result, messages sent in 2 are never received.

At some point this could be refined to verify that the message has been received by the server, but for now skipping all sent messages will likely only result in a few extra messages being transmitted and avoids any potential for a false positive.

### DIFF
--- a/data/src/message.rs
+++ b/data/src/message.rs
@@ -189,7 +189,9 @@ impl Message {
     }
 
     pub fn can_reference(&self) -> bool {
-        if matches!(self.target.source(), Source::Internal(_)) {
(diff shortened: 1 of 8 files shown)


casperstorm (approved):
Looks good to me!

PR #668, opened 2024-12-06
`CASEMAPPING` Support

andymandias:
Uses the `CASEMAPPING` `RPL_ISUPPORT` parameter to normalize targets (user/channel) names, then uses normalized targets for routing.  Still undergoing testing and open to be refactored.

Known issue(s):
- [x] First version of target name seen (sent by the server/bouncer) is used in the sidebar.  If that is one that has been pre-normalized, then the normalized version of the target will appear in the sidebar.  E.g.  if a message has been sent by `Guest59` then it may appear in the sidebar as `guest59`.

### DIFF
--- a/data/src/buffer.rs
+++ b/data/src/buffer.rs
@@ -6,7 +6,8 @@ pub use self::away::Away;
 
 pub mod away;
 
-use crate::{channel, config, message, target, Server};
+use crate::target::{self, Target};
+use crate::{channel, config, message, Server};
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
 #[serde(untagged)]
@@ -62,10 +63,10 @@ impl Upstream {
         }
     }
 
-    pub fn target(&self) -> Option<target::Target> {
+    pub fn target(&self) -> Option<Target> {
         match self {
-            Self::Channel(_, channel) => Some(target::Target::Channel(channel.clone())),
-            Self::Query(_, query) => Some(target::Target::Query(query.clone())),
+            Self::Channel(_, channel) => Some(Target::Channel(channel.clone())),
+            Self::Query(_, query) => Some(Target::Query(query.clone())),
             Self::Server(_) => None,
         }
     }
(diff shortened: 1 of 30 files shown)


andymandias:
For now using incoming messages to resolve the case of the nickname of a user (used in the sidebar).  In testing it performs better than a more complicated `/WHOIS`/`/WHOWAS` scheme for nearly all cases, and it is *much* simpler to implement.

`CASEMAPPING` been working well in my testing, so I think it's ready for review when time permits.

englut:
@andymandias I took the liberty of fixing some merge conflicts (I've been merging your branch to help test!) caused by my merges.

I'm sure you can resolve the conflicts yourself, but in case you want a relatively quick patch:
[I've pushed up my fixes](https://github.com/englut/halloy/tree/feat/casemapping).

andymandias:
> @andymandias I took the liberty of fixing some merge conflicts (I've been merging your branch to help test!) caused by my merges.
> 
> I'm sure you can resolve the conflicts yourself, but in case you want a relatively quick patch: [I've pushed up my fixes](https://github.com/englut/halloy/tree/feat/casemapping).

Thanks for the reference!  Appreciate all the testing you've been doing, it's been invaluable! 🙏

data/src/buffer.rs
```diff
@@ -56,17 +55,17 @@ impl Upstream {
...
-    pub fn channel(&self) -> Option<&str> {
+    pub fn channel(&self) -> Option<&target::Channel> {
         match self {
             Self::Channel(_, channel) => Some(channel),
             Self::Server(_) | Self::Query(_, _) => None,
         }
     }
 
-    pub fn target(&self) -> Option<String> {
+    pub fn target(&self) -> Option<target::Target> {
```
tarkah, line 65: Nit: redundant, just import Target directly
 Same with elsewhere in the PR

tarkah (approved):
LGTM! It's a big diff but that mostly comes down to newtyping channel / query / target strings. 

It's nice that change enables us to layer in case mapping so easily. Thanks for tackling this!

casperstorm (approved):

PR #670, opened 2024-12-09
Initial support for a native Tor proxy

cvengler:
This commit implements native support for using Tor (i.e. not using an intermediate SOCKS5 connection) into Halloy, using Arti, the official Tor implementation in Rust.

It does so, by adding a third proxy type named `Tor`, that co-exists between `Socks5` and `Http`.  In order to achieve that change, a breaking change in the proxy syntax was necessary, as a native Tor proxy obviously does not need a host nor a port.

With this change, configuring a proxy looks like this:
```toml
[proxy]
Http = { host = "127.0.0.1", port = 9150 }
Socks5 = { host = "127.0.0.1", port = 9150 }
Tor = {}
```

Besides this, it also wraps the internally used `TcpStream` data structure behind an `IRCStream` enum which implements the `AsyncRead`, `AsyncWrite` and `Unpin` traits and also wraps the `DataStream` used by Tor into it.

It remains an open question how good this implementation is. Personally, I am a bit afraid that it will leak at a few places, namely the checking for updates as well as the file transfer feature.  Further evaluation is needed here.

Regarding the usefulness of the feature: Primarily I think that this is a neat addition by having a monolithic binary that uses Tor without depending on any other system resources.  Besides this, it may even be possible to drop support for `Socks5` and `Http` proxies entirely if this feature matures even further, as Tor is basically the only reason why modern applications still include support for `Socks5` (and maybe `Http`).  Alongside it would probably also make Halloy the first IRC client ever with native Tor support.  🙂

As a disclaimer: Professionally I am a part of the team responsible for developing Arti.  This pull request is a personal project however which I develop entirely in my free time.

### DIFF
--- a/book/src/configuration/proxy.md
+++ b/book/src/configuration/proxy.md
@@ -2,20 +2,15 @@
 
 Proxy settings for Halloy.
 
-## `type`
+1. [http](#proxyhttp)
+2. [socks5](#proxysocks5)
+3. [tor](#proxytor)
 
-Proxy type to use.
+## `[proxy.http]`
 
-```toml
-# Type: string
-# Values: http, socks5
-# Default: not set
+Http proxy settings.
 
-[proxy]
-type = "socks5"
-```
-
-## `host`
+### `host`
 
 Proxy host to connect to.
 
@@ -24,11 +19,13 @@ Proxy host to connect to.
 # Values: any string
 # Default: not set
 
-[proxy]
+# Required
+
+[proxy.http]
 host = "192.168.1.100"
 ```
- 
-## `port`
+
+### `port`
 
 Proxy port to connect on.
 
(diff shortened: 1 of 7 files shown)


cvengler:
It might also be a worthy discussion whether we should disallow plain-text IRC when having the Tor proxy defined, as using a plain-text protocol over Tor is obviously a huge compromise of privacy.

casperstorm:
Thanks for this PR! Will take a look at this during this week <3

tarkah:
Hey @cvengler thanks for this! Sorry for the delay on review. I've been wrapping up url previews but I'll look at digging into this after we get that merged. Just wanted to give you a heads up that we haven't forgotten about this.

data/src/config/proxy.rs
```diff
@@ -1,37 +1,48 @@
...
-
-#[derive(Debug, Clone, Deserialize)]
-pub struct Proxy {
-    #[serde(rename = "type")]
-    pub kind: Kind,
-    pub host: String,
-    pub port: u16,
-    pub username: Option<String>,
-    pub password: Option<String>,
+pub enum Proxy {
```
tarkah, line 4: Let's add `#[serde(rename_all = "snake_case")]` which would be consistent with having snake case keys for all of our config

book/src/configuration/proxy.md
```diff
@@ -6,47 +6,29 @@ Proxy settings for Halloy.
 
 ```toml
 [proxy]
-type = "socks5"
-host = "192.168.1.100"
-port = 1080
+Socks5 = { host = "192.168.1.100", port = 1080 }
```
tarkah, line 9: How does toml handle a unit type for `Tor` variant? 

```toml
[proxy]
Tor
```

Doesn't seem like valid toml but I'm on my phone and can't validate. Or do we need:

```toml
[proxy]
Tor = {}
```

Might be good to document

irc/src/connection.rs
```diff
@@ -1,8 +1,10 @@
 use std::net::IpAddr;
 use std::path::PathBuf;
+use std::pin::Pin;
 
+use arti_client::DataStream;
```
tarkah, line 5: Can we alias this to `TorStream`? `DataStream` is a bit generic to know where it's coming from

tarkah (commented):
Code looks good! A couple minor nits. I'll want to test this and assess the dependency used prior to merging. Thanks for this!

book/src/configuration/proxy.md
cvengler: The latter one is correct.

cvengler:
@tarkah is there any release deadline or something coming closer? I would prefer to address the issues you have mentioned next week due to time issues.

casperstorm:
> @tarkah is there any release deadline or something coming closer? I would prefer to address the issues you have mentioned next week due to time issues.

Should be fine, we already decided to wait for a few big features so we can use the extra testing period before we release.
We can easily wait for this one as well.

cvengler:
I have adressed the issues and rebased it to the current main branch.

tarkah:
@cvengler Thanks! I've updated the dependency to use `rustls` since that's what we use for our TLS connection and I added the feature to compile sqlite statically since otherwise the build isn't as portable & also fails to build on windows.

I tested w/ ergo testnet and the proxied connection works :+1: 

I noticed Libera doesn't allow these tor proxied connections and instead suggests connecting to their onion service. It seems we can do that w/ `arti-client` by enabling the `onion-service-client` feature, but I wasn't able to get this working w/ libera. Maybe we can eventually get that enabled / figured out.

I think this is good to go!

tarkah (approved):

tarkah:
@casperstorm can you confirm testing on windows & mac to `testnet.ergo.chat`?

PR #674, opened 2024-12-12
Highlight notifications for `/me` actions.

englut:
Found bugs with highlighting notifications when someone mentions you with a `/me` action.

There's three related issues: 
- Notification events for `/me` action highlights are not triggered.
- `/me` action is not visually highlighted.
- `/me` action highlights not showing in Highlights buffer.
- Highlights in the query buffer is missing.

Above issues should be addressed with this PR.

### DIFF
--- a/src/buffer/channel.rs
+++ b/src/buffer/channel.rs
@@ -151,15 +151,14 @@ pub fn view<'a>(
                             .push(nick)
                             .push(space);
 
-                        let text_container =
-                            container(message_content).style(move |theme| match our_nick {
-                                Some(nick)
-                                    if message::references_user(user.nickname(), nick, message) =>
-                                {
-                                    theme::container::highlight(theme)
+                        let text_container = container(message_content).style(move |theme| {
+                            if let Some(nick) = our_nick {
+                                if message::references_user(user.nickname(), nick, message) {
+                                    return theme::container::highlight(theme);
                                 }
-                                _ => Default::default(),
-                            });
+                            }
+                            Default::default()
+                        });
 
                         match &config.buffer.nickname.alignment {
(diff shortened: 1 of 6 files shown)


englut:
Update! I figured out how to get the `ACTION` messages to be highlighted. Needed to update `message::Source::Action` to include `User` optionally. 

There is 1 more issue, which is having these highlights displayed in the `Highlights` buffer. I'll keep digging to see if I can figure this out too!

englut:
Update 2: Figured it out :pray: But I found another issue.... mentions in DMs are not sent to the Query window.. which seems pretty bad. Will fix this before marking this PR as ready.

src/buffer/channel.rs
```diff
@@ -203,25 +203,42 @@ pub fn view<'a>(
...
+                                            message,
+                                        ) =>
+                                    {
+                                        theme::container::highlight(theme)
+                                    }
+                                    _ => Default::default(),
+                                },
+                                _ => Default::default(),
+                            });
+
```
casperstorm, line 233: This could be simplified a bit, not tested but something like:

```rust
let text_container = container(message_content).style(|theme| {
    if let (Some(user), Some(nick)) = (user, our_nick) {
        if message::references_user(user.nickname(), nick, message) {
            return theme::container::highlight(theme);
        }
    }
    Default::default()
});
```

englut: Ah! That's so much nicer. I didn't know that you could combine the `Some(..)` assignment that way. I did think that there was something to optimize there! Thanks for this :pray:

englut:
I noticed that highlights in query weren't present. I've added this to the PR.

casperstorm (approved):

PR #675, opened 2024-12-12
`chathistory` User Command Support

andymandias:
In the event that `chathistory` malfunctions, it would be nice to have some user remedy (beyond wiping their history clean).  So I've added some support for the user sending their own `chathistory` commands.  In particular:

1. Command completion for `chathistory` commands (and fix a minor issue in command completion for sub-commands)
2. Send the output of the `CHATHISTORY TARGETS` command to the server buffer (when it's user requested).

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -1571,6 +1571,8 @@ impl Client {
                 }
             }
             Command::CHATHISTORY(sub, args) => {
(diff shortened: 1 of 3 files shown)


casperstorm (approved):
Looks good! @tarkah you have anything before we merge?

Issue #673, opened 2024-12-12
nick_password_command server configuration not working


theRAAPster:
I'm trying to use the nick_password_command server configuration, but when joining the server I'm getting password mismatch errors. I'm running Halloy 2014.14 on OS X 15.1.1, using the onepassword CLI to pull a password from my vault. The command seems to work correctly from my own shell, and for testing purposes I've removed all special characters from the password to make sure there wasn't an issue there.

I couldn't really find detailed logging about what Halloy is pulling in with the command, but looking at the code I think I've found an issue. Should lines 124 and 127 be config.nick_password instead of config.password? My apologies if I'm off here since I'm not a developer, but it didn't look right compared to the previous if statements in that function.

https://github.com/squidowl/halloy/blob/591b728fdec62fe96f8d26aa6a9ea644f9f5ea74/data/src/server.rs#L123-L128

andymandias:
I believe you have correctly diagnosed the problem.  If you want to make it into a pull request I'd be happy to review it.  Or I can make the pull request if you'd prefer not to make it yourself.

PR #692, opened 2025-01-05
Bugfix for nick_password_command server configuration

theRAAPster:
This is a potential fix for #673,  however I lack the skills necessary to build/test this change to confirm it does solve the problem.


casperstorm (approved):
This looks correct to me. @andymandias?

andymandias (approved):
LGTM as well.  Maybe we should note the fix in the changelog?

PR #696, opened 2025-01-09
Autocomplete configuration

### LINKED ISSUE #650: Avoid adding : when tab completing something that is not a nickname in the channel
Current: typing out characters that is not an active nickname in the channel makes Halloy add ": ".

Wish: Nothing should happen if the characters at the caret are not an active user in the channel.

Optionally: configuration for how tab complete behaves. Adding : or other character(s) or just completing the nick. Could be simply a string of what to add after the nick, defaulting to ": ".

casperstorm:
Fixes #650.

Fixed
- Don't add suffix to word if there is no valid nick to autocomplete

Added
- Configuration for changing suffix and sorting direction when autocompleting


andymandias (approved):
LGTM!  Tested, works as described.

Issue #584, opened 2024-09-21
WHO is being rate-limited


aeadio:
Reopening #293 

This appears to either be unfixed, or else a regression has occurred. Happening on the latest release, 2024.12 (MacOS). 

`who_poll_interval` and `who_retry_interval` within a `[servers]` block seem to be ignored. Snippet from config:

```
[servers.znc-libera]
server = "<redacted>.ts.net"
port = 6501
use_tls = false
username = "<redacted>/libera"
password = "<redacted>"
nickname = "<redacted>"
realname = "<redacted>"
who_poll_interval = 3600
who_retry_interval = 3600
```

I get spammed with rate-limited messages regardless of which value is provided to those tunables. In fact, the rate limiting seems to be so bad, I can't do anything on Libera after startup -- like join additional channels or send messages.

I'm not interested in seeing who is away, so it's bizarre to me that this feature is not configurable.

andymandias:
ZNC [supports away-notify](https://ircv3.net/software/clients), so you shouldn't be seeing continued WHO spam. There will be a burst when launching Halloy, but that should be it. What version of ZNC and enabled modules are you running?  I would like to reproduce this behavior.

casperstorm:
Thanks for (re)opening the issue! Slightly unrelated to it, but to address:

> I'm not interested in seeing who is away, so it's bizarre to me that this feature is not configurable.

If there's one thing I've learned from working on Halloy, it's that no matter how much you can configure, it's never enough 😅. People who use IRC as their daily driver in 2024 each have very specific requirements, and for a new client like ours, it's hard to nail everything on the first try. Opening feature requests is the best way to let us know something is missing!

That said, it does sound a bit strange with the bug, and not something i can reproduce. Could this be a issue on your ZNC end?

aeadio:
ZNC doesn't exhibit the issue with WeeChat on Linux. I'm running ZNC version 1.9.1. I'm not running a lot of modules and I haven't changed most of the settings from the defaults. 

Global modules

- corecaps

User modules

- chansaver
- controlpanel
- log

Server modules

- log
- sasl
- simple_away

I'm currently in 25 channels. I don't know if the WHO spam is just the initial scrape when Halloy starts, but I've left it up for at least 15 minutes and it's still going at that point. The client is completely unusable on Libera during this time.

aeadio:
[...]

PR #699, opened 2025-01-10
WHO Polling Fixes/Improvements

andymandias:
This contains a number of changes to WHO polling, primarily to address the behavior reported in #584.  Includes:

- Use WHO polling queue to perform WHO polls in serial, rather than in parallel.  Greatly reduces the effective rate of WHO polling for users in more than a few channels (the default rate has been increased a bit as a result, but should still be slower than before for most users).  And, ensures that the maximum WHO polling rate can be more-straightforwardly controlled by the user configuring `who_poll_interval` and `who_retry_interval`.
- Switches the WHO show/hide logic to hide all WHO replies unless a user initiated WHO request was sent.  Previously WHO replies would be shown unless a WHO poll was sent.  Ensures WHO replies sent by other clients are not shown, even if they are not routed to the requesting client (e.g. when using ZNC with multiple clients & without route_replies enabled).
- Fixes a bug where WHO polls that were sent before CAP negotiation completed could be parsed incorrectly (effectively throwing away the result of the first WHO poll request).

In testing this allows server WHO rate-limit to be better respected, but I do see some rate-limit notifications due to the initial WHO polling initiated by channel JOINs on Libera.  I expected those WHO polls to be valid since Solanum appears to give a "WHO credit" for each JOIN, so that remains to be sorted out.

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -267,35 +267,39 @@ impl Client {
         if matches!(message.command, Command::WHO(..)) {
             let params = message.command.clone().parameters();
 
-            if let Some(target) = params.first().and_then(|target| {
-                target::Channel::parse(
-                    target,
+            if let Some(mask) = params.first() {
+                let channel = if let Ok(channel) = target::Channel::parse(
+                    mask,
                     self.chantypes(),
                     self.statusmsg(),
                     self.casemapping(),
-                )
-                .ok()
-            }) {
-                // Record user WHO request(s) for reply filtering
-                let status = WhoStatus::Requested(
-                    WhoSource::User,
-                    Instant::now(),
-                    params
-                        .get(2)
-                        .and_then(|token| token.parse::<WhoToken>().ok()),
-                );
-
-                if let Some(who_poll) = self
-                    .who_polls
-                    .iter_mut()
-                    .find(|who_poll| who_poll.channel == target)
-                {
-                    who_poll.status = status;
+                ) {
+                    Some(channel)
+                } else if mask == "*" {
+                    Some(target::Channel::from_str(mask, self.casemapping()))
                 } else {
-                    self.who_polls.push_front(WhoPoll {
-                        channel: target,
-                        status,
-                    });
+                    None
+                };
+
+                if let Some(channel) = channel {
+                    // Record user WHO request(s) for reply filtering
+                    let status = WhoStatus::Requested(
+                        WhoSource::User,
+                        Instant::now(),
+                        params
+                            .get(2)
+                            .and_then(|token| token.parse::<WhoToken>().ok()),
+                    );
+
+                    if let Some(who_poll) = self
+                        .who_polls
+                        .iter_mut()
+                        .find(|who_poll| who_poll.channel == channel)
+                    {
+                        who_poll.status = status;
+                    } else {
+                        self.who_polls.push_front(WhoPoll { channel, status });
+                    }
                 }
             }
         }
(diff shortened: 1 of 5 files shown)


data/src/client.rs
```diff
@@ -1246,47 +1286,58 @@ impl Client {
...
+                        } else {
+                            self.who_polls.push_front(WhoPoll {
+                                channel: target_channel.clone(),
+                                status: WhoStatus::Requested(WhoSource::Join, Instant::now(), None),
+                            });
+
+                            self.handle.try_send(command!("WHO", channel))?;
                         }
+
+                        log::debug!("[{}] {channel} - WHO requested", self.server);
```
tarkah, line 1333: Logging needs to only happen once we actually send the command, otherwise it's misleading

data/src/client.rs
```diff
@@ -1246,47 +1286,58 @@ impl Client {
...
+                                whox_params.fields().to_string(),
+                                whox_params.token().to_owned()
+                            ))?;
+                        } else {
+                            self.who_polls.push_front(WhoPoll {
+                                channel: target_channel.clone(),
+                                status: WhoStatus::Requested(WhoSource::Join, Instant::now(), None),
+                            });
+
+                            self.handle.try_send(command!("WHO", channel))?;
```
tarkah, line 1330: We are still sending out `num_channels * WHO` without any queue / buffering in-between. Yes, our retry logic is queued / buffered, but this initial spam of WHO might still cause issues due to rate limiting or contribute to a rate limit problem

Why not queue up everything and pop them off initially sequentially?

data/src/client.rs
```diff
@@ -1322,73 +1373,153 @@ impl Client {
...
+                        {
+                            self.who_polls.push_back(who_poll);
+                        }
+                    }
+
+                    if let Some(who_poll) = self.who_polls.front_mut() {
+                        if matches!(who_poll.status, WhoStatus::Received) {
+                            who_poll.status = WhoStatus::Waiting(Instant::now());
+                        }
                     }
```
tarkah, line 1514: After the initial WHO off all channels and the way the queue works, won't this cause us to only fetch 1 channel between `who_poll_interval` durations?

So if `who_poll_interval` is `60s` and they have `10` channels, it'll take `10 minutes` to refresh all 10 channels?

I'm not sure this is the desired result. I'd expect `who_poll_interval` to be the amount of time it takes between trying to poll for WHO in aggregate? Obviously if the user has 50 channels but a poll interval of `60s` it'll need to take way longer than `60s` between WHO refreshes due to rate limiting.

Do we need a different parameter to configure how long between channels? 

This number will be tricky to figure out and I'm not sure if it should be configurable. I'm curious if we instead dynamically find some ideal delay via some exponential backoff calculation via `RPL_TRYAGAIN`?

data/src/client.rs
andymandias: This is unchanged from how it was being logged previously, but moving it makes sense to me.  I'll put the logging in `data/src/stream.rs` similar to how QUIT is logged.

data/src/client.rs
andymandias: I was primarily basing this off of solanum/Libera which afaict gives a "WHO token" for each channel joined that allows for one non-limited WHO.  Assumed that meant there was a cultural expectation to request a WHO per join, so I went with that.  I'm not attached to that behavior at all, and in fact I think it's probably preferable to operate more politely by default by adding channels to the queue as you suggest.

data/src/client.rs
andymandias: My thinking here was that servers probably have rate limits that are not dependent on how many channels that the user is connected to (I have not surveyed to server software to determine this).  If that's the case, then a `who_poll_interval` that would be rate-limiting for a user that's in a 100 channels would be greatly different than a user who is in 2/10/54 channels.  And a user that joins a couple channels may suddenly find themselves rate-limiting when they weren't previously.  In that sense I think it would be helpful, if we're making it configurable, to have the setting directly control how frequently WHO polls are sent; that way we could, for example, tell an inquiring user to set `who_poll_interval` to #s for Libera and wouldn't have to worry about the rest of their configuration.

Regardless of the configuration setting (or lack thereof) that gets chosen, having an exponential backoff via `RPL_TRYAGAIN` makes sense to me.

data/src/client.rs
tarkah: Sorry I misread these, I thought we were only pushing these to the queue when I first read it, disregard

andymandias: No problem, I'll revert the change :)

data/src/client.rs
andymandias: If I'm understanding your question correctly, we still get rate limiting because the who token is only good for the first who after joining.  When the next poll comes around there aren't any tokens to use, and the 50 channels having their requests sent all at once are no longer tolerated.

casperstorm (approved):
I want @tarkah last say on this, but looks good to me now.

tarkah (approved):
LGTM. I haven't had a chance to test, but I know you're pretty thorough @andymandias. Thanks for tackling this!

Issue #411, opened 2024-07-09
Window position and size not restoring 2024.8


darienm:
[I did not see an option to re-open the issue that claims this feature was implemented #391, so here is a new issue. Feel free to delete or merge or whatever is appropriate.]

Version 2024.8 MacOS does not return the window to the last used position or size. It's sometimes in a similar region on-screen (no longer centered) but not particularly near where it was last closed, which doesn't achieve the goal. Do you need more info? Primary screen is 3840x2160 and secondary screen is 2560x1600. (1) Window does not return to proper screen (2) Window seems to get smaller with each launch. (3) Window seems to disregard X,Y positions greater than around 1400x1400.

casperstorm:
I have debugged this, and it seems like there are some strange things going on.
When setting the `window.json` to

```json
{"position":{"x":1000.0,"y":500.0},"size":{"width":500.0,"height":500.0}}
```

I notice it gets overwritten right after launch with the following on Windows by a `iced::window::Event::Moved` and `Resized `event.

```json
{"position":{"x":909.0909,"y":454.54544},"size":{"width":454.54544,"height":454.54544}}
```

This only happens on Windows though, on Mac it doesnt get replaced like on Windows - but it sometimes ends up in a strange position, like you describes. I don't think its a issue on Halloys end, but rather a issue with winit 🤔 
@hecrj have you noticed or seen anything similar?

casperstorm:
Yeah, I think its a upstream issue. On Halloy we really don't do anything special other than capturing the move/resize events as can be seen [here](https://github.com/squidowl/halloy/blob/main/src/window.rs#L115-L120) and then save them as json, as can be seen [here](https://github.com/squidowl/halloy/blob/main/data/src/window.rs#L42).

sr-gi:
I can confirm this happens on MacOS too. `window.json` gets re-written every time Halloy is opened

PR #702, opened 2025-01-11
Restore window position and size.

casperstorm:
Fixes https://github.com/squidowl/halloy/issues/411.

I've introduced a hack to skip a few initial events which was causing problems on Windows due to a winit bug: https://github.com/rust-windowing/winit/issues/2094.

Tested on
- [X] Windows
- [X] macOS
- [x] Linux

Could you test on Linux @andymandias or @tarkah?

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -30,7 +30,7 @@ use data::{client::Notification, environment, server, version, Server, Url, User
 use iced::widget::{column, container};
 use iced::{padding, Length, Subscription, Task};
 use screen::{dashboard, help, migration, welcome};
-use tokio::runtime;
+use tokio::{join, runtime};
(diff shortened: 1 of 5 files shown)


src/main.rs
```diff
@@ -66,14 +66,18 @@ pub fn main() -> Result<(), Box<dyn std::error::Error>> {
...
 
-        rt.block_on(Config::load())
-    };
+        rt.block_on(async { 
+            let config = Config::load();
+            let window = data::Window::load();
 
+            join!(config, window)
+        })
+    };
```
tarkah, line 80: We're already inside an `async` block here, let's just `.await` each one, no need for `join`

tarkah (approved):

PR #710, opened 2025-01-16
Allow Whitespace before Terminating CRLF

andymandias:
DejaToons sends some messages with a space between the message content and `\r\n`.  This change is to permit that by ignoring spaces preceding `\r\n` (or `\r\r\n`).


casperstorm (approved):

tarkah (commented):
Shouldn't this whitespace be part of the last parameter of the command?

andymandias:
In the test message I encountered this issue with, I don't believe so; the modestring should not have whitespace in it. But I did not consider the trailing parameter fully. 🤔

PR #714, opened 2025-01-17
Expand Echo Search

andymandias:
Increases the search time interval when inserting an echo (message sent by the user received from the server) into message history.  This is to help address the duplicate messages issue discussed in #709.

### DIFF
--- a/data/src/message.rs
+++ b/data/src/message.rs
@@ -155,7 +155,8 @@ impl Target {
 #[derive(Debug, Clone, Copy, Serialize, Deserialize)]
 pub enum Direction {
     Sent,
-    Received,
+    /// is_echo
+    Received(bool),
 }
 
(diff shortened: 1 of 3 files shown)


casperstorm (approved):
Looks good to me

data/src/message.rs
```diff
@@ -155,7 +155,8 @@ impl Target {
 #[derive(Debug, Clone, Copy, Serialize, Deserialize)]
 pub enum Direction {
     Sent,
-    Received,
+    /// is_echo
+    Received(bool),
```
tarkah, line 159: @andymandias This breaks `history`. All my messages got deleted :(

We need to make sure we add things in a backwards compatible way


This means we need a custom De-serialize implementation for `Direction`

PR #719, opened 2025-01-22
Add link to base16-halloy repo

4e554c4c:


casperstorm:
When adding a new page you have to update the SUMMARY.md as well.
The normal work flow is to run `mdbook serve` and then when you change the file it automatically add the new chapters (files).

casperstorm (approved):

PR #723, opened 2025-01-23
Minor `chathistory` Tweaks

andymandias:
Fixes a couple minor issues in the `chathistory` implementation that were found when testing #720.


tarkah (approved):
Thanks!

PR #724, opened 2025-01-23
Fix various performance issues

### LINKED ISSUE #640: rfc: refactoring `Server`
## The Problem

When working on #77 I have run into numerous issues with how to represent a "server" in halloy. The `Server` struct, which is a wrapper around `String`, is currently ubiquitous in the halloy codebase, however it has many issues. Since `Server` is a `String`, `&Server` is `&String`, which is an anti-pattern. Variables of type `&Server` must be `.clone()`d often to satisfy lifetime issues and in order to satisfy lifetime constraints, which leads to the same "server name" being duplicated all over the codebase. For example, it must be cloned every time a message is received by the IRC server to communicate the server updated between the "stream" and "client".

In order to implement #77, the way that servers are referenced must be changed as well. In a bouncer-networks setup, several "primary" servers each set up any number of secondary servers. Thus there is a hierarchical relationship between servers, and a single config is shared between many. Furthermore, servers can no longer be identified by their name. Names may not be unique (two secondaries on under two different primaries may match), and are not the canonical way to refer to a server (the canonical identifier for a "secondary" server is its bouncer NETID).

Therefore, however we represent `Server`, it cannot also be used as the human-visible name for a server, which causes issues elsewhere in the codebase

## A Proposed Solution

Ideally, in order to send a `Server` identifier with every single irc message, the `Server` type should be `Copy` (not just `Clone`!) and cheap to replicate. Furthermore, we should be able to get additional data that a `Server` represents by asking a centralized source of information.

My proposal is as follows: `Server` should be a `Uuid` (either v4 or v5). This data can be cheaply copied and sent between tasks without a performance overhead. Then, when it is necessary to look up server information (such as user-visible name) this can be gathered by looking at the a `HashMap<Uuid, ServerInfo>` which contains server config name, bouncer ID, bouncer name, etc.

### CCs 
@tarkah , @casperstorm let me know what you think. Last time I brought this up you mentioned that I should open an RFC before doing any huge refactors with `Server` and friends :smile_cat:

4e554c4c:
This commit includes various fixes that I have found working on bouncer-networks. It includes several performance benefits—such as changing `Server` to be an `Arc<str>` which removes much of the copying overhead. I have also made sure that there is only one copy of the server config in the `Halloy` struct, which was previously duplicated for seemingly no reason.
Still, several performance issues go un-addressed. such as the duplication of server configs between threads (should this also be an `Arc<config>`?) and the copying of `client` structs.

I have also made some changes to the client. The capability negotiation process has been streamlined, and I've eliminated some branching with preference toward pattern matching. I've also fixed a slight error with halloy, making sure that it detects the end of registration correctly (matching other clients).

I have also made changes to the `Cargo.toml` in the `data` crate. This is to fix some compilation issues when building that crate in isolation. I would like this fixed so that it'd be easier to run `data` crate tests in isolation.
Ideally I think that each of the sub-crates should be tested individually in CI—but that's out of scope for this PR.

In my opinion, this resolves #640

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -159,6 +159,20 @@ impl fmt::Debug for Client {
         f.debug_struct("Client").finish()
     }
 }
+const UNCONDITIONAL_CAPS: &[&str] = &[
+    "account-notify",
+    "away-notify",
+    "batch",
+    "chghost",
+    "draft/read-marker",
+    "extended-monitor",
+    "invite-notify",
+    "labeled-response",
+    "message-tags",
+    "multi-prefix",
+    "server-time",
+    "userhost-in-names",
+];
 
 impl Client {
     pub fn new(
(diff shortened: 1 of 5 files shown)


data/src/client.rs
```diff
@@ -155,6 +155,19 @@ impl fmt::Debug for Client {
         f.debug_struct("Client").finish()
     }
 }
+const UNCONDITIONAL_CAPS: &[&str] = &[
```
andymandias, line 158: I like that this synchronizes this aspect of the CAP requests.  Based on the name I would expect `labeled-response` and `account-notify` in here as well.  And `soju.im/bouncer-networks` probably shouldn't be in here yet.

data/src/client.rs
```diff
@@ -646,92 +659,64 @@ impl Client {
...
-                            requested.push("extended-join");
-                        }
+                    if contains("extended-join") {
+                        requested.push("extended-join");
                     }
-                    if contains("batch") {
-                        requested.push("batch");
+                }
+                if contains("batch") {
+                    requested.push("batch");
```
andymandias, line 685: I think this should be covered by the `UNCONDITIONAL_CAPS`, right?

data/src/client.rs
4e554c4c: yeah i was on the fence about these, but I think that the name definitely implies that I should add them. changed.

Also thanks for catching `bouncer-networks`, I accidentally staged that one.

casperstorm (approved):
Looks good to me, thanks for a great PR @4e554c4c.
I'd like to see if @tarkah has any comments before we merge it.

data/src/client.rs
```diff
@@ -791,33 +765,16 @@ impl Client {
...
-                    requested.push("server-time");
-                }
-                if newly_contains("chghost") {
-                    requested.push("chghost");
-                }
-                if newly_contains("extended-monitor") {
-                    requested.push("extended-monitor");
-                }
                 if contains("account-notify") || newly_contains("account-notify") {
                     if newly_contains("account-notify") {
```
andymandias, line 779: I think we can skip this and condense the conditionals, similar to what you did above.  Everything else looks good to me.

4e554c4c:
updated

I experimented with some other things (in particular using `HashSet` operations to update caps) that didn't end up so well due to struct field borrowing issues...
Eventually I think it'd be nice to move this into a `Cap` struct that handles most of the logic, since that would simplify a lot (especially because the logic for `CAP NEW` and `CAP LS` should basically be equivalent).

data/src/client.rs
```diff
@@ -1168,11 +1103,18 @@ impl Client {
                 // Updated actual nick
                 let nick = ok!(args.first());
                 self.resolved_nick = Some(nick.to_string());
+            }
+            // end of registration (including ISUPPORT) is indicated by either RPL_ENDOFMOTD or
```
andymandias, line 1107: From the link on the next line it reads as though `RPL_WELCOME` through `RPL_ENDOFMOTD` or `ERR_NOMOTD` is technically all post-registration.  It doesn't really matter what that series of messages is called though, this update is correct;  we were utilizing `RPL_ISUPPORT` parameters in channel parsing for sending `JOIN`s, but feature advertisements come *after* `RPL_WELCOME` (and before `RPL_ENDOFMOTD` or `ERR_NOMOTD`).

andymandias (approved):
Looks good to me, and I didn't find any issues in testing.

src/main.rs
```diff
@@ -327,13 +331,12 @@ impl Halloy {
...
                             Ok(updated) => {
                                 let removed_servers = self
-                                    .servers
+                                    .servers()
                                     .keys()
                                     .filter(|server| !updated.servers.contains(server))
                                     .cloned()
                                     .collect::<Vec<_>>();
 
-                                self.servers = updated.servers.clone();
```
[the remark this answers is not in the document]
4e554c4c: i would be ok with this if there were only one `Config`, but server configs are copied _everywhere_ which is a huge pain for keeping things in sync.

Not to mention that the current logic deletes all servers which are not in the config whenever the config updates, *including* dynamically added servers, so these two structs are essentially treated as equal except in some edge cases

data/src/server.rs
```diff
@@ -14,23 +15,35 @@ use crate::config::Error;
 pub type Handle = Sender<proto::Message>;
 
 #[derive(Debug, Clone, Hash, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
-pub struct Server(String);
+#[serde(transparent)]
+pub struct Server {
+    name: Arc<str>,
```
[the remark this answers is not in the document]
4e554c4c: Unfortunately servers and server names are traeted as the same thing basically all throughout the codebase. The config is treated as the de-facto location where servers are stored (with everything else being cloned from there) so we kind of have to start changing things here unless we change the entire codebase

Doing things this way makes the [bouncer networks change](https://github.com/squidowl/halloy/pull/730/commits/f6c8655a13ed547d6e2fac748bcaf2c5900a568e) a fairly modest ~100 patch before bouncer-networks specific logic which i thought would be better than totally refactoring everything

src/main.rs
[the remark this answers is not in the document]
tarkah: > i would be ok with this if there were only one Config, but server configs are copied everywhere which is a huge pain for keeping things in sync.

The only thing accessing `Config.servers` is us cloning it to `Halloy.servers`. And the only thing `Halloy.servers` is used by is logic for diffing when new servers are added (via `irc://`) or removed (via config refresh) and then it's our source of truth for which server subscriptions we maintain (active IRC connections). 

But by using `Config.servers` as our source of truth, we have to mutate our config at runtime, which is a hard no, and we also lose control of how to manage runtime servers outside the context of "configuration"

4e554c4c:
closing this until v2

PR #731, opened 2025-01-24
`chathistory` Highlights

andymandias:
Add messages received via `chathistory` request(s) to the highlight buffer (as appropriate).

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -642,6 +642,22 @@ impl Client {
                                 if ctcp::is_query(text) && !message::is_action(text) {
                                     // Ignore historical CTCP queries/responses except for ACTIONs
                                     vec![]
+                                } else if let Some(user) = message.user().filter(|user| {
+                                    message::references_user_text(
+                                        user.nickname(),
+                                        self.nickname(),
+                                        text,
+                                    )
+                                }) {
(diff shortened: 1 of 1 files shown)


data/src/client.rs
```diff
@@ -643,7 +643,29 @@ impl Client {
...
+                                                Notification::Highlight {
+                                                    enabled: false,
+                                                    user,
+                                                    target,
+                                                },
+                                            ));
+                                        }
+                                    }
+
+                                    events.push(Event::Single(message, self.nickname().to_owned()));
```
tarkah, line 666: `Single` should be on the else branch since `Event::Notification` already adds the message to history, so we're adding it to history twice currently

andymandias: Ahh, right.  Thanks for catching that!  The downside of deduplication is this didn't immediately show up in testing.

tarkah: Lol true. I found a semi related issue where I hashed messages on received time instead of server time, which would be hit by this dedupe bug.

The cloned highlight message would have the older received time and the go to message functionality would be broken because of this. 

I've fixed that in my url preview branch btw

tarkah (approved):
Thanks!

PR #732, opened 2025-01-25
Expand Echo Search (Take Two)

andymandias:
Increases the search time interval when inserting an echo (message sent by the user received from the server) into message history.  Unlike the previous implementation (#714), existing history is properly deserialized.

### DIFF
--- a/data/src/message.rs
+++ b/data/src/message.rs
@@ -299,7 +299,7 @@ impl Message {
 
         Message {
             received_at,
-            server_time: Utc::now(),
+            server_time,
             direction: Direction::Received,
             target: Target::Query {
                 query: query.clone(),
@@ -472,7 +472,7 @@ impl<'de> Deserialize<'de> for Message {
             Content::Plain("".to_string())
         };
 
-        let is_echo = is_echo.unwrap_or(false);
+        let is_echo = is_echo.unwrap_or_default();
 
         let hash = Hash::new(&server_time, &content);
 
--- a/data/src/message.rs
+++ b/data/src/message.rs
@@ -172,11 +172,13 @@ pub struct Message {
     pub id: Option<String>,
     pub hash: Hash,
     pub hidden_urls: HashSet<Url>,
+    pub is_echo: bool,
 }
 
 impl Message {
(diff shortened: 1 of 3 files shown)


data/src/message.rs
```diff
@@ -152,10 +152,49 @@ impl Target {
...
+            Received,
+        }
+
+        let data = Data::deserialize(deserializer)?;
+
+        Ok(match data {
+            Data::Sent => Direction::Sent,
+            Data::Received => Direction::Received { is_echo: false },
+        })
+    }
```
tarkah, line 197: I don't love that we're discarding this from state, so it's effectively a run-time only variable that's discarded upon restart.  There's nothing preventing someone from depending on this field in more complex ways that then breaks due to this.

As much as I'd love to add this to `Direction`, it doesn't need to be encoded on it. Let's just add a new `is_echo` field to the message and this is backwards compatible since older clients will just ignore that in the json.

tarkah (commented):
Server time is defined as a variable above

data/src/message.rs
andymandias: Was hoping there would be some serde incantation that would preserve `is_echo` only when `Direction::Received`, but I couldn't figure it out.  Your rationale for the new field makes sense to me → updated (& tested) the implementation.

data/src/message.rs
```diff
@@ -292,7 +299,7 @@ impl Message {
 
         Message {
             received_at,
-            server_time,
+            server_time: Utc::now(),
```
tarkah, line 302: Server time is a variable above

data/src/message.rs
```diff
@@ -456,6 +472,8 @@ impl<'de> Deserialize<'de> for Message {
             Content::Plain("".to_string())
         };
 
+        let is_echo = is_echo.unwrap_or(false);
```
tarkah, line 475: Nit: unwrap or default

andymandias:
> Server time is defined as a variable above

Ah shoot, didn't catch rebase adding that.

tarkah (approved):
LGTM, thanks!

PR #735, opened 2025-01-26
`setname` and Standard Replies

andymandias:
Simple, straightforward implementations for [Standard Replies](https://ircv3.net/specs/extensions/standard-replies) and [`setname`](https://ircv3.net/specs/extensions/setname.html).  Bundled because `setname` will produce a Standard Reply when it fails, providing an easy way to trigger a `FAIL` Standard Reply (on InspIRCd servers).

Standard Replies introduces three new colors to themes (for `FAIL`, `WARN`, and `NOTE`), which default to the server message color if not specified.  Standard replies all all shown in the server buffer (triggering the unread indication).


casperstorm (approved):
Clean. Haven't tested it but code looks good 👍🏻

andymandias:
Error color makes sense to me for `FAIL`.  I'm not sure if text-secondary will attract attention, at least with the default Ferra theme, but that might just be me.  Any preference for the fallback `WARN` coloring, @casperstorm?

tarkah (approved):
Clean! LGTM

Issue #733, opened 2025-01-25
Ability to exclude specific buffers from notifications.


englut:
Someone suggested in the #halloy channel.

```
 > So, can I somehow exclude sound notifications only for some channels?
```

I think the ability to exclude notifications for specific buffers make sense to prevent abuse or to mute people/channels if you so desired.

List of notification types to add exclusion to:

- `direct_message`
- `file_transfer_request`
- `highlight`

I was thinking of using the keyword `mute` but I see that `exclude` is an existing convention used. 
Example config of excluding along with mixed buffer types:
```
[notifications]
highlight = {
  sound = "dong",
  exclude = ["HalloyUser123", "#halloy"]
}
direct_message = { 
  sound = "peck",
  show_toast = true,
  exclude = ["HalloyUser123"]
}
```

As I'm already a contributor for notifications, I can add this feature. But if anyone else wants to, it'd be a good first issue to tackle! Just let me know here.  😃

casperstorm:
Yeah we could follow the include, exclude pattern!
Feel free to attack this @englut.

DownrightNifty:
My use case for this would be turning off notifications from SaslServ every time I reconnect.

deepspaceaxolotl:
I'm glad this is implemented! Relatedly though, would it be possible to have a way to hide muted channels, or at least a config that would prevent them from showing an unread message indicator? I get messaged by a bot every time I connect to a channel and it's rather annoying TBH, would be great to be able to hide that buffer entirely or at least be able to ignore it and focus on buffers that actually have new messages rather than automated ones!

andymandias:
> I'm glad this is implemented! Relatedly though, would it be possible to have a way to hide muted channels, or at least a config that would prevent them from showing an unread message indicator? I get messaged by a bot every time I connect to a channel and it's rather annoying TBH, would be great to be able to hide that buffer entirely or at least be able to ignore it and focus on buffers that actually have new messages rather than automated ones!

Sounds like that could be a useful feature for some.  I suggest you open a new feature request issue, as it will likely get lost/forgotten if it's only in a comment on a closed issue.

PR #741, opened 2025-01-30
Optionally exclude `direct_message`, `file_transfer_request` and `highlight` notifications.

englut:
Addresses #733.

Example config:
```
[notifications.highlight]
show_toast = true
sound = "dong"
exclude = ["ChanServ", "NickServ", "#halloy"]
```

Also, added the `include` property with support for wildcards: `*` and `all`.

### DIFF
--- a/data/src/config/notification.rs
+++ b/data/src/config/notification.rs
@@ -10,6 +10,7 @@ pub struct Notification<T = String> {
     pub show_toast: bool,
     pub sound: Option<T>,
     pub delay: Option<u64>,
+    pub exclude: Option<Vec<String>>
 }
 
 impl<T> Default for Notification<T> {
@@ -18,6 +19,7 @@ impl<T> Default for Notification<T> {
             show_toast: false,
             sound: None,
             delay: Some(500),
+            exclude: None
         }
     }
 }
@@ -64,6 +66,7 @@ impl Notifications {
                 show_toast: notification.show_toast,
                 sound: notification.sound.as_deref().map(Sound::load).transpose()?,
                 delay: notification.delay,
+                exclude: notification.exclude.to_owned(),
             })
         };
 
--- a/data/src/config/notification.rs
+++ b/data/src/config/notification.rs
@@ -11,7 +11,9 @@ pub struct Notification<T = String> {
     pub sound: Option<T>,
     pub delay: Option<u64>,
     #[serde(default)]
-    pub exclude: Vec<String>
+    pub exclude: Vec<String>,
+    #[serde(default)]
+    pub include: Vec<String>,
 }
 
 impl<T> Default for Notification<T> {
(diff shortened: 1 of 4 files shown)


src/notification.rs
```diff
@@ -89,28 +89,48 @@ impl Notifications {
...
-            }
             Notification::Highlight {
                 enabled,
                 user,
                 target,
             } => {
-                if *enabled {
+                if !config.highlight.exclude.as_ref().map_or(false, |exclude| {
+                    exclude.contains(&target.to_string())
+                        || exclude.contains(&user.nickname().to_string())
```
englut, line 131: There's a weird scenario where.. if your nick is added to `exclude` for highlights, you'll not get any sort of highlight notifications for channels or dms. Not sure if that's acceptable or not. Could be a feature :thinking:  What do y'all think?

src/notification.rs
```diff
@@ -89,28 +89,46 @@ impl Notifications {
...
-            }
             Notification::Highlight {
                 enabled,
                 user,
                 target,
             } => {
-                if *enabled {
+                if !config.highlight.exclude.as_ref().is_some_and(|exclude| {
+                    exclude.contains(&target.to_string())
+                        || exclude.contains(&user.nickname().to_string())
```
englut, line 129: There's a weird scenario where.. if your nick is added to exclude for highlights, you'll not get any sort of highlight notifications for channels or dms. Not sure if that's acceptable or not. Could be a feature 🤔 What do y'all think?

englut: Actually, I just tested this some more. If you exclude a nick in `highlight`, you end muting their highlight notifications in both DMs and channels. This feels like it's the desirable functionality to me.

data/src/config/notification.rs
```diff
@@ -10,6 +10,7 @@ pub struct Notification<T = String> {
     pub show_toast: bool,
     pub sound: Option<T>,
     pub delay: Option<u64>,
+    pub exclude: Option<Vec<String>>
```
tarkah, line 13: We can remove Option and add serde default attribute. The option is redundant here and means the same as an empty vec.

This will clean up your code paths below to remove the is_some_and calls

src/notification.rs
tarkah: Makes sense to me! Alternatively we have separate config arrays for channels vs nicks, but I don't have a preference. @casperstorm thoughts?

tarkah (commented):
Thanks for this! Just one minor nit, otherwise looks great :D

data/src/config/notification.rs
englut: Ahhh that's a great call out. Thank you @tarkah I've pushed up the changes :pray:

src/notification.rs
casperstorm: Make sense to me as well.

casperstorm:
> Makes sense to me! Alternatively we have separate config arrays for channels vs nicks, but I don't have a preference. @casperstorm thoughts?

I think this is fine. Let's see if we should split it eventually.

casperstorm:
Should we have `include` as well? Would that make sense to keep it aligned with others who have exclude/include?
Like, if you want to exclude everything, but #halloy channel? similar to what we have for server messages https://halloy.squidowl.org/configuration/buffer/server_messages/join.html#include

englut:
> Should we have `include` as well? Would that make sense to keep it aligned with others who have exclude/include?
> 
> Like, if you want to exclude everything, but #halloy channel? similar to what we have for server messages https://halloy.squidowl.org/configuration/buffer/server_messages/join.html#include
> 
> 
> 
> 

Oh interesting. I like that and it'll be consistent with the other `exclude` usage. I'll add `include` and make sure `*` in `exclude` works too!

englut:
Updated with `include` support, with priority over exclude, along with wildcard support for `*` and `all`.

PR #751, opened 2025-02-03
Fix window shrink

casperstorm:


casperstorm:
@tarkah can you test this on Wayland when you have time?

tarkah:
> @tarkah can you test this on Wayland when you have time?

I don't run Wayland so I can't test that, but I can confirm that I don't see any shrinking on X11

andymandias:
Can't reproduce the window shrink on Fedora Asahi Remix 40 GNOME 46 (Wayland) unfortunately.  No window size issues seen on `fix-window-shrink` (or `main`).

casperstorm:
Thats good news, then it means not all is affected. I still think we should get this PR added as its abit more robust than previous.
You two agree? @tarkah @andymandias

andymandias:
I agree. Will hopefully address the issue for those who encounter it, and it looks like no ill effects for those that don't.

tarkah (approved):

PR #752, opened 2025-02-04
Emoji Picker

andymandias:
In-app method for entering emoji via picker that shows matching emoji when typing `:shortcode` (all [GitHub shortcodes](https://github.com/ikatyang/emoji-cheat-sheet/blob/master/README.md) should be supported).

### DIFF
--- a/src/buffer/input_view/completion.rs
+++ b/src/buffer/input_view/completion.rs
@@ -1,5 +1,6 @@
 use std::collections::HashMap;
 use std::fmt;
+use strsim::jaro_winkler;
 
 use data::buffer::SortDirection;
 use data::user::User;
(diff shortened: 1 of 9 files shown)


src/buffer/input_view/completion.rs
```diff
@@ -118,6 +146,15 @@ impl Entry {
 
                 new_input
             }
+            Entry::Emoji(emoji) => {
+                let mut words: Vec<_> = input.split_whitespace().collect();
+
+                let last_word = words.last_mut().unwrap();
```
casperstorm, line 152: Let's ensure we don't use unwrap here.

casperstorm:
This works really well! Good job, @andymandias.

1. If we want skintone i think we should simply add it as a configuration option. Set and forget.
2. Would be cool to have the mapping the other way, if someone did `:thinking:` that we map it to a emoji, also behind a configuration option.

Im not sure we need either of above implemented before we merge this tho, could easily be added later.

casperstorm (approved):
Looks good to me! @tarkah?

tarkah (approved):
This works great, very nice!

tarkah:
Probably an iced / font issue: certain emojis show up as "text" instead of colored / rendered. But that's just an observation that might need to be fixed, nothing to do w/ this PR.

tarkah:
@andymandias can you fix the conflict? Then feel free to merge

PR #754, opened 2025-02-05
Default Value for `infinite_scroll`

andymandias:
The documentation lists the default value as `true`, but the default being set is `false`, this is just to make them consistent.  I decided to go with `true`, because I suspect that is the more modern expectation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+Fixed:
+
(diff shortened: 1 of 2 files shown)


casperstorm (approved):

casperstorm:
Perhaps add a entry to CHANGED in Changelog. Not sure if needed ill let you decide that.
Otherwise looks good 👍🏻

tarkah (approved):

PR #755, opened 2025-02-05
History Compatibility

andymandias:
Trying to fix history compatibility issues between versions 2024.14 and the 2025.1+.  This version is working between itself and 2024.14, but I believe it's incompatible with 2025.1 history.

### DIFF
--- a/data/src/message.rs
+++ b/data/src/message.rs
@@ -61,8 +61,6 @@ static URL_REGEX: Lazy<Regex> = Lazy::new(|| {
     .unwrap()
 });
 
-pub type Channel = String;
-
 pub(crate) mod broadcast;
 pub mod formatting;
 pub mod source;
@@ -107,7 +105,7 @@ impl From<Encoded> for proto::Message {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub enum Target {
     Server {
         source: Source,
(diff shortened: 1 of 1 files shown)


tarkah:
Damn, I'm super bummed we missed this :(

data/src/message.rs
```diff
@@ -107,7 +105,7 @@ impl From<Encoded> for proto::Message {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq)]
```
tarkah, line 108: My understanding is the only breakage is we renamed `Query.nick` to `Query.query`. We can use `#[serde(alias)]` to provide upgrade compatability from < 2025.1 to >= 2025.1

tarkah: Shit, `message::Source::Action` is also breaking as we added a payload to it :/

tarkah: We may want to do a bigger, more comprehensive update that can accomplish the following:

- Safely upgrades from 2025.1
- Is more easily upgradable in the future without breakage (no unit variants, only struct variants so  we can always add fields)

We need to separate the serialization format from the data format we use at runtime. Having them be the same makes it way too easy for breakage to occur. Having them separate and a simple `From / Into` conversion makes it super easy to see what needs to change in a _compatible_ way w/ the serialization format, when the data format changes.

andymandias:
Closing this as the scope of the problem is larger than previously realized.  A different approach will be necessary.

PR #756, opened 2025-02-07
Minor Documentation Fixes

andymandias:
A couple minor documentation fixes (reported by a user in #halloy).

### DIFF
--- a/book/src/configuration/buffer.md
+++ b/book/src/configuration/buffer.md
@@ -348,7 +348,7 @@ Horizontal alignment of nicknames.
 
 ```toml
 # Type: string
-# Values: "left", "right", "top"
+# Values: "left", "right"
 # Default: "left"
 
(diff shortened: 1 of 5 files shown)


book/src/configuration/buffer.md
```diff
@@ -348,7 +348,7 @@ Horizontal alignment of nicknames.
 
 ```toml
 # Type: string
-# Values: "left", "right", "top"
```
casperstorm, line 351: Why remove `top`?

andymandias: Because I thought I had tested it and it didn't work, but I must've changed the other horizontal alignment of nicknames values in my test.  Thank you for catching that!

casperstorm: All good :D

casperstorm (approved):

PR #757, opened 2025-02-08
Completion Tweaks

andymandias:
This PR has three ~related changes to completion (which can be split further if needed):
  - Fixes a bug where text completion will not replace the previous completion properly when the specified suffix contains a space (i.e. during typical usage).  This bug was introduced as part of a previous minor bug fix, where whitespace in a message would be compressed (e.g. multiple spaces replaced with a single space) when using text/emoji completion.
  - When cycling through text completions, the last name would require two tab presses to loop back to the first completion from the last completion.  This fixes that behavior by cycling back to the original prompt before looping back to the first completion (with one tab between each).
  - Select the first command/emoji automatically, when using the command/emoji picker.  This was suggested by @englut, and I like how it works;  I think this is more in line with how other messengers work, and will probably be more intuitive to users.  The only drawback I can see is that it will prevent sending "uncompleted emoji" without adding a space afterward.  Maybe we could enable Ctrl-Enter and Command-Enter to bypass the command/emoji picker and send the message?  And/or, we could allow tab to cycle to "no command/emoji" selected as part of its loop.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -435,6 +435,17 @@ impl Buffer {
             Buffer::Highlights(highlights) => Some(highlights.scroll_view.is_scrolled_to_bottom()),
         }
     }
+
+    pub fn close_picker(&mut self) -> bool {
+        match self {
+            Buffer::Empty | Buffer::FileTransfers(_) | Buffer::Logs(_) | Buffer::Highlights(_) => {
+                false
+            }
+            Buffer::Server(state) => state.input_view.close_picker(),
+            Buffer::Channel(state) => state.input_view.close_picker(),
+            Buffer::Query(state) => state.input_view.close_picker(),
+        }
+    }
 }
 
 impl From<data::Buffer> for Buffer {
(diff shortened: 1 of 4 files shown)


englut:
This works great! Thank you for addressing the feedback so quickly <3

> The only drawback I can see is that it will prevent sending "uncompleted emoji" without adding a space afterward. 

I think that actually makes sense behaviourally. If you intended to type `:smile` (and not an emoji) and adding a space after would be a good way to end the picker. But.. I guess if you don't want a space that might make sense. maybe letting the user close out the picker if they hit `esc`?

casperstorm:
This make sense to me!

andymandias:
Having thought it over I think being able to tab into a no-command/emoji-picked state is probably not good UX.  At some point it might be nice to also allow ctrl/cmd-enter to submit a send, but for now I think being able to dismiss the picker with escape is enough.

casperstorm (approved):

PR #758, opened 2025-02-09
Add firewall exception to Windows during installation

casperstorm:
This adds two inbound rules to Windows Defender which ensures we can connect through the firewall.

image

### DIFF
--- a/scripts/build-windows-installer.sh
+++ b/scripts/build-windows-installer.sh
@@ -8,8 +8,9 @@ scripts/build-windows.sh
 # install latest wix
 dotnet tool install --global wix
 
-# add required wix extension
+# add required wix extensions
 wix extension add WixToolset.UI.wixext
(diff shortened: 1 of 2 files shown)


KaiKorla:
TBH, i wouldn't implement this as default. I think this is only necessary for DCC and it requires more setup steps in NAT environments aka port forwarding and so on. NAT is widely the default for home users, as they are in most cases behind a router. If a user doesn't know how to do it properly, it opens the door to attackers. In additions, ports under 1024 (well-known ports) and registered ports should be used for this.

IMO, instead i recommending to write a documentation for DCC (behind NAT) or refer to public available howtos.

casperstorm:
I agree, @KaiKorla. This was an attempt to fix the issue which i initial thought was firewall caused.
Turned out, it wasn't: https://github.com/squidowl/halloy/pull/776.

I'll close this.

PR #765, opened 2025-02-11
Message Compatibility Tests

andymandias:
Adds some forward and backward compatibility tests with release version 2025.1 for (de)serializing messages.

### DIFF
--- a/data/Cargo.toml
+++ b/data/Cargo.toml
@@ -43,6 +43,8 @@ anyhow = "1.0.91"
 image = "0.24.9"
 html-escape = "0.2.13"
 emojis = "0.6.4"
+irc_2025_1 = { git = "https://github.com/squidowl/halloy.git", tag = "2025.1", package = "irc" }
+data_2025_1 = { git = "https://github.com/squidowl/halloy.git", tag = "2025.1", package = "data" }
 
 [dependencies.irc]
 path = "../irc"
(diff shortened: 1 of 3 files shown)


data/Cargo.toml
```diff
@@ -43,6 +43,8 @@ anyhow = "1.0.91"
 image = "0.24.9"
 html-escape = "0.2.13"
 emojis = "0.6.4"
+irc_2025_1 = { git = "https://github.com/squidowl/halloy.git", tag = "2025.1", package = "irc" }
+data_2025_1 = { git = "https://github.com/squidowl/halloy.git", tag = "2025.1", package = "data" }
```
tarkah, line 48: This is mega cursed, we will link to all past versions this way, which will grow linearly with each new version.

I'd rather see static json test fixtures for all known existing variants and just attempt to Deserialize it. If it fails, we have a breakage.

We will obviously need to update these tests as we add new forward compatible fields

andymandias: It is indeed mega-cursed!  I think I mislabeled my tests, so for clarity I'm going to avoid using the terms forward/backward compatibility for the moment.  If we have static json test fixtures for all known existing variants we can ensure the current version can read all older serialized messages, but we can't ensure old versions can read the current version's serialized messages.  I was operating under the assumption that we wanted tests to ensure users could downgrade as well.

Normally I'd try and build these tests into an exterior system, but I wasn't sure which would be appropriate here.

PR #771, opened 2025-02-13
Ability to hide image for card preview

casperstorm:
- Added an option to show or hide images in preview cards by default
- Adjusted colors slightly

image

### DIFF
--- a/src/buffer/scroll_view.rs
+++ b/src/buffer/scroll_view.rs
@@ -991,18 +991,20 @@ fn preview_row<'a>(
                 keyed::Key::Preview(message.hash, idx),
                 button(
                     container(
-                        column![
-                            column![text(title)
+                        column![column![text(title)
+                            .shaping(text::Shaping::Advanced)
+                            .style(theme::text::primary)]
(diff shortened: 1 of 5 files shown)


src/buffer/scroll_view.rs
```diff
@@ -991,18 +991,20 @@ fn preview_row<'a>(
...
-                                .style(theme::text::url)]
-                            .push_maybe(description.as_ref().map(|description| {
-                                text(description)
-                                    .shaping(text::Shaping::Advanced)
-                                    .style(theme::text::primary)
-                            })),
-                            container(image(path).content_fit(ContentFit::ScaleDown))
-                                .max_height(200)
-                        ]
+                                .style(theme::text::secondary)
```
tarkah, line 1000: Is the text style change here intended?

From url / primary for title / description to primary / secondary

tarkah: Nvm I see it is

casperstorm: Yep, it is intended. Looks better.

tarkah (approved):

PR #775, opened 2025-02-14
Add rose-pine theme

mikemykhaylov:
Really like this palette, could be a good addition


casperstorm (approved):
Love that theme myself. Thanks for the contribution.

PR #776, opened 2025-02-15
Ignore errors while adding der certificates

### LINKED ISSUE #678: Can connect with URL links, but not via config (Windows 11, Halloy v2024.14)
Hello, 

I just downloaded Halloy and attempted to use it. Clicking an IRC URL (ex. irc://irc.snoonet.org/reddit) will connect me. I'm unable to connect to Librachat's #halloy channel, as it doesn't seem to exist if I search for it via Librachat's own site. Hence, I'm making an issue here. I've changed the channel in my config to #linux, due to this, but I'm still unable to make a connection.  If I use the default config that comes with Halloy, with the only change being my username; it will not connect. It simply shows `connecting to server...` over and over. As a note, the config is as follows:
```toml
# Halloy config.
#
# For a complete list of available options,
# please visit https://halloy.squidowl.org/configuration/index.html

[servers.liberachat]
nickname = "Alcatraz"
server = "irc.libera.chat"
channels = ["#halloy"]
```
Logs:
```
01:00 INFO  RouteRecived: "ircs://irc.libera.chat/#halloy"
01:00 DEBUG loaded metadata for server on libera
01:00 DEBUG loaded history for server on libera: 0 messages
01:00 DEBUG flushed history for server on libera
01:01 INFO  dashboard saved
01:01 INFO  RouteRecived: "ircs://irc.libera.chat/"
01:01 DEBUG closed history for server on libera
01:01 DEBUG loaded history for server on libera: 1 messages
01:01 INFO  dashboard saved
```

This is on Windows 11, using version Halloy 2024.14 that was installed via the `winget` method as noted in the documentation. I thought perhaps it was just a Windows firewall issue, so I've gone ahead and made explicit rules to allow Halloy to connect. No dice. 

It's possible that this is user error, however, considering I'm using the default configuration I don't think that's the case. I also don't think it's a Windows or ISP issue, since clicking IRC URLs works without issue. Please let me know what other information I can provide to help resolve this, as it seems odd to me that I'm running into this issue.

casperstorm:
Fixes #678.

The problem in #678 seems to be that Windows (10 specifically) can have some root certificates which rustls does not understand.
This causes a panic, since we assume (unwrap) a result where in fact we have an error:

> InvalidCertificate(Other(OtherError(UnsupportedCriticalExtension)))

Thanks to @impliedgg for tracking this down.

### DIFF
--- a/irc/src/connection/tls.rs
+++ b/irc/src/connection/tls.rs
@@ -30,7 +30,7 @@ pub async fn connect<'a>(
         let mut roots = rustls::RootCertStore::empty();
 
         for cert in rustls_native_certs::load_native_certs()? {
-            roots.add(cert).unwrap();
+            let _ = roots.add(cert);
         }
 
         if let Some(cert_path) = root_cert_path {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Fixed:
 - [`infinite_scroll`](https://halloy.squidowl.org/configuration/buffer.html?highlight=infinite#infinite_scroll) was defaulting to `false`, contrary to its documented default value.  Now defaults to `true`
 - A rare issue where the app opens smaller than when closed
 - Closing the application while it is minimized (on Windows) causes it to open with the wrong size and position next time
+- TLS connection issue on Windows (10+)
 
 # 2025.1 (2025-02-02)
 


irc/src/connection/tls.rs
```diff
@@ -30,7 +30,7 @@ pub async fn connect<'a>(
         let mut roots = rustls::RootCertStore::empty();
 
         for cert in rustls_native_certs::load_native_certs()? {
-            roots.add(cert).unwrap();
+            let _ = roots.add(cert);
```
casperstorm, line 33: I don't think it's necessary to log the failed DER certs, so for the first proposal, I’ll just ignore the result instead.

tarkah (approved):

Issue #446, opened 2024-07-29
Improvement request: text input should always have focus


smakkerlak:
Currently, clicking a channel in halloy while another application has focus, or clicking a different channel than the currently focused one, automatically makes the text input active so you can type.

I keep clicking something that moves focus away from the text input and start typing into nothing. Send help.

Basically, I'd love it if any window with a text input would always have its focus be the text input, no matter what the user does. This includes things like clicking empty areas in nicklist or channel bar. I cannot currently think of a situation where you would want focus to not be the text input, expect perhaps if there are multiple text input fields and one was clicked explicitly (not sure if any exists)

Running Win11, built from latest main

dmd:
Yes. If I am in another window, and I click ANYWHERE in halloy, the text input most closely associated with where I click should have focus. There is no situation permitted where hallow should be the frontmost app but no text input has focus.

voidlily:
Agreed, I'm coming at this from muscle memory in hexchat where no matter where you clicked in the window, the text input would always have focus, and the current behavior is throwing me off a lot

Issue #784, opened 2025-02-20
Have halloy remember which pane was active in the layout


kalebris:
currently when I start halloy the top left pane is active, regardless of which one was active previously. Halloy is very nice in remembering the layout, but it doesn't seem to remember which pane was active when I quit.

Would it be possible to add this functionality?

PR #787, opened 2025-02-20
Always Focus Input

andymandias:
This is an attempt to address #446.  Currently needs to have the following resolved before before it could be merged:

- [x] Utilizes an addition to Iced I would need to get merged before this could be merged.
- [x] In many cases it maintains focus by immediately re-focusing the appropriate input when focus is taken away from it.  Ideally focus would not be taken away in the first place. (edit: I wasn't able to find a reasonable implementation for the ideal approach at the moment, so I'm marking this as "completed".)


andymandias:
I haven't seen a straightforward way to avoid utilizing immediate re-focus, rather than prevent loss of focus entirely, so I think this is about as good as I can think to make it at the moment.  In my testing I find it works well, but I have only done so on Fedora under Wayland.

Should more-or-less address #446 and #784, with some room for improvement in the future.

casperstorm:
So strange that this is not building...

andymandias:
Looks like it fixed itself in the meantime!  Maybe the commit was too new? :shrug:

casperstorm (approved):
Works great, i don't have anything. @tarkah?

src/widget/left_click.rs
```diff
@@ -0,0 +1,49 @@
...
+use iced::advanced::{mouse, Clipboard, Layout, Shell};
+use iced::event;
+
+use crate::widget::{decorate, Renderer};
+use crate::Element;
+
+pub fn left_click<'a, Message>(
+    content: impl Into<Element<'a, Message>>,
+    message: Message,
+) -> Element<'a, Message>
```
tarkah, line 11: We don't need this over `button` w/ bare styling?

src/screen/dashboard.rs
```diff
@@ -1126,12 +1095,13 @@ impl Dashboard {
...
                         theme,
-                        main_window,
+                        window != self.main_window(),
                     )
                 })
                 .spacing(4)
                 .on_click(pane::Message::PaneClicked),
-            )
+                pane::Message::MaintainFocus,
+            ))
```
tarkah, line 1104: We already have `on_click(pane::Message::PaneClicked)`, why do we need this other click / event? Just use `pane::Message::PaneClicked` to refocus the text input

src/screen/dashboard.rs
```diff
@@ -1659,60 +1617,90 @@ impl Dashboard {
...
-        &mut self,
-        main_window: &Window,
-        window: window::Id,
-        pane: pane_grid::Pane,
-    ) -> Task<Message> {
-        if self.focus != Some((window, pane)) {
-            self.focus = Some((window, pane));
-
-            if let Some(task) = self.panes.iter(main_window.id).find_map(|(w, p, state)| {
+    pub fn focus_focused_pane_buffer(&mut self) -> Task<Message> {
```
tarkah, line 1647: Nit: Naming is confusing, maybe `refocus_pane`?

src/buffer/input_view.rs
```diff
@@ -308,6 +309,15 @@ impl State {
...
             Message::Escape => (Task::none(), None),
+            Message::Focus(is_focused) => {
+                let task = if is_focused {
+                    Task::none()
+                } else {
+                    text_input::focus(self.input_id.clone())
+                };
+
+                (task, None)
+            }
```
tarkah, line 320: Why do we care if it's already focused? Focusing something that's already focused is a no-op.

src/main.rs
```diff
@@ -486,7 +486,13 @@ impl Halloy {
...
                             .broadcast(&server, &self.config, sent_time, Broadcast::Reconnected)
                             .map(Message::Dashboard)
-                    }
+                    };
+
+                    let focus_focused_pane_buffer = dashboard
+                        .focus_focused_pane_buffer()
+                        .map(Message::Dashboard);
+
+                    Task::batch(vec![broadcast, focus_focused_pane_buffer])
```
tarkah, line 495: Why refocus here? Is focus taken away somehow or are we using this as the initial thing to focus the pane? Can we "refocus" instead when first entering dashboard screen?

src/screen/dashboard.rs
```diff
@@ -1659,60 +1617,90 @@ impl Dashboard {
...
+    pub fn focus_first_pane(&mut self, window: window::Id) -> Task<Message> {
+        let pane = self
+            .panes
+            .iter()
+            .find_map(|(w, pane, _)| (w == window).then_some(pane));
+
+        pane.map_or(Task::none(), |pane| self.focus_pane(window, pane))
+    }
+
+    pub fn focus_last_focused_or_first_pane(&mut self, window: window::Id) -> Task<Message> {
```
tarkah, line 1692: Nit: Maybe `focus_window`? Since we seem to call this only when window focus changes.

src/screen/dashboard.rs
```diff
@@ -1750,65 +1733,59 @@ impl Dashboard {
...
-    ) -> Task<Message> {
+    fn close_pane(&mut self, window: window::Id, pane: pane_grid::Pane) -> Task<Message> {
         self.last_changed = Some(Instant::now());
 
-        if window == main_window.id {
+        if window == self.main_window() {
+            self.focus_history = self
+                .focus_history
+                .clone()
+                .into_iter()
```
tarkah, line 1781: `iter().filter().cloned()` would technically be more efficient as you only clone the retained values.

src/screen/dashboard.rs
```diff
@@ -1750,65 +1733,59 @@ impl Dashboard {
...
-        main_window: &Window,
-        window: window::Id,
-        pane: pane_grid::Pane,
-    ) -> Task<Message> {
+    fn close_pane(&mut self, window: window::Id, pane: pane_grid::Pane) -> Task<Message> {
         self.last_changed = Some(Instant::now());
 
-        if window == main_window.id {
+        if window == self.main_window() {
+            self.focus_history = self
```
tarkah, line 1778: We have two `if window == self.main_window()` guards here that can be combined

src/buffer/input_view.rs
```diff
@@ -326,7 +336,7 @@ impl State {
     }
 
     pub fn focus(&self) -> Task<Message> {
-        text_input::focus(self.input_id.clone())
+        text_input::is_focused(self.input_id.clone()).map(Message::Focus)
```
tarkah, line 339: You don't need `Message::Focus`, you can `Task::then` to chain `text_input::focus` after this using the conditional from the previous operation

src/buffer/input_view.rs
tarkah: Discussed on IRC, this is to avoid cursor jump 👍

src/screen/dashboard/pane.rs
```diff
@@ -22,6 +21,7 @@ pub enum Message {
     Popout,
     Merge,
     ScrollToBottom,
+    MaintainFocus,
```
tarkah, line 24: I think `MaintainFocus` can be removed from here and `sidebar` and we can instead expose `LeftClick` from `event.rs` which is always emitted (even when event is captured) and use that to always call `refocus` in dashboard. 

This is logically simple to think about. Any time the user left clicks within the application, doesn't matter where, we refocus our text input. Clicking doesn't happen and isn't a high volume event so this isn't a big deal and then we don't have to figure out all places someone can click to lose focus from text input.

This actually probably allows us to remove almost most code in this PR that isn't focus history (which we still need obviously). Everything else should be captured by the above.

tarkah: The only time we wouldn't refocus is if `self.focus` is None if we unfocused the pane entirely

tarkah: If we do the above, we should be fully covered w/ input focus via left click anywehre when pane has focus, window focus changing, and clicking on a pane from unfocused state.

src/event.rs
```diff
@@ -37,6 +38,9 @@ fn filtered_events(
             key: keyboard::Key::Named(keyboard::key::Named::End),
             ..
         }) if ignored(status) => Some(Event::End),
+        iced::Event::Mouse(mouse::Event::ButtonPressed(mouse::Button::Left)) if ignored(status) => {
```
tarkah, line 41: Do we want to include captured statuses? It's technically possible a button click is captured and then text input still processes that event to defocus itself I think.

src/screen/dashboard/sidebar.rs
```diff
@@ -576,7 +564,7 @@ fn upstream_buffer_button(
                             BufferFocusedAction::ClosePane => Some(Message::Close(window, pane)),
                         }
                     } else {
-                        None
+                        Some(Message::Focus(window, pane))
```
tarkah, line 567: Is this still needed now that we handle `Event::LeftClick`?

tarkah (commented):
Changes look really good! I haven't had a chance to test yet, I had to run out, but I'll test either later or tomorrow. 

That Event::LeftClick really cleaned this MR up and I love all the removal of `main_window` everywhere.

src/screen/dashboard/sidebar.rs
andymandias: I think you're right, and after reverting it to `None` I wasn't able to get any sidebar buttons to defocus the pane.

src/event.rs
andymandias: I wasn't able to produce a defocus via button click in my testing, but I may just not know the prerequisite(s) to do so.

src/widget/left_click.rs
casperstorm: We have a `widget::transparent_button` now.

tarkah (approved):
I've added the new shortcuts and it's working great! 

@andymandias please merge after you've tested

PR #792, opened 2025-02-22
Default dcc save path & and show remaining time

### LINKED ISSUE #615: Set Default XDCC Settings
XDCC is cumbersome using Halloy because you have to manually accept each transaction and set the download location. Can be a real issue when trying to XDCC hundreds of files.

You should be able to set a default download location and trusted senders (whitelist) which automatically accept file transfers.

using [file_transfer](https://halloy.squidowl.org/configuration/file_transfer/index.html#file_transfer) you can explicitly set [save_directory](https://halloy.squidowl.org/configuration/file_transfer/index.html#save_directory) but it only sets a default directory for the file save window to open to... It doesn't actually set a default location for all income XDCC files because you have to __manually__ click "save" for each file.

### LINKED ISSUE #738: Show remaining time on transfers
It would be great to see the remaining time for file transfers in the transfer window.

casperstorm:
I've changed the `file_transfer.save_directory` to now work as the default download directory. That means if you set it, and press approve on a transfer it will start instantly and save the file to the above mentioned directory (fixes #615).

I've also added time remaining to easier see when a transfer is done. (fixes #738).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Unreleased
 
+Added:
+
+- Time remaining is now displayed during file transfers
+
+Changed:
+
+- [file_transfer.save_directory] is now default download path for transfers. If set, files will be downloaded there by default. Otherwise, you'll be prompted to choose a location.
+  
 # 2025.2 (2025-02-20)
 
 Added:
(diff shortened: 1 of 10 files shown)


tarkah (approved):

tarkah:
Nice, I like this!

Issue #647, opened 2024-11-10
Feature request: Configuration option for changing default pane stacking


i8x3:
Hello!

Currently, there is no way to change the default the way panes stack in the configuration. In `src/screen/dashboard.rs`, I saw that there was already a comment that said "Default split could be a config option" (line 1239).
```
// Default split could be a config option.
let axis = pane_grid::Axis::Horizontal;
```
I modified the line right under it, changing `Horizontal` to `Vertical`, built it, and it worked just like I hoped for. If you guys could make this an option in the configuration, that would be great.

Thank you very much!

casperstorm:
Fixed this in #793 - it will be released with 2025.3 :)

PR #793, opened 2025-02-22
Control default pane split direction

casperstorm:
Fixes https://github.com/squidowl/halloy/issues/647.
Introduced a new `[pane]` config which intention is to control global pane settings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+Added:
+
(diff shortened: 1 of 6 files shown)


andymandias (approved):
Works perfectly in my testing.

"Default axis used when splitting a pane." is a correct description, but I think it could be misunderstood if not read carefully.  Maybe, "Default axis used when splitting a pane (i.e. default orientation of the divider between panes)." would help?  Not necessary though.

PR #800, opened 2025-02-28
`/msg target` to Open Query/Channel

andymandias:
Allows `/msg target` (without any message text following) in the input to open the target (query or channel) in a new pane.  This behavior matches the description given for `/msg` in `src/buffer/input_view/completion.rs`, and allows for the user to open a query without sending a message or right-clicking the target user's name and selecting "Message".

### DIFF
--- a/data/src/command.rs
+++ b/data/src/command.rs
@@ -6,8 +6,38 @@ use itertools::Itertools;
 
 use crate::{buffer, ctcp, message::formatting};
 
+#[derive(Debug, Clone)]
+pub enum Command {
+    Internal(Internal),
+    Irc(Irc),
+}
+
+#[derive(Debug, Clone)]
+pub enum Internal {
+    OpenBuffer(String),
+}
+
+#[derive(Debug, Clone)]
+pub enum Irc {
+    Join(String, Option<String>),
+    Motd(Option<String>),
+    Nick(String),
+    Quit(Option<String>),
+    Msg(String, String),
+    Me(String, String),
+    Whois(Option<String>, String),
+    Part(String, Option<String>),
+    Topic(String, Option<String>),
+    Kick(String, String, Option<String>),
+    Mode(String, Option<String>, Option<Vec<String>>),
+    Away(Option<String>),
+    SetName(String),
+    Raw(String),
+    Unknown(String, Vec<String>),
+}
+
 #[derive(Debug, Clone, Copy)]
-pub enum Kind {
+enum Kind {
     Join,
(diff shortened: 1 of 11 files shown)


casperstorm (approved):

data/src/command.rs
```diff
@@ -65,10 +65,34 @@ pub enum Command {
...
+            Command::Kick(_, _, _) => false,
+            Command::Mode(_, _, _) => false,
+            Command::Away(_) => false,
+            Command::SetName(_) => false,
+            Command::OpenBuffer(_) => true,
+            Command::Raw(_) => false,
+            Command::Unknown(_, _) => false,
+        }
+    }
+}
```
tarkah, line 94: Let's have `Command::Internal` and `command::Internal::OpenBuffer`. New variants can be added there and its type enforced better.

Then `Input::internal` can return `Option<command::Internal>` and caller doesn't have to deal w/ impossible states of dealing w/ non-internal commands during pattern matching

src/buffer/input_view.rs
```diff
@@ -225,7 +234,22 @@ impl State {
...
+                        (
+                            Task::none(),
+                            Some(Event::OpenBuffer {
+                                target: Target::parse(
+                                    target.as_ref(),
+                                    clients.get_chantypes(buffer.server()),
+                                    clients.get_statusmsg(buffer.server()),
+                                    clients.get_casemapping(buffer.server()),
+                                ),
+                                history_task,
```
tarkah, line 247: Is history task needed here? It's only used for recording input which isn't relevant for `OpenBuffer`

tarkah:
I don't think internal commands should hit `input` API at all, it's leaking everywhere and we can handle it more gracefully if we know prior to returning an `Input` that it's a `command::Internal`. Let me see if I can cook up something real quick

src/buffer/input_view.rs
andymandias: `record_input` saves to the input history (which also clears the input).  I was using it for that purpose, hadn't fully considered the relevant part of `record_input` would be better just being called directly.

tarkah (approved):
@andymandias lmk how you like the refactor I did. I think it simplifies things a ton

tarkah (approved):
LGTM! @andymandias feel free to merge once you've reviewed my latest changes

andymandias:
Tested, everything is working as intended.  Thanks for your help!

andymandias:
Thanks for the note!  I'll take a look at that.

Issue #801, opened 2025-02-28
action upon double click in nicklist?


0ther0ne:
something like:

```
[buffer.channel.nicklist]
double_click = "query"

```

to change behavior from _mention_ the selected user to _open a query_ with selected user

joehonkey:
Are you specifically talking about when you 'click' a username to open a query rather than clicking a username and it goes to the 'send message' buffer?

0ther0ne:
@joehonkey exactly

casperstorm:
I'll grab this.

casperstorm:
This is done now and will be released along with the next public version.
The configuration looks like this:

```toml
[buffer.nickname]
click = { interaction = "single-click", action = "insert-nickname" }

[buffer.channel.nicklist]
click = { interaction = "double-click", action = "open-query" }
```

0ther0ne:
wow, so fast!
thanks!

PR #803, opened 2025-02-28
clicking a nickname is now configurable

casperstorm:
Ability to add a `click` config to control what action should happen when you either click or double click on a nickname.
You can set it on both `[buffer.nickname]` and `[buffer.channel.nicklist]`.

Config looks like:

```toml
[buffer.nickname]
click = "insert-nickname"

[buffer.channel.nicklist]
click = "open-query"
```

Fixes https://github.com/squidowl/halloy/issues/801.


casperstorm:
@tarkah, the last commit (63bc1c7e4cd250607f052c32e1b30ff5a4a9c202) was just me moving some files around in data/config as i feel like they were misplaced. Let me know what you think.

andymandias (commented):
It's working well for me in testing.  There are two comments I have:
- Having single/double-click behave differently for usernames in messages and usernames before messages is a little confusing to my mind.  E.g. `<andymandias> hey casperstorm!` will have double-click query for `andymandias` and single-click insert nickname for `casperstorm`.  Perhaps my mind is too rigid in this manner, but it would be clearer to me if they react the same to clicks.
- In testing I both liked the double-click to open query action and missed the single-click to insert nickname action.  At the risk of unnecessarily complicating the configuration, would it make sense (and/or be feasible) to allow both single-click and double-click to have actions associated with them?

casperstorm:
@andymandias I've changed this a bit now so that we always single click, no option for double-click. I feel like that is better and easier to understand. Let me know what you think.

src/buffer/user_context.rs
```diff
@@ -174,17 +174,22 @@ pub fn view<'a>(
...
-        .style(theme::button::bare)
-        .on_press(Message::SingleClick(user.nickname().to_owned()));
+    let message = match click {
+        data::config::buffer::NicknameClickAction::OpenQuery => {
+            Message::Query(server.clone(), target::Query::from_user(user, casemapping))
+        }
+        data::config::buffer::NicknameClickAction::InsertNickname => {
+            Message::InsertNickname(user.nickname().to_owned())
+        }
+    };
```
tarkah, line 188: Why do we need `single_click` widget over `button`?

casperstorm: Ive removed it!

tarkah (approved):
LGTM

PR #805, opened 2025-03-02
Persist buffer settings

### LINKED ISSUE #526: remember nicklist visibility per channel
If a pane is closed and reopened, the nicklist appears even if it had been previously hidden.

I think that that visibility state should be retained, at least while Halloy is running. I don't think it needs to (or should) be saved to disk/preserved between runs.

casperstorm:
Previously buffer-settings were only saved while they were on the dashboard.
With this, we now persist buffer-settings by a key.

Fixes #526


andymandias (approved):
LGTM!  I believe the issue primarily affects users with `new-pane` as the default sidebar action.  Works well for me in that scenario.

src/screen/dashboard.rs
```diff
@@ -2008,18 +2057,27 @@ impl Dashboard {
...
+                    b: Box::new(configuration(*b, buffer_settings, config)),
                 },
-                data::Pane::Buffer { buffer, settings } => {
+                data::Pane::Buffer { buffer, .. } => {
+                    let settings = buffer_settings
+                        .get(&buffer)
+                        .cloned()
+                        .unwrap_or(config.buffer.clone().into());
+
                     Configuration::Pane(Pane::with_settings(Buffer::from(buffer), settings))
```
tarkah, line 2081: We don't even need settings at the pane level anymore. We're duplicating it in two places. We can just pass buffer_settings down during view and update for the buffers to use instead of storing it there as well

src/screen/dashboard.rs
```diff
@@ -887,18 +902,39 @@ impl Dashboard {
...
+                        {
+                            let buffer = pane.buffer.data();
+
+                            let settings = pane.update_settings(|settings| {
+                                settings.channel.nicklist.toggle_visibility()
                             });
+
+                            if let Some(buffer) = buffer {
+                                buffer_settings.insert(buffer, settings);
+                            }
```
tarkah, line 916: You can see the issue I mentioned in the other comment here where we have 2 sources of truth. If we just store it at the buffer_settings level, we only have to update it in one place!

src/screen/dashboard.rs
```diff
@@ -887,18 +902,39 @@ impl Dashboard {
...
+                            let settings = pane.update_settings(|settings| {
+                                settings.channel.topic.toggle_visibility()
                             });
+
+                            if let Some(buffer) = buffer {
+                                buffer_settings.insert(buffer, settings);
+                            }
+
+                            self.last_changed = Some(Instant::now());
+                            return (self.save_dashboard(), None);
```
tarkah, line 937: We shouldn't need to manually save the dashboard, this is what `last_changed` is responsible over. We save dashboard during tick if that's set. So we will be saving it twice now.

tarkah (approved):
Nice! Looks great

PR #813, opened 2025-03-07
Add Noctis Lilac theme

VioletSpace:
Adds one of my favourite light themes.


casperstorm (approved):
I like light themes as well! :)

Issue #789, opened 2025-02-21
Graceful Quit-app command


jamestho:
I’m running halloy as a Flatpak on open SUSE tumbleweed, with swayWM. I can force quit the app within sway, but it’s not graceful. 

Can you please create an /command to gracefully allow the halloy app to quit (not just disconnect from the existing irc server(s). 

Thank you!

gdluca:
Glad to see this issue as I was going to ask about the same! :)

casperstorm:
I'll tackle this :)

PR #815, opened 2025-03-08
Quit application thought commandbar and shortcut.

casperstorm:
Fixes https://github.com/squidowl/halloy/issues/789.

This lets you quit Halloy through the command bar (ctrl+k) or through a keyboard shortcut:

```toml
[keyboard]
quit_application = "alt+k"
```

### DIFF
--- a/src/screen/dashboard.rs
+++ b/src/screen/dashboard.rs
@@ -784,7 +784,7 @@ impl Dashboard {
                                 }
                             },
                             command_bar::Command::Application(application) => match application {
-                                command_bar::Application::Quit => (iced::exit(), None),
+                                command_bar::Application::Quit => (self.exit(), None),
                             },
                         };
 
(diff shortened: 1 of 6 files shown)


src/screen/dashboard.rs
```diff
@@ -777,6 +777,9 @@ impl Dashboard {
                                     (window::toggle_fullscreen(), None)
                                 }
                             },
+                            command_bar::Command::Application(application) => match application {
+                                command_bar::Application::Quit => (iced::exit(), None),
+                            },
```
tarkah, line 782: We need to use `Dashboard::exit` which runs our graceful exit logic. This is called during `CloseRequested` in `main`, such as:

```rs
                        window::Event::CloseRequested => {
                            if let Screen::Dashboard(dashboard) = &mut self.screen {
                                return dashboard.exit().map(Message::Dashboard);
                            } else {
                                return iced::exit();
                            }
                        }
```

tarkah (approved):
Nice!

PR #822, opened 2025-03-11
Always keep a pane focused

tarkah:
Extends #787 to remove optional pane focus so we always have a focused pane.


casperstorm (approved):
LGTM, works and tested as well. Really smooth.

andymandias (approved):
A really nice extension on the focus changes.  Works great in my testing!

PR #824, opened 2025-03-11
Bump iced to fd5ed

casperstorm:


src/widget/selectable_text.rs
```diff
@@ -416,13 +416,15 @@ fn draw<Renderer>(
...
-    let x = match paragraph.horizontal_alignment() {
-        alignment::Horizontal::Left => bounds.x,
-        alignment::Horizontal::Center => bounds.center_x(),
-        alignment::Horizontal::Right => bounds.x + bounds.width,
+    let x = match paragraph.align_x() {
+        text::Alignment::Left => bounds.x,
+        text::Alignment::Center => bounds.center_x(),
+        text::Alignment::Right => bounds.x + bounds.width,
+        text::Alignment::Default => todo!(), // TODO: FIX
+        text::Alignment::Justified => todo!(), // TODO: FIX
```
casperstorm, line 424: @tarkah Should we put `Default` together with `Left` arm since we dont have RTL text? And what about Justified?

tarkah: This is just a copy / paste from iced text, looks like both are the same as Left:

https://github.com/iced-rs/iced/blob/master/core%2Fsrc%2Fwidget%2Ftext.rs#L345

Issue #809, opened 2025-03-04
Feature Request: Hotkey to jump to unread channel/buffer


Death916:
When trying to navigate with mainly keyboard on a laptop or other device. If there are many channels joined tabbing through the whole list can be tedious. It would be nice to be able to directly jump to unreads with a keyboard shortcut such as ctrl + u or n or something of the sort.

PR #825, opened 2025-03-13
Shortcuts to Cycle to Next/Previous Buffer with Unread Message(s)

andymandias:
Decided I wanted to tackle issue #809, so this pull request is to add that feature.

I'm uncertain what the default shortcuts should be;  I used `` ` `` & `~` as the characters since that key is near `Tab` which is used in the default shortcuts for cycling buffers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Unreleased
 
+Added:
+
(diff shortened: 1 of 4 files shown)


casperstorm (approved):
LGTM

Issue #568, opened 2024-09-16
Feature request: a way to ignore and notify nicks locally.


Tahomaru:
Hi,
I think it would be great to have a client-side way to ignore and notify nicks using /ignore and /notify commands like many other IRC clients do.
Otherwise we have to rely on the server to implement extensions such as silence/watch/monitor but they're not always available and it might depend on having a user account on said server.
The ignores and notifies could then be managed in the config.toml file in the [ignores] and [notifies] sections like it's done in IRSSI's config, for instance.
Thanks a lot for this great software.

PR #830, opened 2025-03-18
Adds ignore feature

root-goblin:
Addresses #812, #568

Usage:

```toml
# --- $CONFIG/halloy/config.toml
[server.name]
ignored_nicks = [ "nick1", "nick2" ]
```

I'm actually unsure of how to update the documentation for this, some pointers would be appreciated.


andymandias (changes requested):
I think this is a really good start!  I think the main thing we want to change is to allow changes to `ignored_nicks` to be reflected in the application when reloading the configuration while it's running (via the hamburger menu or keyboard shortcut).  See my comment on `new_with_config` for my suggestion how to do so.

data/src/config/server.rs
```diff
@@ -37,6 +37,10 @@ pub struct Server {
     pub password_file: Option<String>,
     /// The command which outputs a password to connect to the server.
     pub password_command: Option<String>,
+    /// A list of nicknames to hide messages from - the messages are not
+    /// removed, and will be shown again when the nick is removed from the list
+    #[serde(default)]
+    pub ignored_nicks: Vec<String>,
```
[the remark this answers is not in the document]
root-goblin: That's a fair point, I was concerned with the legibility of the config file when naming this, I wonder if there's a compromise somewhere here - maybe something like 
```toml
[server.nicks]
ignore = [..]
monitor = [..]
```

data/src/history/filter.rs
```diff
@@ -0,0 +1,71 @@
...
+            .clone()
+            .into_iter()
+            .map(|targ| targ.clone().into())
+            .collect();
+        let filter = Filter::ExcludeSources(target_list);
+        self.filters.push(filter);
+        self
+    }
+
+    pub fn add_time_range_filter<T>(mut self, start: T, end: T) -> Self
```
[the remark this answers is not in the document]
root-goblin: It's not used right now - eventually I was planning to extend smart filters to use this filter architecture, but I should be keeping this out for now.

data/src/history/manager.rs
```diff
@@ -62,10 +64,41 @@ pub enum Event {
 #[derive(Debug, Default)]
 pub struct Manager {
     resources: HashSet<Resource>,
+    filters: HashMap<Server, FilterChain>,
     data: Data,
 }
 
 impl Manager {
+    pub fn new_with_config(config: &Config) -> Self {
```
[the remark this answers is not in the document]
root-goblin: I was actually concerned that this might be happening, I will move this.

data/src/history/manager.rs
```diff
@@ -276,7 +309,13 @@ impl Manager {
...
         buffer_config: &config::Buffer,
     ) -> Option<history::View<'_>> {
-        self.data.history_view(kind, limit, buffer_config)
+        let filters = kind
+            .server()
+            .map(|server| self.filters.get(server))
+            .flatten();
+
+        let view = self.data.history_view(kind, limit, buffer_config, filters);
+        view
```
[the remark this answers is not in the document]
root-goblin: Good catch, thanks!

data/src/config/server.rs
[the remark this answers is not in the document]
andymandias: Hmm, I see what you mean about legibility.  If we put it inside a `nicks` table then that will limit its scope to nicks though.  E.g. if we want to expand to support patterns (which I see are supported in some clients), then it would need to be in another table rather than localized around being a part of the ignore functionality.  It'd be a bit forward-looking, but possibly
```toml
[server.ignore]
nicks = []
```
would be a nice structure for it.  Which could also be expressed in a one-liner as `server.ignore = { nicks = [] }` I believe.

(We won't have this functionality expansion with monitor, since it's limited by the scope of the IRCv3 specification.)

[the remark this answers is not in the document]
root-goblin: With the filter implementation, it's trivial to expand the nicks list to a pattern, the interface outside of the `filters.rs` module should remain the same.

root-goblin:
I've moved all of the data ownership to the `config::Server` struct that gets built when the config gets deserialized - tested and confirmed working with in-client reloads. Note that I've also did a rebase of my branch on top of main before pushing and double checked for any merge conflicts.

andymandias (commented):
I think this is looking really good, and works really nicely now.  There are a couple minor spots I think can be cleaned up, but otherwise I think it's good to go.  The current structure makes the most sense if more filters are to come, so I'm curious to hear what other filters you're planning on.

data/src/config/server.rs
```diff
@@ -272,6 +292,24 @@ impl Sasl {
...
+        .filter_map(|string| {
+            let Ok(user) = User::try_from(string) else {
+                return None;
+            };
+            Some(Source::User(user))
+        })
+        .collect();
+    let filter = Filter::ExcludeSources(sources);
+    log::debug!("loaded filter for ignored nicks {:?}", filter);
+    Ok(vec![filter])
```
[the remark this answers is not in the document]
root-goblin: This is kind of a compromise - the ideal representation for `server::Filters` when halloy is running is a single `Vec<Filter>` or `FilterChain`. 

However since the state is serializing directly from the config, the fields must match the structure as presented by the config so separate fields are required (i.e. ignore).

Serializing directly into a `Vec`  skips allocating a `Vec<Filter>` every time `history::Manager::get_messages()` is called, with the additional benefit of trivially allowing patterns into the mix down the line as a new member of `filters::Filter`. 

I suppose this is premature optimization, but I think this a reasonable compromise.

data/src/config/server.rs
```diff
@@ -182,6 +190,18 @@ pub enum IdentifySyntax {
     PasswordNick,
 }
 
+#[derive(PartialEq, Eq, Debug, Clone, Deserialize, Default)]
+pub struct Filters {
+    #[serde(default, deserialize_with = "deserialize_filters_from_strings")]
+    ignore: Vec<Filter>,
```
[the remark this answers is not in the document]
root-goblin: I believe I chose to leave this as `Vec` for easier manipulation with other filters - `Vec` already has a rich mutation interface, while `FilterChain` requires writing more code.

data/src/history/manager.rs
```diff
@@ -274,9 +276,22 @@ impl Manager {
...
     ) -> Option<history::View<'_>> {
-        self.data.history_view(kind, limit, buffer_config)
+        let server_name = match kind {
+            history::Kind::Server(server) => Some(server),
+            history::Kind::Channel(server, _channel) => Some(server),
+            history::Kind::Query(server, _query) => Some(server),
+            _ => None,
+        };
+
+        let maybe_filters = server_name
```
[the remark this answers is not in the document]
root-goblin: Definitely

data/src/config/server.rs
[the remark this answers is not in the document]
andymandias: My suggestion here is not to change `ignore` from `Vec<Filter>` to `Filter`, but to make the field public.  Then we can ditch `borrow_as_chain` by making `FilterChain::from` take `&Filters` as input instead of `&Vec<Filter>`.

data/src/config/server.rs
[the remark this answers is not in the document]
andymandias: I appreciate your vision for filters is to have a `Vec` that can be nicely iterated, and I agree that would be nice and clean.

For that vision, all filters will need to be deserialized into a single `Vec<Filter>`.  But what's happening here is `ignore` is being deserialized into its own `Vec<Filter>`;  which, with default serde patterns, will be separate from other fields added to the filters config table.  It would need to be combined with other fields for a single `Vec`, at which point I'm not sure that `ignore` being its own `Vec` gets us much.  To avoid that messiness, the `filters` config table would instead need to be deserialized into a single `Vec` (while maintaining the structure on the config TOML).  I suspect the juice is not worth the squeeze there, but I would be happy to be proven wrong.

To be clear, my concern is not so much that `Vec` is too complicated a data structure.  I would avoid it because it introduces state that is meaningless/redundant;  in the case of using it for `ignore` alone.  Having a `Vec<Filter>` for the entire `filters` config table, would not have that issue.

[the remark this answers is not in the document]
andymandias: Actually, having thought it over some more I think the juice will be worth the squeeze.

We can switch the `filters` field in the `Server` config struct to be a `Vec<Filter>` with a custom deserializer.  Then we can utilize the current `Filters` struct inside that deserializer for the purpose of defining the TOML structure only (e.g. similar to the `Data` struct in `deserialize` in `data/src/message.rs`).  We could even have the `ignore` field inside the `Filter` struct be a `Vec<String>`, since it will only live inside the deserializer.  I think that'll give us everything we want.  Let me know if it would be helpful for me to elaborate further.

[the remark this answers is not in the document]
root-goblin: A balanced approach, I'll make the appropriate changes.

As an aside, this is definitely added incentive for me to look into decoupling the application state with the configuration.

data/src/config/server.rs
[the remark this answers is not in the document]
root-goblin: I'll see if this makes sense with the change to the `server::Filters` changes below, pinning to revisit with updates.

root-goblin:
From [`9b263bb` (#830)](https://github.com/squidowl/halloy/pull/830/commits/9b263bb6aa59d1dcd459d1febc01d109168c00e9), it was surprisingly painless to rework the deserializer function to return a `Vec<Filter>`, and work in a custom `Filters` to deserialize into.

This has been tested against the following configurations:

### Well formed with filter
```toml
# ...

[servers.server-name.filters]
ignore = ["nick_name"]

# ...
```

### Malformed with filter
```toml
# ...

[servers.server-name.filters]

# ...
```
### Well-formed without filter
```toml
# ...

# [servers.server-name.filters]

# ...
```

andymandias (approved):
Looks good to me, and works well in my testing.  I think this turned out quite nice!  It will need a changelog entry and updates to the documentation, but let's see if @casperstorm or @tarkah has anything to add.

englut:
Sweet. This works awesome, I am seeing `/me` actions getting through tho. Same with mode changes to the ignored user (e.g. `ChanServ sets mode +v <ignored_nick>`), but I think this isn't a big deal. The `/me` action is more so cause they can write whatever they want in there.

root-goblin:
> I would like an update to book as well as CHANGELOG. The feature works great!

Updated, feel free to take a look

casperstorm:
Let's address the /me action as well @root-goblin. I've also updated a few things, take a look if you like it.

englut:
> Let's address the /me action as well @root-goblin. I've also updated a few things, take a look if you like it.

@root-goblin @casperstorm notifications are also getting through; along with unread indicators in the sidebar.

root-goblin:
> > Let's address the /me action as well @root-goblin. I've also updated a few things, take a look if you like it.
> 
> 
> 
> @root-goblin @casperstorm notifications are also getting through; along with unread indicators in the sidebar. 
> 
> 

I was aware of the issue, I had hope to get the MVP in users' hands first and then address it along with the `/me` issue in a follow up.

I can finish work on both issues in this PR as well if that's how the team prefers. @casperstorm @andymandias

casperstorm:
@root-goblin You still aiming to finish this? :)

root-goblin:
Yes, real life has caught me, but I am rounding back on this this week, in fact.

root-goblin:
Okay, I think it's ready for review.

Things fixed:
- emotes and join/parts no longer appear
- blocked messages no longer trigger unread
- direct queries (DMs) do not open up a new tab

New configuration addition to allow channel-specific filters:

```toml
[servers.server-name.filters]
ignore = [
  "nick_name",
  "#channel another_nick"
]
```

The documentation still needs an update, but the code is ready.

englut:
@root-goblin great work! I love the channel specific ignores. 

I just tested `/me`, and it's ignored! awesome <3. unread indicator is also not triggered by ignored messages, awesome work again!

There is one last scenario: mentions of my username are still getting through by ignored nicks (notifications, the entire message is visible, and unread message indicator).

Overall this is awesome work, thanks so much for all the work and thought you've put in so far @root-goblin :pray:

root-goblin:
> There is one last scenario: mentions of my username are still getting through by ignored nicks (notifications, the entire message is visible, and unread message indicator).
 
That's spicy, let me take a look and see what's going on.

root-goblin:
>mentions of my username are still getting through

@englut I think I've fixed it - I restored some filter steps that cause highlights to leak. If you've time, let me know if this fixed the problem, I have tested it and it seems to be correctly blocking highlights.

englut:
> it seems to be correctly blocking highlights.

@root-goblin I can confirm! Highlights from an ignored Nick doesn't show up and they don't mark the channel as unread! But I am still receiving notifications for the highlight. :pray: soooo close!! I can feel it. Great job! This is feeling and looking amazing :heart:

root-goblin:
> But I am still receiving notifications for the highlight.

Oh this is interesting, by notifications, do you mean desktop notifications or the unread marker in the side bar?

englut:
> Oh this is interesting, by notifications, do you mean desktop notifications or the unread marker in the side bar?

@root-goblin Desktop notifications! I have a notification sound for highlights, and it goes off (along with the notification popup; if I remember correctly, the code for the two are tied together, so this makes sense).

e.g. 
image

englut:
@root-goblin!! your latest changes covers all the test cases I could find :pray: :heart_eyes:

root-goblin:
@andymandias It's ready, all the kinks have been ironed out as far as I'm aware. There's a lot of changes made so I'm pretty sure there's a few things I've missed, but the functionality as tested is all working.

src/buffer/scroll_view.rs
root-goblin, line 1: I could actually revert to using `&config.buffer`, now that history no longer looks up the filter when `get_messages` is called.

data/src/history/manager.rs
```diff
@@ -166,12 +172,29 @@ impl Manager {
...
     }
 
+    pub fn apply_config(&mut self, config: &Config) {
+        log::debug!("Applying new config to history manager.");
+        self.filters.clear();
+        config.servers.entries().for_each(|entry| {
+            if let Some(f) = entry.config.filters.as_ref() {
+                f.ignore.iter().for_each(|filter_string| {
+                    if let Ok(filter) = Filter::try_from_str_with_server(
+                        &entry.server.clone(),
```
root-goblin: This is called only when user has a new configuration, I believe the clone cost is fine here.

data/Cargo.toml
```diff
@@ -46,6 +46,7 @@ const_format = "0.2.32"
 derive_more = { version = "2.0.1", features = ["full"] }
 image = "0.25.5"
 html-escape = "0.2.13"
+regex = "1.11.1"
```
andymandias: We already use `fancy-regex` in this crate.  I believe `regex` is a dependency of `fancy-regex` so it's not a big deal, but if we can get away with only depending on one directly I think that would be preferable.

data/src/history/filter.rs
```diff
@@ -0,0 +1,182 @@
...
+                    None => FilterTarget::Any,
+                };
+
+                (FilterClass::Channel((server.clone(), channel)), target)
+            }
+            None => (FilterClass::Any, FilterTarget::user_try_from_str(value)?),
+        };
+
+        Ok(Self { class, target })
+    }
```
andymandias: `channel_test` as written should work fine most of the time, but we already have functionality for determining a channel, `proto::is_channel`/`data::Client::is_channel` which utilizes the `CHANTYPES` ISUPPORT parameter rather than a default set.  It doesn't currently utilize `CHANNELLEN`, but generally I think we don't want it to (i.e. I think generally we want to consider a channel with a too long name an invalid channel rather than not a channel).  I think we would be better off utilizing that, keeping the canonical channel tests to one if possible.  We will want to use the server's casemapping as well, rather than the default.

data/src/history/filter.rs
```diff
@@ -0,0 +1,182 @@
...
+    ///
+    /// [`Message`]:crate::Message
+    pub fn match_message(&self, message: &Message) -> bool {
+        match &self.target {
+            FilterTarget::User((
+                Source::User(user),
+                Source::Action(action),
+            )) => match &message.target.source() {
+                Source::Action(msg_action) => msg_action == action,
+                Source::User(msg_user) => msg_user == user,
```
andymandias: I'm not sure we need the `Source` tuple as the value for the `FilterTarget::User` variant.  I think a single `User` will be preferable from maintainability and clarity perspectives.  Then, for example, this part of the match arm can look like 
```rust
            FilterTarget::User(user) => match &message.target.source() {
                Source::Action(Some(msg_user)) | Source::User(msg_user) => {
                    msg_user == user
                }
```

data/src/history/filter.rs
```diff
@@ -0,0 +1,182 @@
...
+                _ => match &message.content {
+                    Content::Fragments(fragments) => {
+                        fragments.iter().any(|frag| match frag {
+                            crate::message::Fragment::User(u, _) => u == user,
+                            _ => false,
+                        })
+                    }
+                    _ => false,
+                },
```
andymandias: I don't think this will create any false positives as things currently stand, but I'm not enthusiastic about how broad this is.  It leaves open the potential for unexpected matching in the futue (e.g. if we decided to parse channel topic into fragments with the channel's users, instead of just with the topic setter, then we'd ignore any topic that mentions an ignored nick).

root-goblin:
Known issues that would require ugly fixes: 
- if a user is ignored only in a single channel, all highlight notifications are disabled

data/src/history/manager.rs
```diff
@@ -579,6 +632,33 @@ impl Manager {
...
+
+            self.data
+                .blocked_messages_index
+                .entry(kind.clone())
+                .insert_entry(blocked_message_index);
+        };
+
+        log::debug!("rebuilt blocked message index for {kind}");
+    }
 }
```
andymandias: I think moving most of the computation of filtering away from `history_view` is a good idea.  It does incur a memory overhead, but I think it may be worth it.  (Nothing comes to mind that will avoid the memory overhead without greatly increasing the complexity of history management.)

But, we should ensure it is functionally equivalent as filtering at `history_view`.  Currently, reloading the configuration does not hide messages from any open buffers, as we get with `history_view`-based filtering.  New messages don't appear, but previously unfiltered messages are not hidden.

data/src/history/manager.rs
```diff
@@ -879,35 +966,37 @@ impl Data {
...
-                    .boxed(),
-                )
-            }
+            Some(
+                async move {
+                    let loaded = history::metadata::load(kind.clone()).await;
+                    Message::UpdatePartial(kind, loaded)
+                }
+                .boxed(),
+            )
```
andymandias: Sorry, it's not clear to me: what's the goal in rewriting this section?

data/src/server.rs
```diff
@@ -219,6 +219,10 @@ impl Map {
         self.0.contains_key(server)
     }
 
+    pub fn get(&self, server: &Server) -> Option<Arc<config::Server>> {
+        self.0.get(server).cloned()
+    }
+
```
andymandias, line 225: Doesn't seem like we're using this?

src/main.rs
```diff
@@ -949,8 +946,7 @@ impl Halloy {
                                             ) {
                                                 if dashboard.history().has_unread(
                                                     &history::Kind::Query(server.clone(), query),
-                                                ) || !self.main_window.focused
-                                                {
+                                                ) && !self.main_window.focused {
```
andymandias: If we change this to `&&` then we'll never get a notification for an open buffer, even if the window is otherwise not visible.  I think we want to send those notifications still, no?

book/src/configuration/servers.md
```diff
@@ -207,6 +207,23 @@ Executes the command with `sh` (or equivalent) and reads `password` as the outpu
...
+
+```toml
+# Type: array of strings
+# Values: array of any strings
+# Default: not set
+
+[servers.<name>.filters]
+ignore = ["ignored_user", "another_user"]
+```
+
```
andymandias: Seems like more than just nicknames are supported in the current form of the PR.  If I'm not mistaken, then we should update the documentation.

data/src/history/manager.rs
```diff
@@ -879,35 +966,37 @@ impl Data {
...
         message: crate::Message,
+        blocked: bool,
     ) -> Option<impl Future<Output = Message> + use<>> {
-        use std::collections::hash_map;
+        if blocked {
+            self.blocked_messages_index.entry(kind.clone()).and_modify(
+                |cache| {
+                    cache.insert(message.hash);
+                },
+            );
```
andymandias: I may be overlooking something, but I think we'd want an `.or_insert` as well.  In case this is the first message being blocked.

data/src/history/filter.rs
```diff
@@ -0,0 +1,182 @@
...
+                _ => false,
+            },
+            FilterClass::Any => true,
+            // _ => false,
+        }
+    }
+
+    pub fn is_user(&self) -> bool {
+        matches!(self.target, FilterTarget::User(_))
+    }
```
andymandias, line 162: Seems like this isn't used.

andymandias (changes requested):
There are some good ideas in the new revisions.  If we polish up a few things I think there's potential for a nice new handling of non-temporal filters.

src/main.rs
root-goblin: Setting this to `||` will nullify the filter results and ping every notification (even from filtered highlights) if window is not focused - if we want a notification if the window is focused, we can remove the focus clause altogether, I think. This was actually what englut was running into with my preliminary fix.

book/src/configuration/servers.md
root-goblin: Yes absolutely

data/src/history/manager.rs
root-goblin: Good catch, I wanted to say that the caches will get initialized elsewhere, but it is good to make sure it exists at the insertion point.

data/src/history/filter.rs
root-goblin: For sure, I actually wasn't aware of these helpers

data/src/history/filter.rs
root-goblin: Oh that's crazy - I did not know match can do this kind of black magic.

data/src/history/filter.rs
root-goblin: It is indeed not used, and will be removed.

data/src/server.rs
root-goblin: and removed it will be

data/Cargo.toml
root-goblin: Dropping it makes sense here - regex support is definitely out of scope.

data/src/history/filter.rs
root-goblin: I actually think this `Content::Fragments` match might be vestigial, I can't find a case where a message will get missed from a `Source` match and get caught here.

data/src/history/filter.rs
root-goblin: A few thoughts - this regex is mainly to parse that a a given filter string from the config contains an initial channel name or not - to separate out the `"#channel nick"` versus the `"nick"` usages.
 
After going through how ISUPPORT works, I don't think it makes sense to use these helpers when *constructing* the filter - the filters are built when the config is loaded, before the history is loaded, and most importantly before server connections are made (and `Client::is_channel` can be called). I can see the possibility of confirming the channel's validity after the server connection is available, but I believe the filters need to be usable when the saved history/backlog is loaded.

The failure case for not checking for the validity is not too terrible of an experience - the filter just won't work, likely causing some user troubleshooting + reloading. I feel this is an acceptable failure state.

data/src/history/manager.rs
root-goblin: Done.

data/src/history/filter.rs
root-goblin: This is gone, `match_message` uses something more robust now.

data/src/history/manager.rs
root-goblin: ADHD moment, reverted

data/src/history/filter.rs
andymandias: Ahh, ok I see now, thank you for explaining the expected use case.  I think then it would probably be preferable to go with a less restrictive check.  I don't think it's common, but servers are allowed to set non-standard channel prefixes via [CHANTYPES](https://modern.ircdocs.horse/#chantypes-parameter).  Perhaps we could just do a `str::split_once` with space (which forbidden in channel names & nicknames)?  If the result is `None` it's a `nick` filter, otherwise it's a channel and nick filter.  As you say, having an invalid channel name in there wouldn't be too bad, but being forbidden from using a valid channel name would be frustrating.

src/main.rs
andymandias: I see, we're using that `has_unread` will be `false` for any blocked user as a proxy for blocked status.  I think we should directly check whether the user is blocked instead, something like `has_unread || (!focused && !blocked)`, so that we persist notifications for not-blocked direct messages.

root-goblin: Done!

data/src/history/filter.rs
root-goblin: Done!

data/src/history.rs
```diff
@@ -636,11 +642,11 @@ impl History {
...
-            {
-                message.hidden_urls.insert(url);
+        {
+            message.hidden_urls.insert(url);
 
-                *last_updated_at = Some(Instant::now());
-            }
+            *last_updated_at = Some(Instant::now());
+        }
     }
```
root-goblin, line 650: This is from rustfmt, functionally unchanged

data/src/history.rs
```diff
@@ -320,16 +321,21 @@ impl History {
...
                 ..
             } = self
-            {
-                *max_triggers_unread =
-                    (*max_triggers_unread).max(Some(message.server_time));
-            }
+        {
+            *max_triggers_unread =
+                (*max_triggers_unread).max(Some(message.server_time));
+        }
```
root-goblin, line 338: The indentation change is from rustfmt - the behaviour inside the if-statement remains the same.

src/main.rs
```diff
@@ -989,6 +994,15 @@ impl Halloy {
...
                                     }
+                                    data::client::Event::AddedIsupportParam(param) => {
+                                        if let data::isupport::Parameter::CASEMAPPING(casemapping) = param {
+                                            FilterChain::sync_channels(
+                                                dashboard.get_filters(),
+                                                &server,
+                                                casemapping
+                                            );
+                                        }
+                                    }
```
root-goblin, line 1005: Should be a match here, but no other cases need handling in this PR

root-goblin:
@andymandias I've addressed all the discussion from your previous review, to sum it up:

- Building filters no longer checks a regex - `CaseMap` support is there, and filters are dynamically updated when server sends the `ISUPPORT` parameter
- `match_message` no longer does any `Content::Fragments` checking, and looks strictly at `message.target.source()`
- Pane update logic is added to make sure views are updated when configuration is reloaded from disk
- Notifications now distinctly check for filtered messages in addition to all previous requirements before they are allowed to fire (@englut this actually fixes notifications for nicknames blocked in specific channels!)
- Book is updated to reflect the new syntax for channel-specific ignores

data/src/history/filter.rs
```diff
@@ -0,0 +1,230 @@
...
+    fn from(nick: &str) -> Self {
+        Self::User(Source::User(User::from(Nick::from(nick))))
+    }
+}
+
+#[derive(Error, Debug)]
+pub enum FilterError {
+    #[error("Unable to generate filter from nickname {0:?}")]
+    TryFromUserError(String),
+}
```
andymandias: As far as I can tell we don't use this anymore, since we're trusting the user input (as I believe we should in this case).  Looks like we can get rid of it and the associated `use`.

data/src/history/filter.rs
```diff
@@ -0,0 +1,230 @@
...
+            .filter(|f| f.match_kind(kind))
+            .any(|f| f.match_message(message))
+    }
+
+    pub fn sync_channels(
+        filters: &'f mut [Filter],
+        server: &Server,
+        casemapping: isupport::CaseMap,
+    ) {
+        log::debug!("updating casemap for {server:?}");
```
andymandias: For stylistic consistency (which will hopefully make logs easier to scan), let's change this to something like:
```rust
log::debug!("[{server}] updating casemapping in filters");
```
where the `[{server}]` prefix is the part I'd like to keep consistent.

data/src/history/filter.rs
```diff
@@ -0,0 +1,230 @@
...
+    Server(Server),
+    Channel((Server, Channel)),
+    Any,
+}
+
+#[derive(Debug, Clone)]
+pub enum FilterTarget {
+    User(Source),
+    Any,
+}
```
andymandias: Unless I'm overlooking something, we can change the `User` variant to have a `User value, as:
```rust
pub enum FilterTarget {
    User(User),
    Any,
}
```
A bit of clean-up on a few match arms will fall out of that, but otherwise it looks like it won't affect anything.

data/src/client.rs
```diff
@@ -2058,6 +2059,11 @@ impl Client {
                                                     .try_send(message)?;
                                             }
                                         }
+                                        return Ok(vec![
+                                            Event::AddedIsupportParam(
+                                                parameter,
+                                            ),
+                                        ]);
```
andymandias: This will return early, breaking the containing for loop.  As is done elsewhere, we should have a `let mut events = vec![];` that we append `Event::AddedIsupportParam` events into, then return with the array after the for loop is complete.

data/src/history/manager.rs
```diff
@@ -65,6 +67,8 @@ pub enum Event {
 #[derive(Debug, Default)]
 pub struct Manager {
     resources: HashSet<Resource>,
+    filters: Vec<Filter>,
+    cached_channels: HashSet<history::Kind>,
```
andymandias: Given how closely `cached_channels` is tied to `blocked_messages_index` in function, it would be nice if we could keep that information all in one place.  Could we get rid of `cached_channels` and just use the presence of the channel as in key in `blocked_messages_index` instead?

root-goblin:
@andymandias requested changes have all been applied

andymandias (approved):
Looks good to me, and works well in my limited testing.  We should run this through `rustfmt +nightly`, but before that let's see if @tarkah or @casperstorm has anything they want to add.

data/src/history/filter.rs
```diff
@@ -0,0 +1,218 @@
...
+    Server(Server),
+    Channel((Server, Channel)),
+    Any,
+}
+
+#[derive(Debug, Clone)]
+pub enum FilterTarget {
+    User(User),
+    Any,
+}
```
casperstorm: Does `target` and `class` (and `FilterClass` and `FilterTarget`) needs to be `pub`?

data/src/history/filter.rs
```diff
@@ -0,0 +1,218 @@
...
+pub enum FilterClass {
+    Server(Server),
+    Channel((Server, Channel)),
+    Any,
+}
+
+#[derive(Debug, Clone)]
+pub enum FilterTarget {
+    User(User),
+    Any,
```
casperstorm: With this PR we are able to create the following filters:

```rust
[
"nickname",
"channel nickname",
]
```

What is the purpose of `Any` in `FilterTarget` and  `Server(Server)` in `FilterClass`?

casperstorm (commented):
We are very close! Tested and it works great ❤️ 
Just a few questions and then i'd suggest a few tests in `data/src/history/filter.rs` just to convert a few strings to filters.

data/src/history/filter.rs
root-goblin: It's for completeness - it allows (for example) clear declaration that a filter that matches `FilterClass::Any` and some `FilterTarget::User(specific_user)`. 

Granted, the use case for `FilterClass::SomeSpecific` and `FilterTarget::Any` is not apparent yet - maybe something like `FilterClass::JoinMessage`.

data/src/history/filter.rs
root-goblin: no, I think this is actually vestigial

data/src/history/filter.rs
root-goblin: On revisiting, I suppose the logic can just be re-added when the cases require it.

casperstorm: Yeah I think so too, let's not try to solve tomorrows problem! Once that's fixed I'd say we are good to go!

casperstorm (approved):
🥳 Good job on this one @root-goblin

Issue #781, opened 2025-02-17
High CPU usage when window not visible


Okkonen-GitHub:
I'm getting crazy high cpu usage on Sway (wayland) when the app is not visible. Halloy will take 80 % of one cpu core and Sway's cpu usage jumps to 40 % too. Note that this only occurs when the window is completely not visible (so just unfocusing doesn't trigger this behaviour)
It also does this when I pop out a chat, if either window is not visible.

## System
- Halloy: 2025.1.1
- OS: Artix Linux (rolling)
- DE/WM: Sway
- Arch: amd64

Exagone313:
I can reproduce. When the window is focused on a workspace, the halloy process uses about 3% CPU with some random spikes of 30% or 70%. When the window is on another workspace, it constantly uses at least 80% CPU, usually 130%.

Exagone313:
Could be a dup of #589 though it says *several hours* and we reproduce instantly.

casperstorm:
I believe https://github.com/iced-rs/iced/issues/1810 is the upstream issue.

Exagone313:
I can confirm that after upsetting `WAYLAND_DISPLAY` (thus forcing the usage of Xwayland) the CPU issue disappears.

englut:
I'm getting the CPU issue too now. With both [AUR package](https://aur.archlinux.org/packages/halloy) and the latest commit. Testing the flatpak version of Halloy, it seems unaffected by this issue and still uses Wayland.

casperstorm:
This has been fixed in [2025.4](https://github.com/squidowl/halloy/releases/tag/2025.4)

Okkonen-GitHub:
I regret to inform to you that I am still experiencing the same issue using the version 2025.4 via the [AUR](https://aur.archlinux.org/packages/halloy).

Using this as a temporary fix works for me too:

> I can confirm that after unsetting `WAYLAND_DISPLAY` (thus forcing the usage of Xwayland) the CPU issue disappears.

Issue #589, opened 2024-09-24
Halloy freezes and uses 100% of one CPU thread after it's left minimized for several hours


zsrv:
Halloy freezes and uses 100% of one CPU thread for up to a few minutes when I unminimize its window after leaving it minimized for several hours. During this time, Halloy is mostly unresponsive, but I will see old messages slowly appear in the channel I currently have open. After up to a few minutes, the CPU usage returns to normal and Halloy becomes responsive again. If I leave Halloy minimized for 30 minutes or so then unminimize the window, the problem will reoccur.

OS: openSUSE Tumbleweed
Halloy version: 2024.12 from the openSUSE Tumbleweed package repositories
If it matters, I am in ~50 channels. As of this moment, Halloy is using 530MB of memory.

Daxot:
Same for me on sway. 
When halloy on another workspace that unfocused it's kinda "freezes in time" and after an hour or two when app gets focused again it speeds up to present time: it gets very laggy, all new messages in channel appears one after another, users in users channel list join and leave and so on.
Don't think that is related to channel count, because i'm in 12 channels right now.

OS: Void Linux
Halloy version: 2024.12

casperstorm:
@tarkah @andymandias Any of you noticed this?

andymandias:
I have seen this recently, as I've had Halloy open for longer periods of time.  I suspect this is a consequence of the solution for #364;  while in the background events queue up to be resolved, then the queue is processed once Halloy is foregrounded again.  I've only had to wait ~10s at the longest, but that's still slower than I would otherwise expect (you can watch the processing take place, with enough time to see events unfold).  My expectation is that this behavior is Wayland specific.

tarkah:
Yeah seems winit has a bug for Wayland which causes the event loop to suspend. We keep processing messages in the background and queue them in a channel, and as soon as Halloy is brought back to the foreground, winit resumes and has to process all those messages.

wezm:
>  My expectation is that this behavior is Wayland specific.

> Yeah seems winit has a bug for Wayland which causes the event loop to suspend.

I'm not sure this issue is limited to Wayland. I see the same thing on my X11 based Linux system (I'm only in 3 low traffic channels). I've started up a build with debug symbols and will poke at it in gdb if it does it again.

Daxot:
After [7333371](https://github.com/squidowl/halloy/commit/733337195cecdfad783b007aa07a4b1adfd1f9a2) commit everything works great!

PR #831, opened 2025-03-18
Update iced

andymandias:
Updates iced to include [PR #2849](https://github.com/iced-rs/iced/pull/2849).  Should address #589 and #781.  In my testing it addresses what I believe is the underlying cause of the both issues, that Halloy's event processing would be blocked when not visible on Wayland.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+Fixed:
+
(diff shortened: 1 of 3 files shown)


casperstorm (approved):

casperstorm:
Can we add a note to changelog saying its fixed?

Issue #832, opened 2025-03-19
NOTICE command not sending full notice message.


RealKindOne:
```Zamolxis``` in ```#halloy``` found this bug.

Sending a /notice in halloy does not send the entire message.

I'm using git 4338b37 for this report:

```
/notice ##kindone a b c d e 
```
Seen from another client:
```
[13:49:42] -EvilOne:##kindone- a
```
Including ```:``` does not send the full message

```
/notice ##kindone :a b c d e
```
Seen from another client:
```
[13:50:04] -EvilOne:##kindone- :a
```

Issue #142, opened 2023-07-10
feature request: config to choose where NOTICEs appear


ineeee:
i would like to be able to choose where the NOTICE commands show up. the default choice would have NOTICE commands arrive in their own pane/window just as they do now, and the alternative would behave similarly to WHOIS replies, where its shown in the currently active buffer.

this would improve communication with bots that reply via NOTICE, such as network services or game chat bots.

mgdudas:
Hi folks, just wanted to see if there was any movement on this? I found Halloy today via Hacker News, and it looks amazing so far. I love the configuration the most. But not being able to pick where notices go is kind of a bummer. :-(

casperstorm:
ill pick it up now!

casperstorm:
@ineeee @mgdudas could you two provide more context to what would be the best solution here? I am not interacting with bots so I am not used to NOTICE.

casperstorm:
@mgdudas and @ineeee i pictured the following:

```
[event.notice]
reroute = "follow-buffer' | "server" | "default" 
```
with `follow-buffer` all NOTICE would land the current selected buffer.
with `server` all NOTICE would land in server buffer
with `default` all NOTICE would land in default target (query, etc)

I am wondering how the `follow-buffer` ux would be since (if i understand correctly) you could type, eg `/query Q hello` and the `hello` would go the a new query buffer with Q, but Q would reply back into the buffer you initially typed `/query Q hellp` from. Is that the expected behavior? Or would you expect the `hello` to also appear in the current buffer, but not visible to others?

mgdudas:
Thank you for your work on this @casperstorm This all looks great.

My understanding of the functionality of follow-buffer would be that you would receive the notice in the currently selected buffer, as long as the sender is also in that buffer. (ie; you're in the same channel)

An example I have from other irc clients I use this on is we have a bot in #channel named chanbot. I can set chanbot to either send me a message via privmsg or notice. If I set it to notice, the client will print said notice in #channel for only me to see. The idea being that the notice would not follow me to another active buffer that chanbot was not in. If I have a private message buffer as well, it would also be printed there. An edge case I've seen in hexchat is that if there is a private message buffer at all, it will not print the notice in #channel even if #channel is active. But that seems to only be the case in that client.

Let me know if this makes sense and if you need further explanation from me.

casperstorm:
[...]

PR #833, opened 2025-03-19
Expand Commands

andymandias:
A bit of a grab bag re: input commands.
- Adds the `/notice` command, addressing #832.
- Incorporates ISUPPORT parameters into command parsing.
- Provide immediate feedback for some input errors, rather than waiting for the user to attempt to send, to make them easier for the user to address.

### DIFF
--- a/src/buffer/input_view.rs
+++ b/src/buffer/input_view.rs
@@ -3,7 +3,7 @@ use data::input::{self, Cache, Draft};
 use data::target::Target;
 use data::user::Nick;
 use data::{buffer, client, history, Config};
-use iced::widget::{container, row, text, text_input};
+use iced::widget::{column, container, text, text_input};
 use iced::Task;
 
 use self::completion::Completion;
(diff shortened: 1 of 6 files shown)


src/buffer/input_view.rs
```diff
@@ -87,12 +87,10 @@ pub fn view<'a>(
...
-    let overlay = state
-        .error
-        .as_deref()
-        .map(error)
-        .or_else(|| state.completion.view(cache.draft, config))
-        .unwrap_or_else(|| row![].into());
+    let overlay = column![]
+        .spacing(4)
+        .push_maybe(state.completion.view(cache.draft, config))
+        .push_maybe(state.error.as_deref().map(error));
```
casperstorm, line 93: This looks GREAT @andymandias

casperstorm:
I dont want to scope creep, but does it make sense to take a look at this as well when you are working with notices? https://github.com/squidowl/halloy/issues/142

andymandias:
> I dont want to scope creep, but does it make sense to take a look at this as well when you are working with notices? #142

To my mind it's a bit of an open question how this should function (should notice replies be sent to the active buffer, the last buffer that sent a notice to that target, requiring labeled response, or something else).  So I'm going to hold off on trying to work it into this PR, but will keep it on my radar.

casperstorm:
I think this looks nice, @tarkah do you mind going through this?

tarkah (approved):
Sorry for the late review, this looks great!! I love the error formatting on the input now. Really nice UX improvements

PR #837, opened 2025-03-22
Case-Insensitive Emoji Picker

andymandias:
A few small changes to make the emoji picker case insensitive.


casperstorm (approved):
Ah nice, good change.

PR #843, opened 2025-03-25
Navigate Pickers with Arrow Keys

andymandias:
Allows the command and emoji pickers to be navigated with up & down arrow keys.


casperstorm (approved):
Nice QOL

PR #844, opened 2025-03-26
Pre-Login ChangeHost Broadcast to Server Buffer

andymandias:
Broadcasts for chghost messages received during the login process would show up in the history of queries only (not in channels, since they haven't been joined at that stage, nor the server buffer).  This PR is to suppress broadcasts from queries (to match the behavior seen in channels), since they can start to pile up, and broadcast to the server buffer instead.  A minor added benefit that the information is available to users even if they don't happen to have a query open at launch.


casperstorm (approved):
Not tested, but codes looks good.

Issue #845, opened 2025-03-27
Use same nick colours in other messages


auronandace:
The nicklist and normal messages use the nick colours.
Join/part/quit/change_host messages only use the set colour in the theme config for the whole message. Can they be changed so that the nick portion of the message uses the same colours as the nicklist? I noticed that urls are coloured separately in quit messages so it seems the same could be done for nicks.
This can help with visual identification making it easier to follow when people change nicks or part/quit. Weechat does this.

PR #853, opened 2025-03-29
User aware server messages

casperstorm:
This will show nicknames colored correctly in server messages.

Fixes: https://github.com/squidowl/halloy/issues/845

Screenshot 2025-03-29 at 19 47 38

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 Added:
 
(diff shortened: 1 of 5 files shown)


andymandias (approved):
Looks really good.  And it really improves the legibility of the server messages for me.  Didn't test all server message variants, but seeing as how they all utilize `parse_fragments_with_user` or `parse_fragments_with_users` I expect they'll all look just as good.

tarkah (approved):
Really nice 👌

PR #869, opened 2025-04-01
Expanded Buffer Action Settings

andymandias:
Adds configuration settings for various buffer actions that are ambiguous as to how they should work.  For example, a user might want clicking on `Message` in a user context menu to open a new pane, replace the currently focused pane, or open a new window.

In order to do this I did a fair amount of refactoring to centralize the buffer action logic into `open_buffer`.

This is a bit of a draft, as I'm not entirely sure I've picked the correct default settings or setting categories.  I'm open to changing either or both.


book/src/configuration/buffer_actions.md
```diff
@@ -0,0 +1,120 @@
...
+```toml
+# Type: string
+# Values: "new-pane", "replace-pane", "new-window"
+# Default: "new-pane"
+
+[sidebar]
+click_highlight = "new-pane"
+```
+
+## `click_user_name`
```
casperstorm, line 70: Small nit but i think we use "username" other places (without space)

PR #873, opened 2025-04-01
add lints.clippy section to Cargo.toml and specify and tackle one

auronandace:
Now we are using edition 2024 we can safely include a `[lints.clippy]` section to the `Cargo.toml` without worrying about the MSRV.

I took the liberty to tackle one lint from the pedantic category that I think improves the readability of the codebase. I've kept it at `warn` level for now but not sure if you'd prefer `deny` to really enforce it.

The idea is that we can cherry pick which lints we want. It would likely make for a good first time contribution for people new to the codebase.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -102,3 +102,6 @@ opt-level = 1
 # Enable a large amount of optimization in the dev profile for dependencies.
 [profile.dev.package."*"]
 opt-level = 3
+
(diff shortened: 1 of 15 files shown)


andymandias (approved):
Haven't tested, but looks good to me.

casperstorm (approved):

PR #874, opened 2025-04-02
Save Dashboard After Cycling via Keyboard Shortcut

andymandias:
Minor fix to mark the dashboard as changed when cycling a buffer via keyboard shortcut (so that it is saved).


casperstorm (approved):

tarkah (approved):
I like this diff :)

Issue #864, opened 2025-03-31
New messages being placed above chat history messages when sent in close succession


McFarland97:
Possible replication steps:

Utilize these in config: 
use_tls = true
chathistory = true

Example of behavior in question:

https://i.sebz.xyz/r/Wug19t.png?compress=false

McFarland97:
Here's a screen recorded example for easier understanding of the issue: https://i.sebz.xyz/u/B9xfVh.mp4

casperstorm:
I've seen this once before where @smakkerlak had the issue. His NTP and machine was drifted apart.

@andymandias perhaps we should look at timestamp update.

smakkerlak:
Can confirm. Here, i told @casperstorm that the new themes thingy was neato, he then said thanks.

![Image](https://github.com/user-attachments/assets/d0baf8de-0a54-453b-bf26-3560de8431bb)

andymandias:
> [@andymandias](https://github.com/andymandias) perhaps we should look at timestamp update.

I'll take a look at it this week.

PR #875, opened 2025-04-02
Update Timestamps from Echoes

andymandias:
Changes to `insert_message` in order to update a message's timestamp.  Primarily used for updating a sent message's timestamp when receiving an echo.  May also be used when updating a message via id.

Should help with #864.  But, to limit the required computation, only if the clock difference is less than or equal to 5m.

Also enables requesting the `echo-message` capability even when the `labeled-response` capability is not available.  The heuristic echo determination seems to be working well, and `labeled-response` has limited support.

### DIFF
--- a/data/src/history.rs
+++ b/data/src/history.rs
@@ -539,15 +539,15 @@ pub fn insert_message(messages: &mut Vec<Message>, message: Message) {
     let mut replace_at = None;
 
     for stored in &messages[start_index..end_index] {
-        if (message.id.is_some() && stored.id == message.id)
-            || ((stored.server_time == message.server_time
-                || (matches!(stored.direction, message::Direction::Sent)
-                    && matches!(message.direction, message::Direction::Received)
-                    && message.is_echo))
-                && has_matching_content(stored, &message))
+        if replace_at.is_none()
+            && ((message.id.is_some() && stored.id == message.id)
+                || ((stored.server_time == message.server_time
+                    || (matches!(stored.direction, message::Direction::Sent)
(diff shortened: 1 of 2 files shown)


casperstorm (approved):
LGTM, @tarkah any comments?

data/src/history.rs
```diff
@@ -557,11 +558,32 @@ pub fn insert_message(messages: &mut Vec<Message>, message: Message) {
...
+                Ordering::Less => {
+                    messages.remove(index);
+                    messages.insert(insert_at, message);
+                }
+                Ordering::Equal => messages[index] = message,
+                Ordering::Greater => {
+                    messages.insert(insert_at, message);
+                    messages.remove(index);
+                }
+            }
```
tarkah, line 586: Why do we need another binary search here when we've already identified an insert_at index above?

andymandias: If we're in this branch then `replace_at` has been set, after which we `break`, so `insert_at` may not be accurate.  But we should be able to do the search over a smaller range (or just finish the linear search in the smaller range).

tarkah: Yeah let's just remove the `break` and add an if replace.is_none guard so we don't overwrite replace_at then finish out that loop.

Although it may actually be less efficient than binary search since O(n) of small search range may be more than O(log(n)) of all messages in history. 

But I think it's worth the potential perf hit since it's overall simpler and easier to grok since we already have an accurate replace_at and insert_at

andymandias: Works for me.  Will be straightforward to revisit if necessary.

tarkah (approved):

PR #878, opened 2025-04-03
Colorize Actions

andymandias:
Performs the same colorizing fragment parsing on actions that are performed on privmsgs, to complement the colorizing of server messages.


tarkah (approved):
Nice 👍

PR #881, opened 2025-04-03
enforce clippy lint for ignored_unit_patterns

auronandace:
This lint helps when it comes to future refactors. By explicitly matching on the unit pattern `()` instead of the catch all `_` the type system will come to your rescue to ensure you've not forgot to match against the newly introduced variants.

Also, I reordered the lint list alphabetically.


casperstorm (approved):
Nice one. Thanks!

PR #882, opened 2025-04-03
enforce clippy lint map_unwrap_or

auronandace:
This lint makes code a little more consise and helps to avoid use of some `unwrap_*` methods. 1 method call instead of 2.

Note that the ordering is that the failure part comes first.


auronandace:
Really sorry for the noise. I have no idea what I'm doing regarding rebasing. Shall I delete this and start afresh?

PR #888, opened 2025-04-04
add cloned_instead_of_copied clippy lint

auronandace:
This is better as you are making sure the type implements `Copy`. If it doesn't then you can use `cloned` instead. Helps you to identify when types are copy and means wherever you've used clone then you know it is more expensive.


casperstorm:
Requested a review from an AI. Let's see what the fuss is about.

PR #894, opened 2025-04-05
add clippy lint if_not_else

auronandace:
Understanding is usually easier when the `if` is checking whether something is `true` instead of relying on negation.


auronandace:
Comments on IRC channel indicate that having the `if` branch as the shorter branch is deemed more readable.

PR #895, opened 2025-04-06
TitleBar Button & Shortcut to Mark Buffer as Read

andymandias:
Adds a titlebar button and keyboard shortcut for marking a buffer's contents as read.  Both actions will send a `MARKREAD` message when `read-marker` is supported by the server and the buffer is a channel/query.

A few design choices that I think need review:
- I picked a circled checkmark for the button to distinguish it from the "Scroll to Bottom" button, but maybe there's a better option?
- ~~The shortcut I picked is `Ctrl/Command + Shift + m` (since `Ctrl/Command + m` is taken).  I couldn't see anything good to move `toggle_nick_list` to, and I think `reload_configuration` has more right to `r` than `mark_read` does.  Maybe there's a better option?~~  Edit:  I misread the `toggle_nick_list` shortcut, but we don't want to use `Ctrl/Command + m` since it's a system shortcut on macOS.
- I chose to show the Mark as Read button whenever there is a message in the buffer's history that (a) is considered viable for setting the read marker and (b) is newer than the buffer's current read marker.  But that message may sometimes not be visible (e.g. a filtered out server message).  This is potentially confusing for users, but I was thinking it was worth it for the potential utility.


casperstorm:
First of all, this is awesome @andymandias.
My first question is: what if we instead of a dedicated button let the scroll-to-bottom call the mark as read?
It felt quite strange to first click "mark as read" and then "scroll to bottom" after 🤔

Another thing, now that we have this functionality: would it make sense to add a button to buffer context (in sidebar) to say "mark as read"? Idea here is like a unread mail you can mark as read without reading it.

andymandias:
Thanks for the review @casperstorm.

Personally, I like to keep scroll-to-bottom and mark-as-read separate actions, but I think combining them is a totally reasonable.  I can add a configuration option for that.

I think adding Mark as Read to the context menu for sidebar buttons is a great idea, I'll do that.

casperstorm (approved):
This came out really nice!

PR #899, opened 2025-04-07
Add trailing 0 to the Windows installer

KaiKorla:
Add trailing 0 to the Windows installer to match the installer version the product version.
Think i missed this in the past. :-/

regards


casperstorm:
Why is that needed?

KaiKorla:
Streamlining the installer Version, the installed product version and the winget package version with the Windows binary ProductVersion (which needs an x.y.z Schema).

casperstorm (approved):

PR #901, opened 2025-04-08
correct toml headings in actions docs

auronandace:
I noticed the new actions docs all had `[sidebar]` as the heading in the code section. Corrected this.


andymandias (approved):
Thanks for catching this!

Issue #857, opened 2025-03-30
[Bug] 0x0F byte it printed at the end of formatted messages.


kasper93:
I see  (0x0F) char at the end of formatted message. This suppose to be "reset" for color. Color itself looks like working, so probably the end char is not stripped correctly.

You can see this in `code-cone` messages for example in #videolan @ Libera.

It looks like this:
![Image](https://github.com/user-attachments/assets/a390bccc-75d7-4f87-94ce-3fb55f2c889b)

We can see 0x0F at the end of line.

Also not sure if the link is formatted correctly, because it looks different in other irc clients. Though, I have not looked at the raw representation of this message.

casperstorm:
@tarkah perhaps you can look at this.

auronandace:
It doesn't seem to be exclusively at the end. See the first line in this image:

![Image](https://github.com/user-attachments/assets/d2452b5b-4ecc-4e9f-96ca-011c107f8337)

Seems to happen at the end of the URL.

tarkah:
Are you positive this is 0x0F and not some other control character? Are you able to RUST_LOG=trace and see what the raw message is?

andymandias:
Raw messages can help determine what might be the source of the bug.  For example, I see this:

![Image](https://github.com/user-attachments/assets/193fd251-2e20-474f-a10e-62935feae84e)

with the corresponding raw message:

`\u{f}\u{3}03VLC\u{f} \u{3}05master\u{f} \u{3}06Marvin Scholz\u{f} * [open] merge request !7118: meson: DeckLink: add dependency on CoreFoundation * \u{3}14https://code.videolan.org/videolan/vlc/-/merge_requests/7118\u{f}`

kasper93:
Very nice, thank you for the input.

auronandace:
> ![Image](https://github.com/user-attachments/assets/193fd251-2e20-474f-a10e-62935feae84e)


I managed to catch the same line as you. This is what it looked like to me:

![Image](https://github.com/user-attachments/assets/99ef907a-f612-43de-a65d-fad50a46856c)

I can confirm that the ending byte of that message was `0x0F` which in IRC formatting represents a reset of all formatting to default values.

auronandace:
[...]

PR #903, opened 2025-04-08
Fix for Unwanted Glyphs

andymandias:
A fix for the unwanted behavior documented in #857.  Modifies test in #898 to reproduce the unwanted behavior.

### DIFF
--- a/data/src/message.rs
+++ b/data/src/message.rs
@@ -676,12 +676,27 @@ fn parse_fragments_inner<'a>(text: String) -> impl Iterator<Item = Fragment> + u
     })
     .flat_map(move |fragment| {
         if let Fragment::Text(text) = &fragment {
-            if let Some(formatted) = formatting::parse(text, &mut modifiers, &mut fg, &mut bg) {
-                if formatted
+            if let Some(fragments) = formatting::parse(text, &mut modifiers, &mut fg, &mut bg) {
+                if fragments
                     .iter()
                     .any(|fragment| matches!(fragment, formatting::Fragment::Formatted(_, _)))
                 {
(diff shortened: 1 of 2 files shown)


casperstorm (approved):

data/src/message.rs
```diff
@@ -1653,10 +1680,41 @@ mod test {
...
-                "\x0f\x0303VLC\x0f \x0305master\x0f \x0306somenick\x0f describe commit \x0314https://someurl/7089\x0f",
+                "\u{f}\u{3}03VLC\u{f} \u{3}05master\u{f} \u{3}06somenick\u{f} \u{3}14http://some.website.com/\u{f} * describe commit * \u{3}14https://code.videolan.org/videolan/vlc/\u{f}",
+                vec![
+                    Fragment::Formatted{ text: "VLC".into(), formatting: Formatting { fg: Some(Color::Green), ..Formatting::default() }},
+                    Fragment::Text(" ".into()),
+                    Fragment::Formatted{ text: "master".into(), formatting: Formatting { fg: Some(Color::Brown), ..Formatting::default() }},
+                    Fragment::Text(" ".into()),
+                    Fragment::Formatted{ text: "somenick".into(), formatting: Formatting { fg: Some(Color::Magenta), ..Formatting::default() }},
+                    Fragment::Text(" ".into()),
+                    Fragment::Url("http://some.website.com/".parse().unwrap()),
```
kasper93, line 1691: Shouldn't the url be formatted to?

andymandias: Currently, URLs are not allowed to be formatted by the sender.  That may change in the future, but probably not as part of this pull request.

PR #904, opened 2025-04-09
Support non-alphanumeric characters that could be used on a keyboard

englut:
Found the config doesn't allow for non-alphanumeric characters for keybinds yet! Here's a PR that adds all the ones I could think of. 

If you've some other keyboard that needs more, let me know :pray:


andymandias:
I'm not sure `"+"` can be accessed with how shortcuts are currently parsed.  In my testing `"ctrl+?"` does not work as the equivalent of `"ctrl+shift+/"` on an ANSI keyboard (with how shortcuts are currently processed), where that equivalence might be expected by a user.  My feeling is that supporting both of those functionalities would be the most user friendly, but I would defer to @casperstorm's judgment on the matter.

englut:
> I'm not sure `"+"` can be accessed with how shortcuts are currently parsed. In my testing `"ctrl+?"` does not work as the equivalent of `"ctrl+shift+/"` on an ANSI keyboard (with how shortcuts are currently processed), where that equivalence might be expected by a user. My feeling is that supporting both of those functionalities would be the most user friendly, but I would defer to @casperstorm's judgment on the matter.

Yeah, they don't work on my keyboard too. Which makes sense, cause my keyboard isn't actually sending that character. I'm just trying to be inclusive, in case there is a keyboard out there that can send that character without needing a key combination :)

But I can remove those keys easily, just let me know!

englut:
> My feeling is that supporting both of those functionalities would be the most user friendly, but I would defer to @casperstorm's judgment on the matter.

The more I think about this, I've changed my mind and am leaning more towards removing these keys as it would cause confusion. :pray:

andymandias (approved):
I think it's reasonable to restrict the newly supported characters to unmodified ANSI keys for now.

Issue #860, opened 2025-03-30
segfault when closing while running vulkan, issue with choosing wgpu backend + device


JkktBkkt:
With selected backend getting set to vulkan (at least on my nvidia system), closing the application via either it's own interface ot window decorations leads to a segfault getting reported.
The issue has been fixed in 24.0.3 of wgpu: https://github.com/gfx-rs/wgpu/commit/b2171dd1a4c2884f20ae95d67c21819ade336a5e
Could you bump the dependency?

I am also wondering what the logic is for choosing which vulkan device to use. On my system it picks intel_icd.x86_64.json over intel_hasvk, virtio and nvidia ones, which results in a panic.
Manually overriding VK_ICD_FILENAMES to nvidia one results in correctly using vulkan as mentioned above, others just fail-over to opengl on the nvidia card.
For reference, most wgpu examples don't launch with this icd either, some others pick the nvidia one despite launch flags, misrepresenting the tests.
I have built the wgpu examples locally, and have verified the crashing on startup behavior with auto-choosing the icd in both the version provided by void-linux packages and binary releases from here on github, 2024.11-2025.3
Trace with auto-picked `intel_icd.x86_64.json` :
```
[destroyed object]: error 7: importing the supplied dmabufs failed
Protocol error 7 on object @0: 

thread 'main' panicked at /host/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wgpu-24.0.1/src/backend/wgpu_core.rs:3432:18:
wgpu error: Validation Error

Caused by:
  In Surface::configure
    Surface does not support the adapter's queue family


stack backtrace:
   0:     0x560c356c0aca - <unknown>
   1:     0x560c356ef693 - <unknown>
   2:     0x560c356bb863 - <unknown>
   3:     0x560c356c0912 - <unknown>
   4:     0x560c356c1ac0 - <unknown>
   5:     0x560c356c18a0 - <unknown>
   6:     0x560c356c2222 - <unknown>
   7:     0x560c356c1fca - <unknown>
   8:     0x560c356c0fd9 - <unknown>
   9:     0x560c356c1c5d - <unknown>
  10:     0x560c33da08b0 - <unknown>
  11:     0x560c33d47ed8 - <unknown>
  12:     0x560c34a6ea37 - <unknown>
  13:     0x560c34a8c4d1 - <unknown>
  14:     0x560c34847ebd - <unknown>
  15:     0x560c33deb297 - <unknown>
[...]

PR #907, opened 2025-04-10
Update wgpu

andymandias:
May fix a segfault on application exit for some systems (e.g. issue #860).  Unfortunately, does not fix the segfault on application exit for my system, but does not have any apparent adverse effects.


JkktBkkt:
It got more elaborate: 
```
warning: queue 0x5615173dea00 destroyed while proxies still attached:
  wl_registry#65 still attached
Segmentation fault
```
The warning is expected and can be seen with vkcube-wayland and other applications, however, halloy still segfaults, perhaps there's something else beside this.

andymandias:
It should be safe to merge on the off chance it helps someone, but I have no evidence that it helps anyone at the moment.  I think we can close this and wait for https://github.com/iced-rs/iced/pull/2887 to land, see if it provides a real resolution to #860.

PR #914, opened 2025-04-13
add clippy lint unnecessary_wraps

auronandace:
This was an interesting discovery for me.

Part I already knew: If all return places from a function are guaranteed to be `Some` then it is best to not return an `Option<Foo>` but just simply return `Foo`.

Part that interested me: You can `.map()` an `Option<Foo>` to turn it into an `Option<Bar>`. If you didn't return an `Option` then you'd need to manually implement `.map()` for your type.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -119,6 +119,7 @@ manual_string_new = "deny"
 map_unwrap_or = "deny"
 semicolon_if_nothing_returned = "deny"
 uninlined_format_args = "deny"
+unnecessary_wraps = "deny"
 
 [workspace.lints.clippy]
 cloned_instead_of_copied = "deny"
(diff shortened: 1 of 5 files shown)


andymandias:
Apologies if you already know this, but manually implementing `.map()` is not so complicated as it might seem.  When `Option::map(|opt| f(opt))` is always passed a `Some(thing)`, then it is equivalent to `Some(f(thing))`.  For example, if you have

```rust
        Message::UserContext(message) => user_context::update(message).map(Event::UserContext),
```

then the manual implementation of `.map()` is

```rust
        Message::UserContext(message) => Some(Event::UserContext(user_context::update(message))),
```

assuming `update` has been modified to return an `Event` instead of an `Option<Event>`.

I think it would be worthwhile to convert in this case.

PR #915, opened 2025-04-13
add test for strikethrough formatting

auronandace:


andymandias (approved):
Looks good.  If you're interested, I think these tests could be converted to a set of `assert_eq!` tests (e.g. similar to those at the bottom of `data/src/message.rs`).  Not necessary for this PR though.

casperstorm:
> Looks good.  If you're interested, I think these tests could be converted to a set of `assert_eq!` tests (e.g. similar to those at the bottom of `data/src/message.rs`).  Not necessary for this PR though.

+1

PR #917, opened 2025-04-13
Sidebar Theming Tweak

andymandias:
While working on some themes I noticed that the sidebar buttons don't have hover state for open/focused buffers.  This PR adds those hover states, with the downside of not disabling the focused buffer's button.


casperstorm (approved):
Good catch buddy.

PR #918, opened 2025-04-14
convert to assert_eq tests

auronandace:
Follow-up to #915.


andymandias (approved):
Looks good to me.  Thanks for taking these on.

PR #923, opened 2025-04-15
add test for fragment highlights

auronandace:
Improves test coverage for parsing fragments with highlights.

I have absolutely no idea how to use or formulate regexes. I was hoping to get `Fragment::HightlightMatch` to show up but I'm certainly doing something wrong.

### DIFF
--- a/data/src/message.rs
+++ b/data/src/message.rs
@@ -1598,10 +1598,8 @@ impl PartialOrd for MessageReferences {
 
 #[cfg(test)]
 mod tests {
-    use fancy_regex::Regex;
-
     use crate::{
-        config::{highlights::{Match, Nickname}, Highlights},
+        config::{highlights::Nickname, Highlights},
         message::{formatting::Color, Content, Formatting, Fragment},
         user::Nick, User
     };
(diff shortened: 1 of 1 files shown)


data/src/message.rs
```diff
@@ -1742,4 +1749,46 @@ mod test {
...
+                    ],
+                    "Bob",
+                    Some(&nick),
+                    &Highlights {
+                        nickname: Nickname {exclude: vec!["Dave".into()], include: vec!["Greg".into()]},
+                        matches: vec![
+                            Match {
+                                regex: Regex::new(r#"[h]"#).unwrap(),
+                                exclude: vec!["hope".into()],
+                                include: vec!["with".into()],
```
andymandias, line 1772: The `exclude` and `include` here are for channels to [include](https://halloy.chat/configuration/highlights.html#include-1)/[exclude](https://halloy.chat/configuration/highlights.html#exclude-1).  I think it would probably be easier to test that functionality elsewhere.

data/src/message.rs
```diff
@@ -1742,4 +1749,46 @@ mod test {
...
+                    Fragment::User(User::try_from("Greg").unwrap(), "Greg".into()),
+                    Fragment::Text(". I hope ".into()),
+                    Fragment::User(User::try_from("Dave").unwrap(), "Dave".into()),
+                    Fragment::Text(" doesn't notice.".into()),
+                ],
+            ),
+        ];
+        for ((text, channel_users, target, our_nick, highlights), expected) in tests {
+            let actual = parse_fragments_with_highlights(text, channel_users, target, our_nick, highlights);
+            assert_eq!(Content::Fragments(expected), actual);
```
andymandias, line 1791: `Content`'s equality is defined such that it ignores formatting.  I think for the test you want, you should use something like
```rust
            if let Content::Fragments(actual) =
                parse_fragments_with_highlights(text, channel_users, target, our_nick, highlights)
            {
                assert_eq!(expected, actual);
            } else {
                panic!("expecting fragments with highlighting");
            }
```

andymandias (approved):
Looks good to me.

PR #925, opened 2025-04-15
Send Connection Actions (`JOIN`/`on-connect`/etc) After 1st MOTD

andymandias:
Moves actions performed on connection to after the first MOTD, which according to https://modern.ircdocs.horse/#connection-registration is the last message we should be guaranteed to get from the server after the registration process completes.  This should ensure we have all of the server's `RPL_ISUPPORT` before we perform any of these actions, and will hopefully help with the report of `on-connect` actions sometimes not being recognized.


casperstorm (approved):
This make sense 👍🏻

PR #927, opened 2025-04-16
ability to show modal prompt before open URL

### LINKED ISSUE #829: Prompt before opening link (privacy/security concern)
# Scenario

People may want to connect to IRC behind a proxy for privacy/security, but their default browser isn't setup behind a proxy.

## Current Situation

Someone could send a user connected behind a proxy a URL or post one in a channel they control. When clicking on the buffer, the user could accidentally click on the link which automatically opens & reveals their IP.

## Proposal

When clicking a link in Halloy, it should prompt before opening in the browser. This gives the user an opportunity to make sure that their browser isn't going to leak their identity.

casperstorm:
Fixes #829 

Screenshot 2025-04-16 at 14 37 16

Test with

```toml
[buffer.url]
prompt_before_open = true
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 Added:
 
+- Ability to show a modal prompt before opening a URL
 - Title bar button and keyboard shortcut to mark a buffer as read (will update the read marker as well, if the `read-marker` capability is available)
 - Mark as Read settings to control when buffers are automatically marked as read
 
--- a/book/src/configuration/buffer.md
+++ b/book/src/configuration/buffer.md
@@ -18,6 +18,7 @@ Buffer settings for Halloy.
 11. [Status Message Prefix](#bufferstatus_message_prefix) - Status message prefix settings
 12. [Text Input](#buffertext_input) - Customize the text input for in buffers
 13. [Timestamp](#buffertimestamp) - Customize how timestamps are displayed within a buffer
+14. [Url](#bufferurl) - URLs in buffers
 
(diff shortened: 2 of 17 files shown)


andymandias (approved):
Looks good, works well in my testing.

Two thoughts:
- Since the related issue is a privacy concern, should we default to the more private setting?  I.e. should we default to true?
- If we default this value to true, should we default `preview.enabled` to false by the same reasoning?

casperstorm:
> Looks good, works well in my testing.
> 
> Two thoughts:
> 
> * Since the related issue is a privacy concern, should we default to the more private setting?  I.e. should we default to true?
> * If we default this value to true, should we default `preview.enabled` to false by the same reasoning?

I rather want to make a small guide section on our website with how to optimize for "privacy".

0100010001010011:
Hi! Thanks for implementing `prompt_before_open`.

I’d like to suggest a related feature: allowing users to configure which browser or command Halloy should use when opening links.

For example, my default system browser is Brave, but for links opened from Halloy I would prefer to use LibreWolf, because I can keep a separate privacy/proxy configuration there.

Something like this in `config.toml` would be useful:

```toml
[buffer.url]
prompt_before_open = true
open_command = "librewolf %u"
```

Or maybe an array-style command:

```toml
[buffer.url]
prompt_before_open = true
open_command = ["librewolf", "%u"]
```

This would help users who want IRC links to open in a dedicated browser/profile instead of the system default browser.

Thanks for your work on Halloy!

PR #928, opened 2025-04-16
Allow Queries to be Matched by Preview Include/Exclude Settings

andymandias:
I currently use `exclude = ["*"]` to hide all card previews, but since queries are excluded from matching that setting I was still getting card previews inside direct messages.  This change allows queries to be matched as well.

Also changes include/exclude matching to utilize the server's casemapping to normalize the strings.  Can be broken off into a separate PR if needed.

### DIFF
--- a/data/src/buffer.rs
+++ b/data/src/buffer.rs
@@ -1,13 +1,11 @@
 use core::fmt;
 
 use serde::{Deserialize, Serialize};
-use url::Url;
 
 use crate::config::buffer::NicknameClickAction;
-use crate::preview::{self, Preview};
 use crate::serde::default_bool_true;
 use crate::target::{self, Target};
-use crate::{Server, channel, config, isupport, message};
+use crate::{Server, channel, config, message};
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
 #[serde(untagged)]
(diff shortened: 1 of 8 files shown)


casperstorm (approved):
Looks good to me, @tarkah do you mind taking a look as well?

data/src/buffer.rs
```diff
@@ -357,6 +359,41 @@ impl From<SkinTone> for emojis::SkinTone {
     }
 }
 
+#[derive(Clone, Copy)]
+pub struct Previews<'a> {
+    collection: &'a preview::Collection,
+    cards_are_visible: bool,
+    images_are_visible: bool,
+}
```
tarkah, line 367: Nit: this could live in preview module

data/src/buffer.rs
```diff
@@ -357,6 +359,41 @@ impl From<SkinTone> for emojis::SkinTone {
...
+
+impl<'a> Previews<'a> {
+    pub fn from(
+        collection: &'a preview::Collection,
+        target: &Target,
+        config: &config::Preview,
+        casemapping: isupport::CaseMap,
+    ) -> Previews<'a> {
```
tarkah, line 375: Nit: I wouldn't use from as name since it collides with From implementation. Maybe just `new`

data/src/buffer.rs
andymandias: Makes sense to me!  Can't remember why I put it here.

PR #921, opened 2025-04-14
PRs quality improvement

KaiKorla:
This PR should helps to improve the quality in development and future PRs:

- Adding a starting .editorconfig
- Adding MegaLinter as additional PR check. It includes four checks for now:
  - EditorConfig, which checks all (applicable) files against the .editorconfig
  - Rust Clippy, which using Clippy to check Rust code files
  - CSpell, which checks spelling in Markdown files
  - Lychee, a link checker written in Rust, which checks local links in Markdown files

The GH-workflow will add a comment to each PR with the status of the checks. For example

image

Hope this helps to improve quality before a PR is merged. **Feedback welcome!**

PR Checklist

- [x] fix typos in documentation #920 
- [x] fix additional typos #924
- [x] fix CSPELL errors / update word list in cspell.json
- [x] fix (final) line endings / newline
- [x] fix line length
- [x] fix Clippy

PR #930, opened 2025-04-17
Update `rustfmt` Settings

andymandias:
This is a companion PR to PR #921;  intended to help facilitate styling contributions to meet style guidelines.  First commit contains all changes to `rustfmt` settings and the second the application of those settings.

- Updates `edition` to the latest stable version (`2024`) and introduces a matching `style_edition` as recommended in (`rustfmt`'s README)[https://github.com/rust-lang/rustfmt?tab=readme-ov-file#style-editions].
- Adds `max_width` as discussed in #921.
- It looks like #921 wants to check line endings as well?  I think that can be handled by `git` itself, but I may be mistaken.  If I am mistaken, then we can add `newline_style = "Unix"`.
- The unstable settings `group_imports` and `imports_granularity` were already present, so I added `unstable_features = true` to support them.  (Commented those options to make it clear why `unstable_features` has been set.)  In order to utilize those unstable features the [nightly version of rustfmt](https://github.com/rust-lang/rustfmt?tab=readme-ov-file#on-the-nightly-toolchain) is needed.

Additional settings can be found [here](https://rust-lang.github.io/rustfmt/?version=v1.8.0).

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,7 +1,10 @@
 fn main() {
     #[cfg(windows)]
     {
-        let _ = embed_resource::compile("assets/windows/halloy.rc", embed_resource::NONE);
+        let _ = embed_resource::compile(
+            "assets/windows/halloy.rc",
+            embed_resource::NONE,
+        );
         windows_exe_info::versioninfo::link_cargo_env();
     }
 }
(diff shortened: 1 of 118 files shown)


andymandias:
Note: The **vast** majority of the changes in this PR are from simply running `rustfmt`. I thought including those in a distinct commit would be helpful to illustrate what the `rustfmt` settings update would produce. They will be easy to re-do if we want to change the `rustfmt` settings. And, all of those changes can easily be dropped in favor of a more gradual change.

casperstorm (approved):

tarkah:
Nice, let's do it!

tarkah (approved):

Issue #564, opened 2024-09-14
WHOIS <nick> <nick> to get info from remote server


steinex:
Hi,

it would be good to support `/whois <nick> <nick>` which would get whois information from the server the whois'd person is on. That would give more information like idletime.

PR #935, opened 2025-04-17
WHOIS command now accept optional server parameter

casperstorm:
Fixes https://github.com/squidowl/halloy/issues/564.
WHOIS command now accepts optional server parameter. See: https://modern.ircdocs.horse/#whois-message.

I also re-did the command-list. I didn't like how we essentially had two of each, it was very error prone.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Added:
 
 - Ability to show a modal prompt before opening a URL
+- WHOIS command now accepts optional server parameter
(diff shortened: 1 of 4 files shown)


andymandias (approved):
Works well in my testing.  I rather like the restructuring to have the command list be one location (and the associated combination of the static and functional forms of commands).

There's a potential bit of confusion in the tooltip because `server` in the command context means something different than what it means in the general Halloy context.  E.g. a user might think the argument would let them `/whois` on OFTC from Libera.  I don't know that I have a particularly good solution though.  Maybe `server_node` instead of `server`?

PR #937, opened 2025-04-18
Characters Until Emojis

andymandias:
Adds a setting to control the number of characters until the emoji picker is shown.  Changes the previous default value from 1 to 2 to avoid showing the emoji picker for most emoticons.


casperstorm (approved):
Perfect

Issue #774, opened 2025-02-14
Feature request: Menu option to open config file


Lancelotbronner:
Would make it easier to quickly make changes on systems where it's not easily accessible!

casperstorm:
You can open the directory with command bar by pressing CTRL + K. Is this what you are looking for?
We could potentially also look at directly opening the config.toml with the default editor set for .toml files.

Lancelotbronner:
I guess it is, I hadn't thought of opening the command bar! Maybe more of a discoverability issue then.

Thanks!

marcelo4768:
seria posible agregar el comando ignore al menu contextual para elegirlo al hacer clic derecho sobre un nick ?

PR #939, opened 2025-04-18
Option to open config file in default editor

casperstorm:
Added a way to open config file from command bar and sidemenu buttons.
Fixes https://github.com/squidowl/halloy/issues/774

Screenshot 2025-04-18 at 18 05 08


andymandias (approved):
Works well for me.  Icons look nice! :slightly_smiling_face:  My only note is maybe "Open config file in default editor" should have higher ranking than "Open config directory" in the command bar.

PR #940, opened 2025-04-18
Correctly wrap urls if they are too long

casperstorm:
There was a problem with my recent PR which adds URL prompt. If the URL was too long, it would clip the view.
This is now fixed with word breaking.

Screenshot 2025-04-18 at 21 27 08


andymandias (approved):
Looks good to me, works as advertised over here in Linux land as well.

PR #922, opened 2025-04-15
CTCP requests

casperstorm:
Fixes #913

PR #943, opened 2025-04-19
CTCP Tweaks

andymandias:
Three tweaks/additions to the CTCP functionality added in #922:

- Expand the tooltip information for CTCP requests.  Required reworking completion logic a bit to allow subcommands to be the second or later argument in a command.
- Adds configuration options to disable the `PING`, `SOURCE`, `TIME`, and `VERSION` automated responses.  `TIME` in particular is the one I wanted to add a configuration option for, just added the other three since they were straightforward to implement.
- Shows all sent CTCP requests.  In my testing not all targets' clients would respond, and when they didn't respond it wasn't clear if the request had been sent or not.  Added some styling to indicate whether a CTCP message is outgoing or incoming.  Completely open to changing the styling.

### DIFF
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -22,6 +22,7 @@
 - [Configuration](configuration/README.md)
   - [Actions](configuration/actions.md)
   - [Buffer](configuration/buffer.md)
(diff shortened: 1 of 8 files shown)


casperstorm:
Great addition!!

PR #946, opened 2025-04-20
Fix final line in assets/fontello/config.json

KaiKorla:
Should be the last (for now) to get #921 green.


casperstorm:
This file auto-created when exporting new icons from https://fontello.com/ - thus i don't want to manually edit it.

KaiKorla:
Will add this file to the ignore list.

PR #949, opened 2025-04-21
Fix #701

### LINKED ISSUE #701: Replace pane not working
If I click on a buffer in the sidebar, then the pane won't swap when they are maximized. ~~This happens _only_ on Windows. It's working as intended on MacOS.~~

/edit
It seems it happens often, but not all the time. Strange somehow.

KaiKorla:
Trying to fix #701. As i'm not a Rust pro, tell me if the code sufficient. tyvm.

### DIFF
--- a/src/screen/dashboard.rs
+++ b/src/screen/dashboard.rs
@@ -2117,11 +2117,6 @@ impl Dashboard {
         window: window::Id,
         pane: pane_grid::Pane,
     ) -> Task<Message> {
-        if self.is_pane_maximized() && window == self.main_window() {
-            self.panes.main.restore();
-            self.panes.main.maximize(pane);
-        }
-
         if (self.focus != Focus { window, pane }) {
             self.focus = Focus { window, pane };
 
@@ -2131,6 +2126,11 @@ impl Dashboard {
                 self.focus_history.push_front(pane);
 
                 self.focus_history.truncate(FOCUS_HISTORY_LEN);
+
+                if self.is_pane_maximized() {
+                    self.panes.main.restore();
+                    self.panes.main.maximize(pane);
+                }
             }
         }
 
--- a/src/screen/dashboard.rs
+++ b/src/screen/dashboard.rs
@@ -611,21 +611,7 @@ impl Dashboard {
                         None,
                     ),
                     sidebar::Event::Focus(window, pane) => {
-                        let task = self.focus_pane(window, pane);
-
-                        let new_buffer = {
-                            let state_ref = self.panes.get(window, pane).unwrap();
-                            state_ref.buffer.clone()
-                        };
-
-                        if let Some(maxed) = self.panes.main.maximized() {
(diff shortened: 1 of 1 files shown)


src/screen/dashboard.rs
```diff
@@ -611,7 +611,21 @@ impl Dashboard {
...
+                        };
+
+                        if let Some(maxed) = self.panes.main.maximized() {
+                            if let Some(state) = self.panes.get_mut(window, maxed)
+                            {
+                                state.buffer = new_buffer;
+                            }
+                        }
+
+                        (task, None)
```
andymandias, line 628: This event is for focusing a pane that is already open.  The other buffers are still open, but they're hidden by the maximized pane (thank you for re-explaining this to me in the issue).  I think we should leave the `sidebar::Event::Focus` as it was, since it is calling the appropriate function (`focus_pane`).  Instead, we can modify `focus_pane` such that the maximization state is transferred along with focus.

Inside `focus_pane`, if the focus is actually changing, then we should:
- Check whether a pane is maximized using `is_pane_maximized`.
- Check whether the pane gaining focus is in the main window.
- If both of those are true, then call `self.panes.main.restore();` followed by `self.panes.main.maximize(pane);`

I believe that should do the trick.

KaiKorla: Restored it and modified focus_pane() as suggested.

src/screen/dashboard.rs
```diff
@@ -2129,6 +2129,11 @@ impl Dashboard {
             }
         }
 
+        if self.is_pane_maximized() && window == self.main_window() {
+            self.panes.main.restore();
+            self.panes.main.maximize(pane);
+        }
+
```
andymandias, line 2136: Looks good!  And works well in my testing.  We only need to do this if the focus changes (sometimes the focused pane is re-focused, and we can skip some actions then).  So, we should be able to move this up into the `if (self.focus != Focus { window, pane }) {` just above it.

src/screen/dashboard.rs
```diff
@@ -2117,6 +2117,11 @@ impl Dashboard {
         window: window::Id,
         pane: pane_grid::Pane,
     ) -> Task<Message> {
+        if self.is_pane_maximized() && window == self.main_window() {
+            self.panes.main.restore();
+            self.panes.main.maximize(pane);
+        }
+
```
andymandias, line 2124: This will still execute for every call to `focus_pane` as well.  To only execute on focus changes it should be after the `if (self.focus != Focus { window, pane }) {` and before the corresponding `}`.  In other words, it can go immediately after `self.focus_history.truncate(FOCUS_HISTORY_LEN);`.  And once it's there the `window == self.main_window()` conditional will be satisfied, so it can be dropped.

andymandias (approved):
Looks good!  Works well in my testing.  Thanks for taking this on.

PR #950, opened 2025-04-21
Image preview action

casperstorm:
This adds a new configuration option to control what action happens when pressing a image:

```toml
[preview.image]
action = "open-url" # or "preview"
```

I've added the `preview` as the new default, but let me know if others disagree.

Screenshot 2025-04-21 at 16 54 04-min


src/modal/connect_to_server.rs
```diff
@@ -17,7 +17,11 @@ pub fn view<'a>(raw: &'a str, config: &config::Server) -> Element<'a, Message> {
                 "Accept invalid certificates",
                 config.dangerously_accept_invalid_certs,
             )
-            .on_toggle(Message::DangerouslyAcceptInvalidCerts),
+            .on_toggle(|toogle| {
+                Message::ServerConnect(
+                    super::ServerConnect::DangerouslyAcceptInvalidCerts(toogle),
+                )
+            }),
```
andymandias, line 24: Minor typo here I think, `toogle` instead of `toggle`.

andymandias (approved):
This works great.  I have one nit, which is that closing the preview with Escape will also close the emoji/command picker (if they were open before clicking on an image).  I think some logic needs to be added to the `Escape` match inside `handle_event` in `src/screen/dashboard.rs`, but I'm not sure what so I'll leave that to you :sweat_smile:

src/modal/connect_to_server.rs
casperstorm: Typo resolved. Thanks.

PR #952, opened 2025-04-22
Autocomplete by Last Seen

andymandias:
Adds `order_by_recency` setting (`true` by default) that makes autocomplete match users based first on how recently they were last seen in the channel (first, then alphabetically ascending/descending based on `sort_direction` for users that have not been seen in the channel).

Also changes matching to use server normalization (rather than Unicode case insensitivity).

### DIFF
--- a/book/src/configuration/buffer.md
+++ b/book/src/configuration/buffer.md
@@ -658,9 +658,28 @@ auto_format = "markdown"
 
 Customize autocomplete.
 
+#### `order_by`
+
+Ordering that autocomplete uses to select from matching users.
+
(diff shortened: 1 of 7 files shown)


book/src/configuration/buffer.md
```diff
@@ -684,6 +684,19 @@ Sets what suffix is added after autocompleting. The first option is for when a n
...
+
+Autocomplete users by their last message in the channel;  the user with most recent message autocompletes first, then increasing older messages.  Users with no seen messages are matched last, in the order specified by `sort_direction`.
+
+```toml
+# Type: bool
+# Values: true, false
+# Default: true
+
+[buffer.text_input.autocomplete]
+order_by_recency = true
```
casperstorm, line 697: Nit; I wonder if we should future proof and make it an enum?

```enum Order
  Recency
```

casperstorm (approved):
This is awesome @andymandias

PR #954, opened 2025-04-23
Exempt WHO Polls from Rerouting

andymandias:
In testing a quirk with `/whois` I found myself receiving rerouted replies for an automated WHO poll that was occurring at the same time as testing.  This PR is to add some additional logic in order to exclude automated WHO polls from rerouting (from being rerouted as well as starting/stopping rerouting).

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -250,6 +250,60 @@ impl Client {
         }
     }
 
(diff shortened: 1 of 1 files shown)


casperstorm (approved):
Make sense to me. Code is clean 👍🏻

PR #956, opened 2025-04-24
Setting to Mark as Read when Sending a Message

andymandias:
Adds an option to mark a buffer as read when sending a message.  Required a bit more plumbing than I was hoping for, but seems to work in my testing.

### DIFF
--- a/src/buffer/input_view.rs
+++ b/src/buffer/input_view.rs
@@ -3,10 +3,12 @@ use std::time::Duration;
 
 use data::buffer::Upstream;
 use data::dashboard::BufferAction;
+use data::history::{self, ReadMarker};
 use data::input::{self, Cache, Draft};
+use data::message::server_time;
 use data::target::Target;
 use data::user::Nick;
-use data::{Config, client, command, history};
+use data::{Config, client, command};
 use iced::Task;
 use iced::widget::{column, container, text, text_input};
 use tokio::time;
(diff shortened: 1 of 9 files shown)


casperstorm:
Looks great, and make sense to have it on by default. Perhaps @tarkah can go through it as it touches a lot he is used to look at.

andymandias:
Realized I missed one case where `MARKREAD` needs to be set.  Will re-open after adding.

src/buffer/input_view.rs
```diff
@@ -446,19 +472,26 @@ impl State {
...
-                                )
+                            record_message_tasks
                                 .into_iter()
-                                .map(Task::future),
+                                .map(Task::future)
+                                .chain(
+                                    update_read_marker_tasks
+                                        .into_iter()
+                                        .map(Task::future),
+                                ),
```
tarkah, line 494: Why not chain these when returning from the original function if they don't need to be handled separately?

andymandias: I originally tried to return a single `Vec`, but I couldn't get the compiler to accept the two `Future`s in the same `Vec` (and some of what I looked up trying to get it to work said it's impossible).  I had assumed these iterators could be chained because they're mapped into `Task::future`, but `Task` isn't available in `iced_core` as far as I could tell.  Wasn't sure we wanted to introduce `iced` as a dependency for `data`, so I went with this.

andymandias: Don't know why I listened to those that said impossible, after thinking on a bit I realized there are a couple ways to do it at least.  Went with boxing as that seemed like the simplest.

casperstorm (approved):

PR #957, opened 2025-04-25
Attempt to Parse Invalid UTF-8

andymandias:
Minor change to still attempt to parse a message even if it's invalid utf-8 via `String::from_utf8_lossy`.  That way there's a chance the message could make it in front of the user, even if malformed, rather than only being present in the logs where it may go unnoticed.


casperstorm (approved):
Looks good to me.

PR #959, opened 2025-04-25
Fix for Panic in Message Parsing

andymandias:
Avoids a potential panic caused by slicing on `u8` indices instead of character indices.  The offending string has been added to the parsing test cases.  Originally implemented using `str::char_indices`, but I think this is cleaner.


casperstorm:
I would love if @tarkah could take a look at this. Great job addressing this @andymandias.

tarkah (approved):

PR #960, opened 2025-04-27
Settings to Only Read First Line of Password Files

andymandias:
Adds a setting to read the first line of a password file only (ignoring the first newline and everything past it).  On by default since many text editors will automatically insert a trailing newline (and very few passwords tend to include a newline, as far as I know).


casperstorm (approved):
That was quick :D

PR #961, opened 2025-04-29
servers.md: Add note about changing to plain text ports.

RealKindOne:
Someone might use a plaintext port and not know or forget they need to change the ```use_tls``` setting. Add a note about.


casperstorm (approved):
Make sense to add. Thanks.

PR #965, opened 2025-05-01
Add `/query` as Alias for `/msg`

andymandias:
Add `/query` as an alias for `/msg` for users that might be familiar with that name from other clients.


casperstorm (approved):
:shipit:

PR #972, opened 2025-05-02
Display Received WALLOPS Messages

andymandias:
Displays received WALLOPS messages in the server buffer in the format:
```
WALLOPS from nickname: message text here
```

I left `WALLOPS` in all caps to try to signal it's an IRC command to those who might not be familiar (rather than the English word "wallop"), but I'm open to different styling.

Gets its own color under the `server_messages` colors table, and takes on the default server message coloring if not specified.

It will trigger the unread status for the server buffer (similar to standard replies and monitor online/offline messages).  I.e. receiving a WALLOPS will color the server globe icon to indicate unread.

Tested on a local inspircd instance.


casperstorm (approved):
Oh nice, great addition

Issue #969, opened 2025-05-01
Clicking menu (lower left) draws directly over click point


darienm:
Clicking the menu icon lower-left draws the menu directly over the click point, which means clicking again _doesn't_ hide it. This is not typically the expected behavior for menu-type interfaces. Suggest adjusting the offset so that click-to-show, click-to-hide works, while still permitting access to the menu items.

PR #973, opened 2025-05-02
Initial fix for #969

KaiKorla:
I tried fixing #969. I'm ok with the sidebar context menu, but unsure with the buffer context menu. Please take a look and give me some feedback. tyvm.

### DIFF
--- a/src/widget/context_menu.rs
+++ b/src/widget/context_menu.rs
@@ -27,7 +27,7 @@ pub enum Anchor {
 }
 
 #[derive(Debug, Default, Clone, Copy)]
-pub enum ToggleBehavior  {
+pub enum ToggleBehavior {
     #[default]
     KeepOpen,
     Close,
@@ -51,7 +51,6 @@ pub fn context_menu<'a, T, Message, Theme, Renderer>(
         },
         anchor,
         toggle_behavior,
-
         menu: None,
     }
 }
@@ -63,7 +62,6 @@ pub struct ContextMenu<'a, T, Message, Theme, Renderer> {
     activation_button: iced::mouse::Button,
     anchor: Anchor,
     toggle_behavior: ToggleBehavior,
-
     // Cached, recreated during overlay if menu is open
     menu: Option<Element<'a, Message, Theme, Renderer>>,
 }
@@ -77,7 +75,7 @@ pub struct State {
 impl State {
     pub fn new() -> Self {
         State {
-            status: Status::Closed,
+            status: Status::Closed(false),
             menu_tree: widget::Tree::empty(),
         }
     }
@@ -85,14 +83,14 @@ impl State {
 
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub enum Status {
-    Closed,
+    Closed(bool),
     Open(Point),
 }
 
 impl Status {
     pub fn open(self) -> Option<Point> {
         match self {
-            Status::Closed => None,
+            Status::Closed(_) => None,
             Status::Open(position) => Some(position),
         }
     }
@@ -154,7 +152,7 @@ where
 
     fn state(&self) -> tree::State {
         tree::State::new(State {
-            status: Status::Closed,
+            status: Status::Closed(false),
             menu_tree: widget::Tree::empty(),
         })
     }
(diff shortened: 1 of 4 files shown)


KaiKorla:
Discussion @ #halloy:

x offset + 5

image

image

image

casperstorm:
> Discussion @ #halloy:
> 
> x offset + 5
> image image image

I think that looks good.

casperstorm:
Sorry for late review on this, @KaiKorla.
This is good, however a small new problem. When you click the sidebar-button it opens and looks great, however if you then instantly press the button again it actually doesn't close it but close then re-opens. Which is techinically what we wanted to fix initially :D

casperstorm:
I just pushed a commit with some name changes, hope that was ok.
Last issue you could address, if you right click a user in a message, it will never close that context menu.

KaiKorla:
It's working so far, but the problem with "inline" nicknames in the buffer consists. I tried to track it down, but with my limited experience in Rust and iced it is not so easy.
IMO the problem with this is, that the state cannot be found for this text-parts and therefore the implementation in my code get always a ```Status::Closed``` on ```tree.state.downcast_mut::<State>();```. I took a look into [selectable_rich_text.rs](url) and wonder about the handling of context_menu there. I don't know if this will cause problems. Maybe i'm totally wrong, but could need some advice anyhow.

@tarkah: @casperstorm told me i should mention you for any advice/help/ideas. so if you got some time please take a look. thank you very much.

casperstorm:
This is not forgotten, but we will need @tarkah to chime in here. Just FYI @KaiKorla.

src/widget/context_menu.rs
```diff
@@ -177,43 +197,67 @@ where
...
+                |
+                (true, Status::Closed, ToggleBehavior::Close, Some(position))
+                => Status::Open(position),
+
+                (false, Status::Open(_), _, None)
+                |
+                (true, Status::Open(_), _, None)
+                |
+                (true, Status::Open(_), ToggleBehavior::Close, Some(_))
+                => Status::Closed,
```
andymandias: We'd want to include `(_, Status::Closed(true), _, _)` in this arm, and have `next_status` set to `Status::Closed(false)`.

src/widget/context_menu.rs
```diff
@@ -510,24 +554,6 @@ where
...
-        {
-            if cursor.position_over(layout.bounds()).is_some() {
-                self.state.status = Status::Closed;
-            }
-        }
-
-        if matches!(self.state.status, Status::Closed) {
-            shell.request_redraw();
-        }
-
```
andymandias: I believe you removed this in order to avoid closing the context menu as part of the click, only to have the underlying widget re-open the menu (as it now sees it as closed).  As you discovered, the way `selectable_rich_text.rs` utilizes `context_menu` expects this behavior.  To avoid messing with `selectable_rich_text.rs`, which we would want to bring in a consult for, we can instead keep this code add a `bool` to `Status::Closed` like:
```rust
pub enum Status {
    Closed(bool),
    Open(Point),
}
```
which can be initialized as `Status::Closed(false)`.  Then, this code can become
```rust
        if let Event::Mouse(mouse::Event::ButtonPressed(_)) = &event {
            if cursor.position_over(layout.bounds()).is_none() {
                self.state.status = Status::Closed(true);
            }
        }

        if let Event::Mouse(mouse::Event::ButtonReleased(mouse::Button::Left)) =
            &event
        {
            if cursor.position_over(layout.bounds()).is_some() {
                self.state.status = Status::Closed(true);
            }
        }
```

src/widget/context_menu.rs
```diff
@@ -177,43 +197,67 @@ where
...
+            let next_status = match (
+                is_activation_mouse_event,
+                prev_status,
+                self.toggle_behavior,
+                position,
+            ) {
+                (true, _, ToggleBehavior::KeepOpen, Some(position))
+                |
+                (true, Status::Closed, ToggleBehavior::Close, Some(position))
+                => Status::Open(position),
```
andymandias: We'd want to have the second expression in this arm only match against `Status::Closed(false)`, preventing the close→immediately-reopen issue.

src/widget/context_menu.rs
```diff
@@ -177,43 +197,67 @@ where
...
+                |
+                (true, Status::Open(_), _, None)
+                |
+                (true, Status::Open(_), ToggleBehavior::Close, Some(_))
+                => Status::Closed,
+                _ => prev_status, // keep status
+            };
+
+            if next_status != prev_status {
+                state.status = next_status;
```
andymandias, line 261: We should maybe check that `!matches!(next_status, Status::Closed(_)) || !matches!(prev_status, Status::Closed(_))` before requesting a redraw.

andymandias (commented):
Apologies for the delay.  I think we can make a few changes that will avoid touching the more complicated bits of the code, and make this PR easier to land.  I've given the rough outline for the strategy in code comments.  Let me know if further details would be helpful.

andymandias:
Helping get this over the line a bit, thank you for getting this most of the way complete @KaiKorla.  Works as intended on Linux for me now.  @casperstorm, when you have a chance could you confirm it's working on Mac/Windows as well?

casperstorm (approved):
This works great. Confirmed working on my end.
Thanks both @andymandias and @KaiKorla for this!

Issue #929, opened 2025-04-16
Font Renders Very Thin


c-alpha:
I have set a different font than Iosevka. It is rendered very thin in halloy as in the below screenshot.

![Image](https://github.com/user-attachments/assets/48cf40ac-d92b-43c9-be18-59e9778f8dcc)

The next screenshot shows the same font at the same size, and with the same colours. Note how the font is rendered differently.

![Image](https://github.com/user-attachments/assets/cd61499a-9d70-4d92-baf3-a68293194527)

The font looks the same as in the second screenshot in all other applications on my system. So the different rendering is in halloy only. 

It seems as if in halloy it were anti-aliased twice perhaps?

This is on macOS, and the font is Menlo at 12pt.

darienm:
Have you tried adjusting the scale_factor value in config.toml? When  was tweaking font settings (also on macOS) I found the best results with a scale_factor = 1.12 (on my 4k display) or 1.22 on my smaller display. The changes are immediately visible after saving config and reloading configuration while Halloy is running. My font size is set to 15, but changing it means quitting and relaunching Halloy to see each change.

c-alpha:
Hello @darienm, yes I have tried increasing the scale factor. When all UI elements are scaled up, the font lines become wider too, of course, but only as a side-effect.

I'd like to keep my font at 12pt, and get it rendered as in all other apps at that size, too, and maintain the UI elements at their default size, please.

darienm:
In your preferred screen shot, the numeral "1" is 18 pixels tall, but in Halloy it is 17 pixels tall. Despite specifying 12pt for each program, there is a 5.8% difference in scale that could be accounting for the appearance change. It would seem that somewhere in the Halloy environment the text windows are being scaled down a touch by default.

c-alpha:
Good point, about the scaling. It had seemed a _little_ smaller to me, too, but unlike you I hadn't made any measurements. The perceived thinness (perhaps amplified by the slightly blurrier edges?) could well be the result of some software scaling in the pixel domain.

Do you observe the same scaling on your platform?

darienm:
I'm also on macOS, and after spending "enough" time with the same issue, went with the scale_factor adjustment and am quite happy with the results, even across different displays with different pixel densities. Maybe try between 1.12 and 1.22 to find your sweet spot.

PR #975, opened 2025-05-03
Font Weight Configuration Option

andymandias:
Adds the ability to specify font weights, for variable weight fonts.  A configuration for the user to have some control of the perceived font weight outside of scaling tricks, to help with issues like #929 (which otherwise I believe Halloy has little control over).  E.g. for Source Code Pro normal:
Screenshot from 2025-05-03 11-35-22
Light:
Screenshot from 2025-05-03 11-35-49
And Medium:
Screenshot from 2025-05-03 11-34-52
Currently it takes the font weight 3 steps higher than the configured weight as the bold font weight.  And, not all fonts have a good range of weights unfortunately ~~(e.g. extra-bold in Source Code Pro apparently is automatically condensed?)~~.

Edit: After double-checking, it looks like the Source Code Pro repository does not have an extra-bold variant, which is why the text looks condensed in the Medium example.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Added:
 - `/hop` command. `/hop` parts the current channel and joins a new one
 - Settings to limit passwords read from a file to the first line of the file only (on by default)
 - Receive `WALLOPS` messages in the server buffer (color configurable in themes)
+- Configuration option for font weight
 
 Changed:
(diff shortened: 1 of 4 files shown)


andymandias:
An example screenshot for Semibold, since the Medium example I picked did not have the correct weight installed to demonstrate bold text:
Screenshot from 2025-05-03 17-41-32

casperstorm (approved):
This is great, Andy.

PR #985, opened 2025-05-08
Fix handling of RPL_NOWAWAY & RPL_UNAWAY

KaiKorla:
RPL_NOWAWAY & RPL_UNAWAY are no sending the nickname back. There is just a custom message attached by the server, for example "You have been marked as being away" or "You are no longer marked as being away".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+Fixed:
+
+- Handling of RPL_NOWAWAY & RPL_UNAWAY to reflect users own AWAY state
+
 # 2025.5 (2025-05-05)
 
 Added:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 Fixed:
 
-- Handling of RPL_NOWAWAY & RPL_UNAWAY to reflect users own AWAY state
+- Handling of RPL_NOWAWAY & RPL_UNAWAY to reflect user's own AWAY state
 
 # 2025.5 (2025-05-05)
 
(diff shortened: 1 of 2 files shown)


data/src/client.rs
```diff
@@ -1777,29 +1777,23 @@ impl Client {
                     }
                 }
             }
-            Command::Numeric(RPL_UNAWAY, args) => {
-                let nick = ok!(args.first()).as_str();
-                let user = User::try_from(nick)?;
+            Command::Numeric(RPL_UNAWAY, _) => {
+                let user = User::try_from(self.nickname().as_ref())?;
```
andymandias, line 1781: Here and for `user` in `RPL_NOWAWAY` below we can use
```rust
                let user = User::from(self.nickname().to_owned());
```
to avoid the extra processing done as part of `try_from`.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Unreleased
 
+Fixed:
+
+- Handling of RPL_NOWAWAY & RPL_UNAWAY to reflect users own AWAY state
```
andymandias, line 5: Minor typo, I think.  I believe you wanted:
```md
- Handling of RPL_NOWAWAY & RPL_UNAWAY to reflect user's own AWAY state
```

andymandias (approved):
Looks good to me.  Might be worth adding a short comment by the `RPL_UNAWAY`/`RPL_NOWAWAY` handling that according to [horse docs](https://modern.ircdocs.horse/#away-message) `RPL_NOWAWAY` and `RPL_UNAWAY` always apply to the client itself (i.e. `away-notify` uses `AWAY` to notify re: other users).

casperstorm (approved):

PR #987, opened 2025-05-12
`/delay <seconds>` Command

### LINKED ISSUE #677: Unable to automatically join invite-only channels on libera: a delay command would help
I want to auto-join an invite-only channel on libera.chat . In order to do this, I need to identify to the nickserv (I do that through SASL), then /msg ChanServ invite #channel , then /join #channel. If I put these commands into on_connect, the join command gets executed too quickly and I'm rejected from joining the channel. It would help if I could add a small delay between the two commands.

andymandias:
Closes #677 by adding a `/delay <seconds>` command to help users control the timing of on-connect commands.


kasper93:
Wouldn't it be better to allow callbacks on invite messages? Instead of adding sleep() synchronization, which will always be flakey.

casperstorm:
I think this is a good approach, and similar to what other clients have.
I don't think we need it now, but i've seen others have: `/timer [repeats] [interval-in-seconds]` so you can do repeats as well. Could be an addition for the future.

tarkah (commented):
Redacted

andymandias:
Restructured things a bit to move the on-connect logic from `main` to `dashboard`.  One of my UX goals was to have **all** commands executed in-order;  i.e. the internal commands would also be handled in-order (and affected by `/delay`).  I don't have a personal use case for delaying those actions, but I think providing that makes on-connect easier for a user to understand (i.e. no special cases where some commands get to skip the line and execute first).  I found it difficult/convoluted to try and provide that behavior in `client` or via an async loop, so that's why it's where it is now.

casperstorm (approved):
Awaiting @tarkah last go.

andymandias:
Thanks for the review @tarkah!  And thank you for the help with restructuring, the divide between data and view is definitely improved.

PR #989, opened 2025-05-12
Extended Theme Settings

andymandias:
This PR draft is for adding the ability to specify font style for some theme settings.  For example, if you wanted to display nicknames in italic you could have in your theme TOML:
```toml
nickname = { color = "#F6B6C9", font-style = "italic" }
topic = { color = "#AB8A79", font-style = "bold-italic" }
```

With the result potentially looking like this:
Screenshot from 2025-05-11 21-30-14

There is more work that would need to be done, but I wanted to get feedback before progressing further.

A lot of the changes are due to changing names (e.g. `Colors` to `Styles`) since theme settings can include font-style as well as color.  The main changes are in:
- `data/src/appearance/theme.rs`
- `src/appearance/theme/font_style.rs`
- `src/font.rs`
- `src/widget/selectable_rich_text.rs` 
- `src/widget/selectable_text.rs`

Changes to apply font styles in:
- `src/widget/message_content.rs`
- `src/buffer/channel.rs`
- `src/buffer/channel/topic.rs`
- `src/buffer/highlights.rs`
- `src/buffer/logs.rs`
- `src/buffer/query.rs`
- `src/buffer/server.rs`
- `src/buffer/user_context.rs`

And the rest are just name changes.

### DIFF
--- a/data/src/appearance/theme.rs
+++ b/data/src/appearance/theme.rs
@@ -6,7 +6,7 @@ use palette::rgb::{Rgb, Rgba};
 use palette::{FromColor, Hsva, Okhsl, Srgba};
 use rand::prelude::*;
 use rand_chacha::ChaChaRng;
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use thiserror::Error;
 use tokio::fs;
 
@@ -17,28 +17,28 @@ const DEFAULT_THEME_CONTENT: &str =
 #[derive(Debug, Clone)]
 pub struct Theme {
     pub name: String,
-    pub colors: Colors,
+    pub styles: Styles,
 }
 
 impl Default for Theme {
     fn default() -> Self {
         Self {
             name: DEFAULT_THEME_NAME.to_string(),
-            colors: Colors::default(),
+            styles: Styles::default(),
         }
     }
 }
 
 impl Theme {
-    pub fn new(name: String, colors: Colors) -> Self {
-        Theme { name, colors }
+    pub fn new(name: String, styles: Styles) -> Self {
+        Theme { name, styles }
     }
 }
 
 // IMPORTANT: Make sure any new components are added to the theme editor
 // and `binary` representation
 #[derive(Debug, Clone, Copy, Serialize, Deserialize)]
-pub struct Colors {
+pub struct Styles {
     #[serde(default)]
     pub general: General,
     #[serde(default)]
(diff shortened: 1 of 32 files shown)


casperstorm:
This is awesome, and you can easily continue down this path here!
I think we should consider handling font styling in Theme Editor.

This is gonna be an awesome addition.

src/font.rs
```diff
@@ -119,3 +120,12 @@ pub fn width_from_chars(len: usize, config: &config::Font) -> f32 {
...
 }
+
+pub fn get(font_style: Option<FontStyle>) -> Option<Font> {
+    font_style.and_then(|font_style| match font_style {
+        FontStyle::Normal => None,
+        FontStyle::Bold => Some(MONO_BOLD.clone()),
+        FontStyle::Italic => Some(MONO_ITALICS.clone()),
+        FontStyle::ItalicBold => Some(MONO_BOLD_ITALICS.clone()),
+    })
+}
```
tarkah: We're making assumptions about how the caller should call this by taking an `Option<_>` instead of just a `FontStyle`. Also, we can remove the return `Option` by mapping `Normal -> MONO`. I think this makes the function much more "functional" since it can be easily used w/ `Option` methods like `map` if the caller has an `Option<FontStyle>`

casperstorm (approved):

data/src/appearance/theme.rs
```diff
@@ -128,72 +128,252 @@ pub struct Buffer {
...
+        }
+    }
+}
+
+#[derive(Clone, Copy, Debug, Default)]
+pub struct OptionalTextStyle {
+    pub color: Option<Color>,
+    pub font_style: Option<FontStyle>,
+}
+
```
tarkah, line 283: Can we just make this version `TextStyle` and use it both where `TextStyle` is used and where `OptionalTextStyle` is used? If someone omits `color` where we previously made it required, we can just `unwrap_or(Transparent)` and clean this up a bunch. 

Or we can go the other direction, we always make `Color` non-optional but we use `default = transparent` w/ serde so if they don't define it, we set it as `transparent`, and does that have the same effect as if it wasn't defined?

Just trying to think of ways to not need both of these 😬

data/src/appearance/theme.rs
```diff
@@ -128,72 +128,252 @@ pub struct Buffer {
...
+    {
+        #[derive(Deserialize)]
+        #[serde(untagged)]
+        enum Data {
+            Basic(String),
+            #[serde(rename_all = "kebab-case")]
+            Extended {
+                color: String,
+                font_style: Option<FontStyle>,
+            },
```
tarkah: Don't we want keys to be `snake_case`?

PR #990, opened 2025-05-12
Fix buffer.channel.topic documentation

kit-ty-kate:
The default values shown for those two variables were not matching reality (as of 2025.5). The topic banner isn’t shown by default and 2z isn’t a number (i’m guessing this was a typo)


andymandias (approved):
Thank you for catching these!

PR #993, opened 2025-05-16
Tweak User Regex

andymandias:
Removes `@` and `*` characters from the user regex, since horse docs [lists them as forbidden](https://modern.ircdocs.horse/#clients) in nicknames.

Changed, in part, to support user-highlighting in @mentions, since some users are accustomed to that style.


casperstorm (approved):
Looks good to me

PR #1000, opened 2025-05-27
Show channel mode in panel

casperstorm:


andymandias (approved):
Works well for me in my testing.  Looks good in the pane title bar.

My only note is that we might want, in a future PR, to have a setting to auto-request channel mode on join (with the response hidden) to quietly populate channel modes.

Issue #871, opened 2025-04-01
[Feature request] undo/redo support in text box


kasper93:
Undo/redo support in text box would be nice. Also going to with UP arrow to history item, should allow to go back with DOWN arrow to previously posted (not sended text).

darienm:
Coming here to say that this bites me so very frequently, as at least on macOS the up-arrow is often used to move the cursor to the start of a line, then I realize I didn't want my chat history, but too late, the message I had typed is gone.

PR #1001, opened 2025-05-28
Save Input Draft when Navigating Input History

andymandias:
Stores the current input text as a draft separate from the current text in the input (previously the draft was also the current text in the input).  The draft is stored separately to allow users to navigate input history without losing their input text.  I.e. if the user presses up to look at their sent message history (or by accident), then pressing down will restore the text that was previously in the input.

If the user modifies the text history text, then it becomes the new draft text (overwriting the previously stored draft).

Addresses secondary comment in issue #871 (not the main feature request).

### DIFF
--- a/data/src/history/manager.rs
+++ b/data/src/history/manager.rs
@@ -268,8 +268,12 @@ impl Manager {
         self.data.input.record(buffer, text);
     }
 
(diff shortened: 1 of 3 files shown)


casperstorm (approved):
Works well here.

PR #1008, opened 2025-05-31
Workflow Updates

andymandias:
This PR was originally intended to be a collection of minor tweaks to workflows, but has unexpectedly grown to be somewhat substantial.  I can split this up if need be, or remove any part that's not desirable.  The changes:
- Since the Typos action checks many files outside of the Rust source code, I moved it out of the `build.yml` workflow into its own workflow (`spell-check.yml`) that can be triggered more generally.
- Added the `book.yml` workflow to perform checks on the documentation.  In doing so noticed that `mdbook-linkcheck` was not catching all of the broken links I created to test the workflow, so I switched the link checking to Lychee (as demonstrated in #921).  One broken anchor was found by Lychee, and fixed in this PR.
- I've been using `cargo-nextest` for testing, because I find the output more legible than `cargo test`.  So that has been tentatively swapped in.
- When I added Typos I didn't pay close enough attention to the naming style that was already in use.  Minor changes to make the new workflow additions better match the existing style.

### DIFF
--- a/.config/lychee.toml
+++ b/.config/lychee.toml
@@ -1,0 +1,5 @@
+verbose = "info"
+no_progress = true
+
+include_fragments = true
+exclude = [ '^https://halloy\.squidowl\.org/.*', '^irc://.*', '^ircs://.*' ]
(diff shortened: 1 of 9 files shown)


casperstorm (approved):
Great additions!!

Issue #979, opened 2025-05-06
[Bug] server_messages = false prevents display but not indicator of new messages


darienm:
Version 2025.5. When config has buffer.server_messages.* set to enabled=false, these messages are hidden from displaying in the buffer. However, they are still triggering the ---backlog--- separator and the checkmark icon in the topic bar giving the impression that there are new messages to review. If configured to be ignored, they should probably not do this.

```
[buffer.server_messages.join]
enabled = false

[buffer.server_messages.part]
enabled = false

[buffer.server_messages.quit]
enabled = false
```

PR #1009, opened 2025-06-02
fix: New message indicator shouldn't show for suppressed message

joshuamegnauth54:
Closes: #979

I tested this with and without...

```toml
[buffer.server_messages.join]
enabled = false

[buffer.server_messages.quit]
enabled = false

[buffer.server_messages.part]
enabled = false
```

...in my config to check if the indicator shows.

Alsooo, there are two small, unrelated changes in `src/appearance/theme.rs` and `src/widget/decorate.rs` because of `cargo fmt`.

### DIFF
--- a/data/src/history.rs
+++ b/data/src/history.rs
@@ -16,7 +16,7 @@ use crate::message::{self, MessageReferences, Source};
 use crate::target::{self, Target};
 use crate::user::Nick;
 use crate::{
-    Buffer, Message, Server, buffer, compression, environment, isupport,
+    Buffer, Message, Server, buffer, compression, config, environment, isupport,
 };
 
 pub mod manager;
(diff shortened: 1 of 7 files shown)


casperstorm (commented):
Could you update CHANGELOG as well?

Edit:
After testing, @andymandias pointed out that this doesn't actually fix the issue. With` if !message.suppressed(config) && message.triggers_unread() {` the messages they've configured to be suppressed are already covered by `triggers_unread` which is `false` for server messages besides monitor messages, standard replies, and wallops.

andymandias:
Also worth noting, the visual indicators referenced in #979 are the backlog divider and the "Mark messages as read" button in the pane titlebar being enabled.  I'm not sure how we want to handle the backlog divider (its presence at the very bottom of the chat window indicates it isn't somewhere up, past the top of the current view), but the "Mark messages as read" button enabling could be addressed by modifying `can_mark_as_read` in `data/src/history.rs`.

joshuamegnauth54:
I _think_ I may have been looking at the wrong element. I'll reexamine this and see if I can fix it.

andymandias:
With #1143 merged the issue has been closed.  Thank you for taking a look at it.

Issue #271, opened 2024-03-15
Icon has Black Box in GNOME Overview


andymandias:
When viewing the icon in GNOME's overview (roughly equivalent to expose in macOS or task view in Windows) a black box appears around Halloy's icon.

![black_box](https://github.com/squidowl/halloy/assets/5768180/652de987-0959-42ae-a57a-1bcc4a9087a6)

This appears to be because GNOME is adding a shadow behind any pixel that is not completely transparent.  If I run Halloy's icon through ImageMagick to flatten everything under 17% opaque to transparent, then (after rebuilding the icon cache) GNOME is able to handle the icon correctly.

![no_black_box](https://github.com/squidowl/halloy/assets/5768180/e4ac76d2-9de1-4014-8fbe-2287550538f5)

Specifically, the command I used is:
```
convert org.squidowl.halloy.png -channel alpha -threshold 17% flattened.png && mv flattened.png org.squidowl.halloy.png
```
But I expect a more elegant solution could be made with the source image;  I'm loathe to modify the logo at all, it's so cool, but I also would rather not see it with a weird black box around it.

PR #1015, opened 2025-06-08
Added new Halloy icon

casperstorm:
The new Halloy icon created by Rune Seir (https://www.instagram.com/rseir/)

halloy_round


andymandias:
Looks great!

Two minor issues:
- `book/src/images/banner-logo.png` is still the old banner.
- Under GNOME I'm having [this issue](https://github.com/squidowl/halloy/issues/271) again, where a black box appears around the icon in the overview view.  If I recall correctly, that happens because their drop shadow algorithm relies on the image being 100% transparent for all pixels outside of the main icon.

~~I believe the failing Book check is due to a misconfiguration of mine in the workflow.  I will fix it, but for now I think it can be ignored.~~

Edit:  Actually, it appears the Book check is accurate.  The replacement for `book/src/images/banner-logo.png` should also be named `book/src/images/banner.png`, and that will take care of that :slightly_smiling_face:

andymandias:
The drop shadow issue has been resolved, currently there are small artifacts in the circle at the cardinal extrema:
Screenshot from 2025-06-09 11-43-33

Issue #443, opened 2024-07-26
Some channel mode changes lost


glguy:
halloy doesn't display the removal of most modes to a channel. In my test it couldn't show removing any of: `tnljf` but it could show removing `k`

Looking around halloy doesn't process ISUPPORT CHANMODES, so it's implementation of which modes take parameters is currently an approximation and wrong for some of the modes on libera.chat, for example. I don't know if this is connected to it dropping the mode changes or not.

andymandias:
The CHANMODES ISUPPORT parameter is currently parsed, then ignored. Utilizing all parsed ISUPPORT parameters is on my to-do list, so I'll be returning to this (unless it's taken care of before I can get to it). I'll put this higher up on that list.

PR #1016, opened 2025-06-08
Utilize `CHANMODES`, `MODES`, and `PREFIX` ISUPPORT Parameters

andymandias:
Utilizes `CHANMODES`, `MODES, and `PREFIX` ISUPPORT parameters for `MODE` parsing.  This information is used every relevant place I could think of:
- To provide as much guidance as possible in completion tooltips.
- To provide additional error checking user input `/mode` commands.
- To ensure the Voice and Operator mode letters in the context menu are accurate (and the membership roles are present on the server).

For channel modes, this now means that `CHANMODES` and `PREFIX` are the primary sources used for parsing.  They are used to determine whether a channel mode exists and whether or not it takes an argument.  For the channel membership modes, this means they are no longer hard-coded, but are provided by `PREFIX`.  I.e. the prefix meanings remain hard-coded in `irc/proto/src/lib.rs`, with the presence of those membership roles and related channel mode letter defined by `PREFIX`.

Unfortunately, no such exhaustive list is available for user modes, so the changes with respect to user modes are minimal.

Utilizes some of the style changes to ISUPPORT handling functions introduced in PR #987, so that's why it's currently based off of it.  (That is, if PR #987 gets changed, then there's a decent chance I'd want to reformulate this PR before it could be accepted.)


andymandias:
When testing I did not find any issue with showing the removal/addition of the modes mentioned in #443, so I believe this can close that issue.

data/src/appearance/theme.rs
```diff
@@ -490,7 +490,7 @@ mod binary {
         BufferServerMessagesStandardReplyFail = 38,
         BufferServerMessagesStandardReplyWarn = 39,
         BufferServerMessagesStandardReplyNote = 40,
-        BufferServerMessagesWallops = 41,
+        BufferServerMessagesWAllOps = 41,
```
tarkah, line 644: Why the weird casing?

data/src/user.rs
```diff
@@ -403,22 +403,48 @@ pub enum AccessLevel {
...
+    Debug,
+    Clone,
+    Copy,
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Ord,
+    Hash,
+    Serialize,
+    Deserialize,
```
tarkah: Do these need to be serializable? Same with `AccessLevel`? I can't think of where we serialize these. We probably want to remove this. Anytime I see a change on something serializable I want to ensure no format breakage so we should ensure either:

- This is persisted & no format breakage
- This isn't actually persisted, remove serde so it's not confusing

data/src/appearance/theme.rs
andymandias: An attempt to indicate the meaning a bit as "w-all-ops", since I believe the type can be expanded to "message sent to users with mode w and all operators".  Not attached to this since it might only be helpful if someone has explained it, just noticed that it wasn't PascalCased and took a crack at it.

data/src/user.rs
andymandias: Looks like the answer is no, they don't need to be (once we remove the same from `AccessLevel`; don't know the provenance there, but nothing complains if I remove the derives from it).

tarkah: Cool 👍 you can just omit it from your new type and maybe we follow up and remove serde from all non-persisted types in a separate unit of work

Issue #1012, opened 2025-06-03
Display server list in same order as appears in config.toml


darienm:
As of 2025.5 it appears as if the server list in the main window is not ordered alphabetically, or reflects the server list order in config.toml. It might be a good idea to offer users the opportunity to determine an order that is sensible to them. I'd suggest that maintaining order from config.toml is the best practice, with alphabetically by servers.KEYNAME as a fallback.

andymandias:
Will be in the next release, thank you for the feature request.

PR #1021, opened 2025-06-12
Sidebar Ordering

andymandias:
Addresses issue #1012 by updating the sidebar ordering and adding a configuration option to choose between lexicographical and as-in-the-configuration-file ordering.  I set the default to lexicographical ordering, since that's my intuition as to what most users will expect.


casperstorm (approved):

casperstorm:
This is a nice change!

PR #1022, opened 2025-06-12
Eliminate common deep clones

4e554c4c:
This commit changes several structures, primarily `Server`, `Target` and `config::Server` to be wrapped in an Arc. There should be little change to existing code, but this should dramatically reduce the number of deep copies of these structs. This is rather important since `Server` needs to be cloned on every message received from the network to be sent across threads.


4e554c4c:
thanks! :smile:

PR #1029, opened 2025-06-16
feat: python implementation

user0-07161:
TBD:
- full HexChat API coverage


4e554c4c:
Is there any reason to prefer python over something lighter like lua?

I think pulling in an entire python interpreter is a bit much

user0-07161:
@4e554c4c lua requires us to have lua C libraries on the device we're building on...

4e554c4c:
It looks like the `vendored` feature of `mlua` supports not linking to system libraries

user0-07161:
huh? i'll look at that

user0-07161:
@4e554c4c have you tried using it? the size of the binary should be noticeably higher

user0-07161:
well, I think pulling an interpreter is the only way if we don't want to depend on system libs... and we cannot expect everyone to have lua libs installed just to compile halloy

4e554c4c:
summarizing the discussion on irc, lua can be embedded without pulling in system libs. however, python is the most important for hexchat compatibility

my preference is toward python not being included by default, or being behind a feature that can be disabled

user0-07161:
Basic plugins work now, I'll do the final cleanup tomorrow and then undraft this

user0-07161:
Basic plugins like [slap.py](https://github.com/hexchat/hexchat-addons/blob/master/python/slap/slap.py) work now, therefore marking as ready to review

casperstorm:
Thank you for this contribution. I'll discuss this with @tarkah and @andymandias so be prepared that this might sit for awhile.

casperstorm:
Again, thanks for the work here; however we have decided that Halloy won't strive to become Hexchat compatible.
We will however create something new, down the line, and offer support to help translate the relevant scripts.

PR #1033, opened 2025-06-18
Change from `dark-light` to `mundy`

4e554c4c:
This commit leaves behind the `dark-light` crate, as it does not support polling for dark/light mode changes. This is an essential feature of a dark/light detecting program, since all major OS support notification of appearance changes.

The mundy crate also supports contrast information and accent color information, which might be useful in the future.

Calling `detect()` is a relatively expensive operation, and now thousands of `detect()` calls a day are eliminated.

Tested on Linux and macOS

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -81,7 +81,7 @@ rfd = { version = "0.15.2", default-features = false, features = [
     "tokio",
 ] }
 uuid = { version = "1.15", features = ["v4"] }
-mundy = { version = "0.1.9", features = ["tokio", "log"] }
+mundy = { version = "0.1.9", default-features = false, features = ["color-scheme", "tokio", "log"] }
 
 [target.'cfg(windows)'.dependencies]
 image = "0.24.6"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,13 +75,13 @@ open = "5.0.1"
 bytesize = "2.0.1"
 rodio = "0.20.1"
 humantime = "2.1.0"
-dark-light = { git = "https://github.com/rust-dark-light/dark-light", rev = "8e1f745f91e1e805fa772a83e4744afe95d70aa1" }
 strsim = "0.11.1"
 rfd = { version = "0.15.2", default-features = false, features = [
     "xdg-portal",
     "tokio",
 ] }
 uuid = { version = "1.15", features = ["v4"] }
+mundy = { version = "0.1.9", features = ["tokio", "log"] }
 
 [target.'cfg(windows)'.dependencies]
 image = "0.24.6"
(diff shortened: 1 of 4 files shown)


casperstorm (approved):

Cargo.toml
```diff
@@ -75,13 +75,13 @@ open = "5.0.1"
...
 rodio = "0.20.1"
 humantime = "2.1.0"
-dark-light = { git = "https://github.com/rust-dark-light/dark-light", rev = "8e1f745f91e1e805fa772a83e4744afe95d70aa1" }
 strsim = "0.11.1"
 rfd = { version = "0.15.2", default-features = false, features = [
     "xdg-portal",
     "tokio",
 ] }
 uuid = { version = "1.15", features = ["v4"] }
+mundy = { version = "0.1.9", features = ["tokio", "log"] }
```
andymandias: I think we can get away with
```toml
mundy = { version = "0.1.9", default-features = false, features = ["color-scheme", "tokio", "log"] }
```
and shave off a dependency.  At least for Linux that still works for me, haven't tested on Windows/macOS.

src/appearance.rs
```diff
@@ -1,56 +1,43 @@
...
-                Mode::Light => light.clone(),
+impl Mode {
+    pub fn theme(&self, selected: &data::appearance::Selected) -> data::appearance::Theme {
+        match &selected {
+            appearance::Selected::Static(theme) => theme.clone(),
+            appearance::Selected::Dynamic { light, dark } => match self {
+                Self::Dark => dark.clone(),
+                Self::Light => light.clone(),
+                // We map `Unspecified` to `Light`.
+                // This is because Gnome never specifies `Light` and only sends `Unspecified`.
```
andymandias, line 37: Technically GNOME has a light mode, but it's unfortunately not exposed in the Settings application (`gsettings set org.gnome.desktop.interface color-scheme prefer-light` for those so inclined).  That doesn't change the logic here, which is sound, I just wanted to note there is an escape hatch if it ever becomes problematic to map `Unspecified`/`NoPreference` to light mode.

4e554c4c: Yep, just going based on my [experience](https://github.com/4e554c4c/darkman.nvim/issues/6) maintaining darkman.nvim that this is a default behaviour that people want

Cargo.toml
4e554c4c: fair, i was mostly thinking that we might want to use some of the other preferences eventually. but i think it'd be better to not have to think about them for now.

4e554c4c: tested on macOS, it works.

4e554c4c:
who is going to merge :smiling_imp:

Issue #550, opened 2024-09-11
IRCv3 replies and reactions


4e554c4c:
There are two IRC features driven by client-only tags that are related: [replies](https://ircv3.net/specs/client-tags/reply) and [reactions](https://ircv3.net/specs/client-tags/react). These features allow for an experience more similar to a modern chat application. It would be nice if halloy could add support for these features after msgid support is added.

reesericci:
Would it make sense to split this into two issues, threads/replies and reactions?

4e554c4c:
the main issue is that reactions _kind of_ depend on replies, since a reaction is a special kind of reply

but you don't need to support all of replies to support reactions and vice versa. confusingly, the reference direction in code is usually opposite for replies and reactions as well (which is why in goguma we have two different msgid lookup tables, one for replies and one for reacts)

PR #1036, opened 2025-06-20
Refactor message renderer

4e554c4c:
This refactor moves message rendering primarily into one place. Previously every kind of buffer had its own message renderer, and now `query` and `channel` buffers share a renderer.

This is in preparation for the react changes, which should preferably all take place in one location.

needed for #550

### DIFF
--- a/src/buffer/message_view.rs
+++ b/src/buffer/message_view.rs
@@ -14,15 +14,46 @@ use crate::widget::{
 use crate::{Theme, theme};
 
 #[derive(Clone, Copy)]
+pub enum TargetInfo<'a> {
+    Channel {
+        channel: &'a target::Channel,
+        our_user: Option<&'a User>,
+        users: &'a [User],
+    },
+    Query,
+}
+
+impl<'a> TargetInfo<'a> {
+    fn users(&self) -> &'a [User] {
+        match self {
+            TargetInfo::Channel { users, .. } => users,
+            TargetInfo::Query => &[],
+        }
+    }
+    fn our_user(&self) -> Option<&'a User> {
+        match self {
+            TargetInfo::Channel { our_user, .. } => *our_user,
+            TargetInfo::Query => None,
+        }
+    }
+    fn channel(&self) -> Option<&'a target::Channel> {
+        match self {
+            TargetInfo::Channel { channel, .. } => Some(channel),
+            TargetInfo::Query => None,
+        }
+    }
+    fn is_channel(&self) -> bool {
+        matches!(self, TargetInfo::Channel { .. })
+    }
+}
+
+#[derive(Clone, Copy)]
 pub struct ChannelQueryLayout<'a> {
     pub config: &'a Config,
-    pub users: &'a [User],
     pub casemapping: CaseMap,
     pub server: &'a Server,
-    // if we are laying out a query buffer, then this and our_user should be None
-    pub channel: Option<&'a target::Channel>,
-    pub our_user: Option<&'a User>,
     pub theme: &'a Theme,
+    pub target: TargetInfo<'a>,
 }
 
 impl<'a> ChannelQueryLayout<'a> {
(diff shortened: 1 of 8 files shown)


src/buffer/message_view.rs
```diff
@@ -0,0 +1,312 @@
...
+use iced::advanced::text;
+use iced::widget::{column, container, row};
+
+use super::user_context;
+use crate::widget::{
+    Element, message_content, message_marker, selectable_text,
+};
+use crate::{Theme, theme};
+
+pub trait MessageFormat<'a> {
```
casperstorm: Wonder if we can find a better name here 🤔

src/buffer/message_view.rs
```diff
@@ -0,0 +1,312 @@
...
+    pub config: &'a Config,
+    pub users: &'a [User],
+    pub casemapping: CaseMap,
+    pub server: &'a Server,
+    pub channel: Option<&'a target::Channel>,
+    pub our_user: Option<&'a User>,
+    pub theme: &'a Theme,
+}
+
+impl<'a> ChannelFormat<'a> {
```
casperstorm: Same as above.

casperstorm (approved):
Small nits, but looks good. Will await @tarkah for final review.

src/buffer/message_view.rs
4e554c4c: maybe `DrawMessage`

src/buffer/message_view.rs
```diff
@@ -0,0 +1,312 @@
+use crate::buffer::scroll_view::Message;
```
tarkah, line 1: We now have a circular dependency between `scroll_view` and `message_view` modules. Seems `scroll_view` should define the trait that consumers need to implement for formatting messages within the scroll view. And this module just provides one of those implementations.

4e554c4c:
> Along those lines, it might nice if the delineation between channel and query was a bit more explicit. That `users`, `channel`, and `our_user` should each be empty/None for queries is not necessarily the most intuitive. But a clearer alternative is not coming to me at this hour, and I don't think it's necessary.

This is fair. I could pretty easily do an enum with two variants, but i think this goes back to violating DRY. I think it's more clean this way, so maybe the answer is more comments in some places making this explicit? feel free to suggest any.

src/buffer/message_view.rs
```diff
@@ -0,0 +1,288 @@
...
+    }
+}
+impl<'a> LayoutMessage<'a> for ChannelQueryLayout<'a> {
+    fn format(
+        &self,
+        message: &'a data::Message,
+        max_nick_width: Option<f32>,
+        max_prefix_width: Option<f32>,
+    ) -> Option<Element<'a, Message>> {
+        //let fm = *self;
```
casperstorm: Nit: remove this

src/buffer/message_view.rs
```diff
@@ -0,0 +1,288 @@
...
+
+#[derive(Clone, Copy)]
+pub struct ChannelQueryLayout<'a> {
+    pub config: &'a Config,
+    pub users: &'a [User],
+    pub casemapping: CaseMap,
+    pub server: &'a Server,
+    // if we are laying out a query buffer, then this and our_user should be None
+    pub channel: Option<&'a target::Channel>,
+    pub our_user: Option<&'a User>,
```
casperstorm: I think it would be cleaner to introduce a `enum` here - something like

```rust
pub enum TargetInfo {
    Channel {
        channel: target::Channel,
        user: User,
    },
    Query,
}
```

src/buffer/message_view.rs
```diff
@@ -0,0 +1,288 @@
...
+
+use super::scroll_view::LayoutMessage;
+use super::user_context;
+use crate::widget::{
+    Element, message_content, message_marker, selectable_text,
+};
+use crate::{Theme, theme};
+
+#[derive(Clone, Copy)]
+pub struct ChannelQueryLayout<'a> {
```
casperstorm, line 51: Perhaps "`BufferLayout`"?

4e554c4c: This doesn't layout all buffers though. Log, server and highlight buffers are excluded

casperstorm: Good point yeah.

casperstorm (approved):

PR #1039, opened 2025-06-22
Reacts

4e554c4c:
Depends on #1036 

This patch is designed to lay the groundwork of irc reactions, so that we can build upon them later. It allows for viewing reactions, and sending reactions that have already been shown on a message.

Future work includes an emoji picker, and removing reactions.

There are two separate commits which can be reviewed separately.


data/src/history.rs
```diff
@@ -180,9 +180,11 @@ pub async fn overwrite(
     kind: &Kind,
     messages: &[Message],
     read_marker: Option<ReadMarker>,
+    reactions: &HashMap<MsgId, Vec<Reaction>>
 ) -> Result<(), Error> {
     if messages.is_empty() {
-        return metadata::save(kind, messages, read_marker).await;
+        // TODO(pounce) Remove clone
```
casperstorm: Nit: wanna look at this now?

data/src/message.rs
```diff
@@ -1139,12 +1325,21 @@ fn target(
     }
 }
 
-pub fn message_id(message: &Encoded) -> Option<String> {
+pub type MsgId = Arc<str>;
```
casperstorm: I am not entirely sure i like `MsgId` as name here. Maybe `Id` is better since we are in message crate after all.
`message::MsgId` vs `message::Id`

casperstorm:
I think UI looks good. Did you use a test server so i can try it out as well?
Code is clean, but would like @tarkah eyes here as well.

4e554c4c:
> Did you use a test server so i can try it out as well?

I used the ergo testnet: https://testnet.ergo.chat/

this PR has no support for adding the first reaction, so I did that in another app (namely, [goguma](https://codeberg.org/emersion/goguma)) and then tested with halloy.

data/src/message.rs
```diff
@@ -173,19 +191,168 @@ pub enum Direction {
...
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct Reaction {
+    pub target: Target,
+    pub sender: Nick,
+    pub text: String,
+    pub id: Option<Id>,
+    pub in_reply_to: Id,
+    pub unreact: bool,
+}
```
tarkah, line 202: We should have a different representation of a reaction for how its identified vs how its stored. When storing it, we only care about which message id it relates to. We don't need to store `unreact` reactions because we're simply removing them from history metadata. Same with `target`, this is only needed to figure out which history buffer to apply the reaction to, but once stored, its implied by where its stored. 

It might be good to have a  `reaction` module that allows us to describe both forms, one used at runtime for routing from decoding -> history. And one that we actually store in history that's slim.

data/src/message.rs
```diff
@@ -234,7 +401,7 @@ impl Message {
         }
     }
 
-    pub fn received<'a>(
+    pub fn _received<'a>(
```
tarkah, line 404: Don't forget to remove this

data/src/message.rs
```diff
@@ -173,19 +191,168 @@ pub enum Direction {
...
+                Self::Reaction(reaction.with_target(target))
+            }
+        }
+    }
+    pub fn target(&self) -> &Target {
+        match self {
+            Self::Message(message) => &message.target,
+            Self::Reaction(reaction) => &reaction.target,
+        }
+    }
```
tarkah, line 283: This method isn't needed. It's only called on a code path where we de-structure to `Message`, which it can just be referenced from that.

data/src/history.rs
```diff
@@ -201,15 +208,18 @@ pub async fn append(
...
     let mut all_messages = loaded.messages;
     messages.into_iter().for_each(|message| {
         insert_message(&mut all_messages, message);
     });
+    let mut all_reactions = loaded.metadata.reactions.into_owned();
+    all_reactions.extend(reactions);
 
-    overwrite(kind, &all_messages, read_marker).await
+    overwrite(kind, &all_messages, read_marker, &all_reactions).await
 }
```
tarkah, line 223: Isn't the full reaction history always held in memory? `append` is for appending new messages when only partial message history is open, but we should be able to just pass through `reactions` directly to `overwrite` since it should be everything.

Extending off the on-disk reactions is actually dangerous because it will re-inject things that were unreacted that we've removed from our in-memory state, but still exists on-disk. (Assuming we never store un-reacts which based on my other comment I think is the best way to go about it)

data/src/history/metadata.rs
```diff
@@ -9,13 +11,15 @@ use tokio::fs;
...
+use crate::message::{source, MessageReferences, Id, Reaction};
 
 #[derive(Debug, Clone, Default, Deserialize, Serialize)]
-pub struct Metadata {
+pub struct Metadata<'a> {
     pub read_marker: Option<ReadMarker>,
     pub last_triggers_unread: Option<DateTime<Utc>>,
     pub chathistory_references: Option<MessageReferences>,
+    #[serde(default)]
+    pub reactions: Cow<'a, HashMap<Id, Vec<Reaction>>>,
```
tarkah, line 22: Idea:

Instead of separately storing reactions in metadata & having to net out unreacts at runtime per message, can we just store reacts directly on  `Message`? This will speed up `view`, computation will only have to be done once in `update`.

When full history is open and we receive a reaction, we can simply append the reaction if its a new react, ignore it if its a react that already exists and remove the react if its an unreact that exists. This gets flushed as normal.

If its partial history, we can do this same logic within `append` after loading the full history from disk. Any new reactions we need to append is just applied in the same way before flushing.

This directly relates to my other comment about having a different representation for reacts for routing vs storage. The stored reacts can live on `Message { reacts: Vec<React> }` and decoded reacts are used for routing at runtime only.

4e554c4c: I didn't want to store reacts directly on message for a few reasons.

I want `Message` to be "object-like", i.e., easy to send around the codebase in an `Arc`. This is because in the future I also want to add a reply feature for #550 which means storing messages in another data structure (indexed by message ID). If we keep messages immutable, it's pretty easy to clone arcs and then keep two references to each message. but then we can't update them to store reacts easily.

We could also have `Message` be something like an `Arc<RefCell<Message>>` with a `Weak<Message>` pointer to the message it's replying to... but this is not the safest option. Also if the message it is pointed to leaves the history then we run into a bit of trouble.

I'm not the most confident in how the history system is designed, so I'm happy to take recommendations. But I think we should make sure each of these problems is solvable.

4e554c4c:
@tarkah curious about your thoughts re: message mutability.

data/src/history/metadata.rs
tarkah: We already mutate Message to store hidden URL previews.

What requirements are there for replies where we need to have the entire Message referencable vs just cloning out the needed data? If there's a legitimate performance consideration here, we can always store the immutable data of a Message behind an Arc while keeping the Message itself bare so it can have mutable data associated to it.

reesericci:
What are the blockers on getting this merged?

4e554c4c:
i got kind of burnt out and didn't want to rewrite the history code a third time mostly

the ui should be done though

i can probably get back to it next week

Hummer12007:
Please also add a knob for that in the config. I wouldn't want to see this in my irc client.

4e554c4c:
closing in favor of #1481, since most of the reviews here will no longer be relevant. please take all comments/concerns there

PR #1045, opened 2025-06-27
Pronunciation Guide

andymandias:
A brief pronunciation guide (and etymology) for Halloy, for those who might be curious.


book/src/guides/pronunciation.md
```diff
@@ -0,0 +1,3 @@
+# Pronunciation
+
+Halloy is an English transliteration of the Danish greeting halløj.  In other words, its pronunciation is close to the way most English speakers would guess based on its spelling.  Halloy's phonemic transcription in IPA is /halɔj/.
```
casperstorm: There's a extra whitespace in the last sentence.
Also, i would be OK with out linking directly to IPA here if you feel like it.

casperstorm (approved):

Issue #77, opened 2023-07-01
soju.im/bouncer-networks support


remexre:
The [Soju](https://soju.im/) bouncer has an IRCv3 capability, `soju.im/bouncer-networks`, to allow automatically creating connections to the networks that are configured in the bouncer.

It would be very convenient to have support for this in halloy.

An example implementation of client support for this as a Weechat script may be found [here](https://github.com/weechat/scripts/blob/master/python/soju.py).

remexre:
Yep, that works as a workaround, but then the list of networks needs to be edited on the client whenever a new network is added.

4e554c4c:
I'd like to take a look at doing this. My general idea of how bouncer-networks works is
1. Client connects to the bouncer and completes registration
2. Client requests list of servers and a notification stream for server changes
3. Client opens a new connection for each server and requests `BOUNCER BIND` during registration

So there are  a few complications, such as whether to show the initial server (used for bouncer communication) in the UI, and also how to structure a UI for editing server info (maybe this part should come later). But in general it seems rather doable.

PR #1048, opened 2025-06-27
Use `BTreeMap` for tags instead of a vector

4e554c4c:
This commit is a minor convenience, and removes a minor footgun (two possible values for an empty tag, which the ircv3 spec forbids discriminating).

`BTreeMap` should have equivalent (that is, pretty fast) performance for <= B (6) tags which is pretty common. For more tags it should most likely give better performance. It also guarentees that we don't care about tag order and duplication.

There is a minor refactor in `format.rs` that allows for paring tag-escaped key-value strings for squidowl/halloy#77.

### DIFF
--- a/irc/proto/src/parse.rs
+++ b/irc/proto/src/parse.rs
@@ -1,14 +1,16 @@
 use nom::branch::alt;
 use nom::bytes::complete::tag;
 use nom::character::complete::{alpha1, char, crlf, none_of, one_of, satisfy};
-use nom::combinator::{cut, map, opt, peek, recognize, value, verify};
+use nom::combinator::{
+    complete, cut, map, opt, peek, recognize, value, verify,
+};
 use nom::multi::{
     many_m_n, many0, many0_count, many1, many1_count, separated_list1,
 };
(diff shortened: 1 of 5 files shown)


irc/proto/src/parse.rs
```diff
@@ -66,15 +83,11 @@ fn tags(input: &str) -> IResult<&str, Vec<Tag>> {
...
-        |(key, value): (&str, _)| Tag {
-            key: key.to_string(),
-            value,
-        },
+        |(key, value): (&str, _)| (key.to_string(), value.unwrap_or_default()),
     );
     // <tag> [';' <tag>]*
-    let tags = separated_list1(char(';'), tag);
+    separated_list1(char(';'), tag)(input)
     // '@' <tags> <SPACE>
```
andymandias: This doesn't make much sense here since you moved the `preceded`/`terminated` to the relevant nom into `message`.  I.e. I think this comment should be on/before line 24 now.

4e554c4c: good catch. fixed.

PR #1051, opened 2025-06-30
Wait for Server-Ack/Timeout After Quit

andymandias:
When quitting the final `MARKREAD` messages (and `QUIT` message) are ignored by my soju instance with the error log "ignoring message on closed connection".  These changes are to not immediately send the `Update::Quit` and enter `State::Quit` after sending the `QUIT` message, but to wait for either the server to respond with an `ERROR` message (which [horsedocs says is a valid acknowledgement of a `QUIT` command](https://modern.ircdocs.horse/#quit-message)) or for a timeout (tentatively set at 1 second).

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -221,6 +221,13 @@ impl Client {
     }
 
     fn quit(&mut self, reason: Option<String>) {
+        self.who_polls.retain(|who_poll| {
+            matches!(
+                who_poll.status,
+                WhoStatus::Requested(_, _, _) | WhoStatus::Receiving(_, _)
+            )
+        });
+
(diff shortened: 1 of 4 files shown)


4e554c4c:
```
<emersion> iirc "ignoring message on closed connection" would happen when one side of the connection got closed but not the other
<emersion> we should probably not do that, and consume all messages until the end…
```
i think this might be a soju issue that needs to be fixed upstream. @emersion mentioned wanting to fix this when I saw it happen somewhere else.

andymandias:
That's all well and good, but it's been affecting us for a while now and the time frame for bouncer roll out of a fix remains indeterminate.  If a server ack is included in the fix, then added wait time is next to nil.

Issue #1050, opened 2025-06-29
Unreasonable amount of time spent in `Client::sync`


4e554c4c:
I noticed this a long time ago, but haven't really had the time to debug/look into it.

I often notice halloy being incredibly slow when receiving a bunch of messages.

I opened [`cargo flamegraph`](https://github.com/flamegraph-rs/flamegraph) again tonight, to check for obvious culprits, and [`Client::sync`](https://github.com/squidowl/halloy/blob/d49ea7a5fa5498cc6bfd7c3c84e4ae209a224489/data/src/client.rs#L2686) is taking upwards of 30% of CPU time (even late at night, when not that many messages are being received). This is due to the channel list for each client, and _the userlist for every single channel for every single client_ in halloy being sorted every time a message is received.

Ideally these should just be kept in-order (by inserting with a binary search!) and this entire sync can be avoided.

PR #1052, opened 2025-06-30
Remove `Client::sync`

4e554c4c:
`Client::sync` sorts the channel list, and every single userlist for every single channel on every message received. This is a huge performance sink for halloy, and my tests reported halloy spends ~30% of its CPU time in this function alone.

This function exists to keep several "views" into the client structures in sync (sorted correctly) with the client structures themselves. However, I think `chanmap` and the user list already have all the information that we would want, in the right order.

This commit includes much refactoring because the client no longer returns slice views to its channels and users. Instead, I've elected to return a channel iterator, and a `ChannelUsers` view for users. This `ChannelUsers` view satisfies a set API, which can actually make many view operations more efficient.

There is a bit of magic here to avoid needless `clone()`s, and I needed to switch from an ordinary `BTreeMap` to an `IndexMap` for `chanmap`, since the order is dependent on runtime information (it was either this, or keeping a global mutex of chantypes, which can lead to UB if modified during operation).

Before: 
Image

After:
image

Resolves squidowl/halloy#1050

### DIFF
--- a/data/src/user.rs
+++ b/data/src/user.rs
@@ -1,7 +1,9 @@
-use std::collections::HashSet;
+use std::cmp::Reverse;
+use std::collections::BTreeSet;
 use std::fmt;
 use std::hash::Hash;
 
+use indexmap::{Equivalent, IndexSet};
 use irc::proto;
 use itertools::sorted;
 use serde::{Deserialize, Serialize};
@@ -18,7 +20,7 @@ pub struct User {
     username: Option<String>,
     hostname: Option<String>,
     accountname: Option<String>,
-    access_levels: HashSet<AccessLevel>,
+    access_levels: BTreeSet<AccessLevel>,
     away: bool,
 }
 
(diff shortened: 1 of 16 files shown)


data/src/user.rs
```diff
@@ -18,39 +19,190 @@ pub struct User {
...
         self.highest_access_level()
-            .cmp(&other.highest_access_level())
+    }
+}
+
+impl<'a> Borrow<dyn UserRef + 'a> for User {
+    fn borrow(&self) -> &(dyn UserRef + 'a) {
+        self
+    }
+}
```
tarkah: Ok this is actually super nifty, never thought of using a trait object as a way to satisfy Borrow using multiple pieces of data that can't be directly referenced

data/src/user.rs
```diff
@@ -18,39 +19,190 @@ pub struct User {
...
     }
 }
-
 impl Eq for User {}
 
-impl Hash for User {
-    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
-        self.nickname.hash(state);
+#[derive(Debug, Default)]
+pub struct ChannelUsers(BTreeSet<User>);
```
tarkah: I'm worried this can introduce a sneaky error where we might accidentally add the same user twice because we have different access levels at different times.

We can use the same approach you have in `get_by_nick` of checking against each access level for the added / removed nickname to safeguard this. Considering its a binary search, its not too much extra cost. But this is starting to get pretty complex.

Alternatively, we scrap this entirely and still do something akin to "sync" as before, but we only call "sync" after certain commands come in that actually mutate the channel / user state.

data/src/user.rs
4e554c4c: yes, I learned this too

data/src/user.rs
4e554c4c: Yes, this has been on my mind too.

I think we should keep this design for two reasons:
1. I think that odd bugs could sneak into the other implementation too. The previous implementation of PartialEq and Ord disagree, which is not allowed and could cause issues.
2. I think the set API is useful for downstream files, and they can offer much more efficient operations with it.


I think a better idea might be to get rid of all of the magic, and keep the previous `Eq` and `Hash` implementations. Then, we could let `ChannelUsers` be an `IndexMap`, which will always match when indexed by a `User`, but it is ordered correctly.

Then care is taken to insert the user to the right location in the channel on `insert` (which is basically what you are suggesting but without a clone).

4e554c4c:
@tarkah can I get an OK on the second comment? Waiting on your review to go ahead.

data/src/user.rs
tarkah: That makes sense to me 👍

casperstorm:
@4e554c4c path forward here is rebase and follow suggestion from @tarkah and we are good to merge this.

4e554c4c:
Rebased and fixed. Thanks :smile_cat:

casperstorm:
Huge, good job on this one buddy.

PR #1053, opened 2025-07-01
Log Unknown Settings

andymandias:
Logs unknown settings in `config.toml`, for easier debugging.  Planning on follow-up PR(s) to make the log easier to use.


casperstorm (approved):
Nice one!

PR #1054, opened 2025-07-02
Logs Pane Tweaks

andymandias:
A bundle of three tweaks to the logs pane to help with its utilization:
- Colorize the level text in log messages based on theme.
- Wrap log messages similar to query/channel buffers, where the message wrap is inset the width of the timestamp + level.  In my opinion this makes it easier to scan the buffer by log level.
- Add a configuration setting (`buffer.internal_messages.log_level`) to filter the logs pane by log level (set to `info` by default).  This works differently than channel/query filtering which hides filtered messages; instead, messages filtered out of the logs pane are not stored in its buffer history.  This is done because debug messages are currently verbose enough that they can relatively quickly fill the max messages for the buffer, causing non-debug messages to disappear from the pane (since all message slots are used for invisible debug messages).  I went with this approach since (a) log messages are somewhat disposable in my view and (b) the setting does not filter out entries to `halloy.log`.

A couple minor notes:
- This changes the message data structure for log messages (to make the log level more readily accessible), so log history will be wiped out when a user runs Halloy with these changes.  I didn't make the effort to avoid this since (as noted above) I think log messages are somewhat disposable.
- I tweaked a few existing log levels to surface what I think an average user will find useful, but additional tuning is probably warranted.


src/appearance/theme/selectable_text.rs
```diff
@@ -153,6 +152,21 @@ pub fn status(theme: &Theme, status: message::source::Status) -> Style {
     }
 }
 
+pub fn log_level(theme: &Theme, log_level: log::Level) -> Style {
+    let color = match log_level {
+        log::Level::Error => theme.colors().text.error,
+        log::Level::Warn => theme.colors().general.unread_indicator,
+        log::Level::Info => theme.colors().buffer.server_messages.default,
+        log::Level::Debug => theme.colors().buffer.code,
+        log::Level::Trace => theme.colors().text.secondary,
```
casperstorm: We could define new colors for this if we want?

andymandias: Sounds good to me.  Maybe we could have the new colors be optional with these (or others, I'm not committed to this exact set) as fallbacks?  So there's less pressure to add new colors to existing themes.

casperstorm:
Sorry, it was a bit early so i thought i was on another branch here :D

src/appearance/theme/selectable_text.rs
casperstorm: Agree. I think you should add the new colors to `text.*`
So perhaps:
* `text.warning`
* `text.info`
* `text.debug`
* `text.trace`

I am not entirely sure about debug and trace, they will most likely not be used a lot else where, but i rather have them instead of using some color from a completetly different area.

Also adding these new colors remember to update the local ferra theme. Perhaps consider:

```
warning: #f5d76e
info: #9cd1bb
debug: #b4a7d6
trace: #dcdcdc
```

I am open for other, but just to get going.

PR #1055, opened 2025-07-02
Option to hide `nick` and `mode` change server messages

casperstorm:
Ability to hide mode and nick changes:

```toml
[buffer.server_messages.change_mode]
exclude = ["*"]
include = ["#foobar"]

[buffer.server_messages.change_nick]
exclude = ["*"]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Added:
 
 - Configuration option per toast type for showing content in toasts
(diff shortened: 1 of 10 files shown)


andymandias (approved):
Looks good, works well in my testing.  Only thing I see is to add the new settings to the list of `server_messages` in `book/src/configuration/buffer.md`.

PR #1056, opened 2025-07-03
Sidebar Regression Fix

andymandias:
Fixes a regression where servers connected via link would not show in the sidebar.

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -3301,6 +3301,10 @@ impl Map {
         })
     }
 
+    pub fn servers(&self) -> impl Iterator<Item = &Server> {
+        self.0.keys()
+    }
+
     pub fn iter(&self) -> std::collections::btree_map::Iter<Server, State> {
         self.0.iter()
     }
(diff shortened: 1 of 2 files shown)


4e554c4c:
i think a better fix is to route `Halloy.servers` through these views
That's what I'm working on in #730

4e554c4c:
For the sake of clarity, i think this PR is good It's best to have a fix because my changes will take awhile to come through.

andymandias:
> i think a better fix is to route `Halloy.servers` through these views That's what I'm working on in #730

I don't have enough context to know exactly what you mean, but the reason `config.servers` is used here is because the user can request the servers in the sidebar be ordered as they are in their configuration file.

4e554c4c:
There are three main lists of servers in the halloy codebase
1. `server::Map` in `Config` (which is static _per config load_)
2. `server::Map` in the `Halloy` state struct (which holds all _active servers_)
3. the `Client` struct in the `Halloy` struct

the first two are well ordered, but only the second holds the correct info

andymandias:
When the info we're concerned with being correct is the way they are ordered in the configuration file, then only the first is guaranteed to contain it.  The second's order will currently mirror the first's for typical use, but by being an implicit mirroring it may inadvertently fall out of sync with future changes (or in non-typical use; e.g. currently, when you leave and rejoin a server).  We can make the mirroring more explicit, but I'll leave that to you.

Issue #624, opened 2024-10-21
Try SASL by default


4e554c4c:
SASL PLAIN should be tried by default before USER/PASS authentication, so that authentication can be completed pre-registration. Halloy could detect whether SASL is allowed by doing capability negotation, or by opening an initial connection to test caps before the main connection (this approach is taken by goguma).

This would be nice so that instead of a config like
```
[servers.foo]
...
[servers.foo.sasl.plain]
username = "..."
password = "..."
```

one could just do
```
[servers.foo]
username = "..."
password = "..."
```

SASL is required for several things, including `soju.im/bouncer-networks` (#77). So if we want people to be able to easily use this feature, we should make SASL easier to use

ell1e:
I ran into this with [ergo](https://github.com/ergochat/ergo/issues/2377) recently. I didn't even know what SASL for IRC is, or that it's apparently the predominant standard now. I think users shouldn't know about this. Given this seems to be the predominant authentication mechanism nowadays, I think Halloy's `nick_password` should just try SASL plain over TLS with the nickname & the password by default.

PR #1057, opened 2025-07-05
 Add support for soju.im/bouncer-networks

### LINKED ISSUE #77: soju.im/bouncer-networks support

4e554c4c:
This Pr allows for specifying a network for a bouncer, which will be queried for all other bounced networks. There should be no change for servers and bouncers which do not implement the bouncer network specification.

Several bouncers are supported, and existing setups which use the `user/network@host` syntax will not be affected.

`soju.im/bouncer-networks` requires that SASL be enabled, so a precursor for this being widely rolled out is #624.

Supersedes #730 
Closes #77 

Test plan: connect to soju with config

```toml
[servers.soju]
nickname = "pounce"
server = "..."
[servers.soju.sasl.plain]
username = "pounce"
password = "..."
```

Result: bounced servers appear in halloy:
image


mhegreberg:
I tested this and it seems to work as expected. I also confirmed that you have to use the [servers.$name.sasl.plain] parameter

andymandias:
When opening the link irc://irc.undernet.org:6667/mIRC (and clicking Accept on the ensuing prompt), Halloy closes for me with:
```bash
thread 'main' has overflowed its stack
fatal runtime error: stack overflow, aborting
```
No stack overflow when trying to reproduce on `main`.

My configuration is currently just three servers, none of the them undernet and none a bouncer (was just trying to test the sidebar changes).  Might be a minute before I can look into it further, so I'm passing along the limited info I have in case you want to investigate.

4e554c4c:
for the sake of clarity, i'm requesting this PR not be reviewed until I realize why opening an `irc://` link causes a stack overflow :sweat_smile:

data/src/bouncer.rs
```diff
@@ -0,0 +1,146 @@
...
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("Unable to parse bouncer network: {0}")]
+    Parse(#[from] ParseError),
+    #[error("Bouncer network missing field: {0}")]
+    MissingField(&'static str),
+    #[error("Invalid network state: {0}")]
+    WeirdState(String),
```
casperstorm: `InvalidState` isntead?

4e554c4c:
source of segfault:
```patch
diff --git a/data/src/server.rs b/data/src/server.rs
index 17d6c108..77666104 100644
--- a/data/src/server.rs
+++ b/data/src/server.rs
@@ -304,7 +304,7 @@ impl Map {
                 // this one is a little complicated.
                 let Some(parent) = server.parent() else {
                     // it's a new server without a parent. yeet to the end
-                    self.insert(server, config);
+                    self.0.insert(server, config);
                     return;
                 };
```
the server insert code is really complicated so this bit didn't get hit very often. Thanks @andymandias !

book/src/guides/connect-with-soju.md
```diff
@@ -2,19 +2,43 @@
...
 
+```toml
+[servers.soju]
+nickname = "<your-nickname>"
+server = "<your-bouncer-url>"
+port = 6697
+[servers.soju.sasl.plain]
+username = "<your-username>"
+password = "<your-password>"
+chathistory = true
```
casperstorm: chathistory should be under `servers.<name>` and is true by default so id say we can remove it.

book/src/guides/connect-with-soju.md
```diff
@@ -2,19 +2,43 @@
...
+```
+
 > 💡  as of 2025.1 Halloy supports [`chathistory`](../configuration/buffer.md#bufferchathistory), so the machine name (like `@desktop`) is no longer needed when `chathistory` is enabled
```
casperstorm: it should link to: https://halloy.chat/configuration/servers.html#chathistory if we still want this notice.

book/src/guides/connect-with-soju.md
```diff
@@ -2,19 +2,43 @@
...
-nickname = "casperstorm"
-username = "<your-username>/irc.libera.chat"
-server = "irc.squidowl.org"
+nickname = "<your-nickname>"
+server = "<your-bouncer-url>"
 port = 6697
+[servers.soju.sasl.plain]
+username = "<your-username>/irc.libera.chat"
 password = "<your-password>"
 chathistory = true
```
casperstorm: Same as above.

src/screen/dashboard/sidebar.rs
```diff
@@ -683,10 +672,24 @@ fn upstream_buffer_button<'a>(
...
-                .shaping(text::Shaping::Advanced)
+            if let Some(network) = &server.network {
+                Element::from(row![
+                    text(network.name.to_string())
+                        .style(buffer_title_style)
+                        .font_maybe(buffer_title_font.clone())
+                        .shaping(text::Shaping::Advanced),
+                    Space::with_width(6),
+                    text(server.name.to_string())
+                        .style(theme::text::tertiary)
```
casperstorm: lets do `theme::text::secondary` here as i treat `tertiary` more as a "hey look at this" where secondary is is used more as a "info".

casperstorm:
Screenshot 2025-07-29 at 13 27 36

Two nits on UX i was considering:
1. I think we should consider minimizing the gab (pink line) when there is no channels. It's a bit big currently.
2. Should we consider a new icon for bounced servers, or the "main" one ("soju") in my case? we use https://fontello.com/ so its easy to go through and consider any new ones.

4e554c4c:
> Two nits on UX i was considering:
> 
> 1. I think we should consider minimizing the gab (pink line) when there is no channels. It's a bit big currently.

I'd rather not in this PR, this sounds like an unrelated issue


> 2. Should we consider a new icon for bounced servers, or the "main" one ("soju") in my case? we use https://fontello.com/ so its easy to go through and consider any new ones.

That's a good idea!

src/main.rs
```diff
@@ -184,12 +182,14 @@ impl Halloy {
             }
         };
 
-        let (screen, config, command) = match config_load {
+        let (screen, servers, config, command) = match config_load {
             Ok(config) => {
-                let (screen, command) = load_dashboard(&config);
-
+                let servers = config.servers.clone().into();
```
andymandias: Currently `order_by = "alpha"` doesn't work for me (for non-bouncer-networks).  I tried changing this to
```rust
                let servers = server::Map::new(
                    config.servers.clone(),
                    config.sidebar.order_by,
                );
```
which does get alpha-sort working, but I haven't fully traced the logic to see if it breaks functionality elsewhere.  But, even with that change, changing `order_by` isn't reflected when I reload the config file (have to restart Halloy).  I'd like to preserve that functionality if we can.

4e554c4c: > changing order_by isn't reflected when I reload the config file

This really isn't possible with how i structured the code without a larger refactor. It was a non-goal for this PR

this is becoming a difficult task to repetitively rebase so I'd very much prefer that we merge and add this as a feature later

casperstorm: We should look into this. A regression is not something we want to introduce. We are not in a hurry so we can take the time we need before we merge.

data/src/bouncer.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub name: String,
+    pub host: Option<String>,
+    pub state: NetworkState,
+    pub port: Option<u16>,
+    pub use_tls: Option<bool>,
+    pub pass: Option<String>,
+    pub nickname: Option<String>,
+    pub realname: Option<String>,
+    pub error: Option<String>,
+}
```
tarkah: This struct probably needs to be broken up into a few things. 

Most of these are related to connection details that need to be submitted when requesting the bouncer adds a new network, not attributes it'll give us when listing networks we want to connect to. This stuff isn't even used yet so we can always omit it for now until we support some GUI to add a network. Probably out of scope since you can just do it via `BOUNCER` commands.

`state` isn't useful for us. We manage our own connection lifecycle and that is our source of truth. If we get a LISTNETWORK for something we're already connected to, we can verify that when handling `Event::BouncerNetwork` that `clients` doesn't already have an active connection for that server / network before adding it to our `server` list. 

In the end, I think `BouncerNetwork` can just be `id` and `name`, which make sense to both immutable store under `Server` and serialize along w/ `Server` in all its forms. All other info doesn't make sense to live under `Server`.

data/src/client.rs
```diff
@@ -835,6 +861,43 @@ impl Client {
...
+                        ..self.server.clone()
+                    },
+                    self.config.bouncer_config(),
+                )]);
+            }
+            Command::CAP(_, sub, Some(_asterisk), Some(caps))
+                if sub == "LS" =>
+            {
+                self.listed_caps.extend(caps.split(' ').map(String::from));
+            }
```
tarkah: This is already handled in the match below this one

data/src/client.rs
```diff
@@ -2270,108 +2363,127 @@ impl Client {
...
-                        group_joins(&channels, &self.config.channel_keys)
-                    {
-                        self.handle.try_send(message)?;
-                    }
+                // Request bouncer networks
+                // TODO(pounce) replace this with "bouncer-networks-notify" after the cap handling
+                // is cleaned up.
+                if self.is_primary() && self.supports_bouncer_networks {
+                    self.handle
+                        .try_send(command!("BOUNCER", "LISTNETWORKs"))?;
```
tarkah:
```suggestion
                        .try_send(command!("BOUNCER", "LISTNETWORKS"))?;
```

data/src/bouncer.rs
```diff
@@ -0,0 +1,146 @@
...
+        if !matches!(case_insensitive_ordering, Ordering::Equal) {
+            return case_insensitive_ordering;
+        }
+
+        let case_sensitive_ordering = self.name.cmp(&other.name);
+        if !matches!(case_sensitive_ordering, Ordering::Equal) {
+            return case_sensitive_ordering;
+        }
+
+        self.id.cmp(&other.id)
```
tarkah: [then_with](https://doc.rust-lang.org/std/cmp/enum.Ordering.html#method.then_with) makes composing these more functional and clean

data/src/server.rs
```diff
@@ -9,30 +9,73 @@ use serde::{Deserialize, Serialize};
...
 }
 
 impl fmt::Display for Server {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        self.0.fmt(f)
+        if let Some(network) = &self.network {
+            network.name.fmt(f)
+        } else {
+            self.name.fmt(f)
+        }
```
tarkah, line 62: `Display` drives some hashing for file access & hashmap key lookups, we probably don't want an accidental collision between a similarly named config network & bouncer network causing collisions. We should concat bouncer name w/ network name to ensure its unique, similar to how you do it for the binary format

data/src/server.rs
```diff
@@ -72,7 +116,7 @@ impl<'a> From<(&'a Server, &'a Arc<config::Server>)> for Entry {
 }
 
 #[derive(Debug, Clone, Default, Deserialize)]
-pub struct Map(IndexMap<Server, Arc<config::Server>>);
+pub struct ConfigMap(IndexMap<ServerName, Arc<config::Server>>);
```
tarkah, line 118: Not sure we need both `Map` and `ConfigMap`. A `Server` can be created from a bare `ServerName`, which is valid to use on `Config.servers`

4e554c4c: Screenshot_20250731-000001
Screenshot_20250731-000047

this is why i designed it this way

data/src/server.rs
4e554c4c: I would rather sniff these out and fix them, I tried to do that in this PR (and I've been daily driving it for a few weeks so i think I've taken care of most of them)

data/src/bouncer.rs
4e554c4c: I wanted to add this so it'd be easy to make a configuration UI eventually

but if we do, that should probably go in the server config somehow anyway.
I can change this to only be id and name (although host would be useful to open links!)

4e554c4c:
ok! server sorting is much more simple now. Each server gets assigned its config order on de-serialization, and that is used to sort servers in config order mode. This also means that we can change between config order and alphabetical order at runtime! yay!

there are also some unresolved conversations with @tarkah where I'm waiting for a response.

book/src/guides/connect-with-soju.md
```diff
@@ -2,19 +2,41 @@
...
+```
+
+> 💡  as of 2025.1 Halloy supports [`chathistory`](../configuration/servers.md#chathistory), so the machine name (like `@desktop`) is no longer needed when `chathistory` is enabled
+
+
+## Using bouncer networks
+
+> ⚠️  The soju bouncer networks specification *requires* that SASL be used. If you do not use SASL, you must add servers in the legacy fashion.
+
+As of 2025.8 Halloy supports [`bouncer networks`](https://codeberg.org/emersion/soju/src/branch/master/doc/ext/bouncer-networks.md) so connecting to individual servers is no longer needed. Instead, Halloy will communicate with soju to determine what networks you are currently connected to and then automatically add them in the UI.
```
trfunk: should be 2025.9, 2025.8 is already released

4e554c4c: it was right at the time...

andymandias (approved):
Overall looks really good!  Hope you don't mind, I made a couple minor tweaks to get the sidebar to respect `sidebar.order_by = "alpha"` on initial load and for `sidebar.order_by = "config"` to use the current config ordering on reload.  Didn't want to delay anything further.

casperstorm:
Thanks for all the effort on this one @4e554c4c

4e554c4c:
@tarkah no worries re: code review. We all have times where we can contribute more to open source projects!
I've appreciated your reviews thus far, and I hope a hiatus treats you well

Issue #856, opened 2025-03-30
Auto Accept DCC file transfer.


RealKindOne:
Here are some ideas for auto accepting files based on other IRC client features:

```
[file_transfer]

# Enable Auto accept:
auto_accept_files = true/false

# Only accept from these nicks:
# mIRC/WeeChat
auto_accept_nicks = ["nick1", "nick2"]

# Only accept from these masks:
# irssi
auto_accept_masks = ["*@foobar.com", "nick!ident@example.com"] 

# Only auto accept these extensions, this overrides the ignore_extensions:
# mIRC
auto_accept_only_allow_extensions = ["*.mkv", *.mp3"]

# Only ignore these extensions, this overrides the allow_extensions:
# mIRC
auto_accept_only_ignore_extensions = ["*.exe", "*.sh"]

# Override both allow_extensions and ignore_extensions:
# mIRC
dangerously_accept_all_files_extensions = true/false

# Replaces spaces in filename with underscore:
# mIRC/WeeChat/irssi
auto_accept_replace_space_with_underscore = true/false

# Set file permissions on creation:
# irssi
auto_accept_file_creation_mode = 644

# Limit max filesize in MB for Megabyte, GB for Gibabyte:
# irssi (uses 0k for no limit. so maybe use that here?)
auto_accept_max_filesize = 1MB

# Check for storage space.
# If not enough storage, don't download and send notification. 
# I don't know if any clients support this.
auto_accept_check_for_storage_space = true/false
```

casperstorm:
Plus points for the detailed feature request. I'll look into this, thanks.

PR #1058, opened 2025-07-05
Auto accept file transfers

casperstorm:
This fixes (some of) https://github.com/squidowl/halloy/issues/856.

```toml
[file_transfer]
save_directory = "/Users/crs/Downloads"

[file_transfer.auto_accept]
enabled = true
nicks = ["boo", "slasker"]
masks = [".*!.*@example.com"]
```
You are now able to auto accept files by using `auto_accept_files` - you also have  `auto_accept_nicks` and `auto_accept_masks` if you wish to only accept certain nicks/masks.

Above would  auto accept a file if:
Their nickname is "boo" or "slasker" and their mask matches ".*@example.com"

### DIFF
--- a/book/src/configuration/file_transfer.md
+++ b/book/src/configuration/file_transfer.md
@@ -15,19 +15,6 @@ Default directory to save files in. If not set, user will see a file dialog.
 save_directory = "/Users/halloy/Downloads"
 ```
 
-## `auto_accept_files`
-
-If true, automatically accept incoming file transfers. Requires `save_directory` to be set.
-
-```toml
-# Type: boolean
-# Values: true, false
-# Default: false
-
-[file_transfer]
-auto_accept_files = false
-```
-
 ## `passive`
 
 If true, act as the "client" for the transfer. Requires the remote user act as the [server](#file_transferserver).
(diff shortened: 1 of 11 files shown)


data/src/file_transfer/manager.rs
```diff
@@ -198,13 +198,57 @@ impl Manager {
...
+                        |nicks| nicks.contains(&from.nickname().to_string()),
+                    );
+
+                let mask_match =
+                    config.file_transfer.auto_accept.masks.as_ref().is_none_or(
+                        |masks| {
+                            let a = from.matches_masks(masks);
+                            println!("masks: {masks:?}");
+                            println!("matches user with mask {a}");
+                            a
```
andymandias: Suspect we want to promote these to log messages or remove them.

There's another `println` in `User.matches_masks` as well.  (And one in `ipc/src/server.rs` that isn't a part of this PR, but might be worth promoting to a log message while we're at it.)

book/src/configuration/file_transfer.md
```diff
@@ -97,3 +97,48 @@ Last port in port range to bind to.
...
+Note `auto_accept` has to be enabled.
+
+```toml
+# Type: array of strings
+# Values: array of any strings
+# Default: []
+
+[file_transfer.auto_accept]
+masks = ["nick!ident@example.com", ".*@foobar.com"]
+```
```
andymandias: For the description, how about we add a sentence to describe the full nickname format:
```md
If true, automatically accept incoming file transfers from these nicks.  Matches are made against the full nickname (i.e. nickname, username, and hostname in the format `nickname!username@hostname`).
Note `auto_accept` has to be enabled.
```

Tangentially related, there are a couple minor typoes in the examples for `bind_port_first` and `bind_port_last` above;  they're written as strings when they should be integers.

Not strictly necessary, but might be good to have the masks given as multi-line string literals as we suggest for [regex](https://halloy.chat/configuration/highlights.html#regex).

andymandias (approved):
Looks good!  Simple tests of all the new settings were all successful.

data/src/file_transfer/manager.rs
casperstorm: Lol, thats a miss from my side. Thanks for spotting them.

PR #1060, opened 2025-07-06
Building for flatpak guide

englut:
Sharing some knowledge on how to build & install flatpaks for latest commit and how to release to Flathub!

### DIFF
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -17,6 +17,7 @@
 - [Monitor users](guides/monitor-users.md)
 - [YAML migration](guides/migrating-from-yaml.md)
 - [Pronunciation](guides/pronunciation.md)
+- [Building for flatpak](guides/flatpaks.md)
 
 # Configuration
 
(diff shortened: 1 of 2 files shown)


casperstorm:
Hmm, shoudn't we add this as a guide instead in the book @englut?

englut:
> Hmm, shoudn't we add this as a guide instead in the book @englut?

@casperstorm I just stuck this as a readme next to the manifest in the `assets/flatpak` folder. Does this get built out into the book? :eyes:

casperstorm:
Nah we need it under the guide folder inside book :)

englut:
> Nah we need it under the guide folder inside book :)

Ah.. darn, my bad. I misread your first comment. :pray:  Gotcha! I've moved the blurb to a guide in the book.

@casperstorm I'm seeing a build error in the "Build & Link Check", which is caused by `efnet.org` not loading. Shall I remove the reference to it? Or better to wait for the server to come back?

andymandias:
> @casperstorm I'm seeing a build error in the "Build & Link Check", which is caused by `efnet.org` not loading. Shall I remove the reference to it? Or better to wait for the server to come back?

I suspect this is temporary, I'm waiting on it for anther PR so I can rerun the check on this PR if/when it's back up.

englut:
> I suspect this is temporary, I'm waiting on it for anther PR so I can rerun the check on this PR if/when it's back up.

Awesome, thanks @andymandias :pray:

PR #1067, opened 2025-07-10
Linux Installation Script

andymandias:
A stripped down version of my personal install-from-source script for Halloy.  There was a request for something like this a while back in `#halloy`.


casperstorm (approved):
Should we have a mention to some of these scripts in a guide file? Doesnt have to be from this MR

PR #1068, opened 2025-07-11
Split `User::try_from` by Use

andymandias:
In `#libera` I saw a user renamed to a nickname that started with numbers (presumably by staff, since I think nicknames are otherwise not allowed to start with numbers on Libera).  In messages the renamed user sent in channel, the numbers were stripped from the beginning of their nickname.  In other places (the nick changed message and in the channel users list, if I recall correctly) the numbers were not stripped.  I think this might be why, but I'm having trouble reproducing the issue.

Even if it's not the fix, I think this is a benign allowance.  I believe we allow nicknames that start with numbers in the `user` function in `irc/proto/src/parse.rs`, and [horsedocs](https://modern.ircdocs.horse/#clients) does not restrict nicknames from starting with a number.

Edit:  Found a way to repro on Ergo's testnet what I have to assume is the same issue.  Initially receiving messages via chathistory on Ergo, usernames starting with numbers do appear with their numbers-prefix.  But, if I close the buffer and reopen it, then the numbers are stripped from the beginning of usernames in already-received messages.  Can confirm this fixes.

### DIFF
--- a/data/src/isupport.rs
+++ b/data/src/isupport.rs
@@ -1194,7 +1194,9 @@ pub fn find_target_limit(
     }
 }
 
-pub fn get_casemapping(isupport: &HashMap<Kind, Parameter>) -> CaseMap {
+pub fn get_casemapping_or_default(
+    isupport: &HashMap<Kind, Parameter>,
+) -> CaseMap {
     if let Some(Parameter::CASEMAPPING(casemapping)) =
         isupport.get(&Kind::CASEMAPPING)
     {
@@ -1205,7 +1207,9 @@ pub fn get_casemapping(isupport: &HashMap<Kind, Parameter>) -> CaseMap {
 }
 
 // https://modern.ircdocs.horse/#chanmodes-parameter
-pub fn get_chanmodes(isupport: &HashMap<Kind, Parameter>) -> &[ModeKind] {
+pub fn get_chanmodes_or_default(
+    isupport: &HashMap<Kind, Parameter>,
+) -> &[ModeKind] {
     isupport
         .get(&Kind::CHANMODES)
         .and_then(|chanmodes| {
@@ -1220,7 +1224,9 @@ pub fn get_chanmodes(isupport: &HashMap<Kind, Parameter>) -> &[ModeKind] {
         .unwrap_or(DEFAULT_CHANMODES)
 }
 
-pub fn get_chantypes(isupport: &HashMap<Kind, Parameter>) -> &[char] {
+pub fn get_chantypes_or_default(
+    isupport: &HashMap<Kind, Parameter>,
+) -> &[char] {
     isupport
         .get(&Kind::CHANTYPES)
         .and_then(|chantypes| {
@@ -1236,7 +1242,9 @@ pub fn get_chantypes(isupport: &HashMap<Kind, Parameter>) -> &[char] {
 }
 
 // https://modern.ircdocs.horse/#modes-parameter
-pub fn get_mode_limit(isupport: &HashMap<Kind, Parameter>) -> Option<u16> {
+pub fn get_mode_limit_or_default(
+    isupport: &HashMap<Kind, Parameter>,
+) -> Option<u16> {
     isupport
         .get(&Kind::MODES)
         .and_then(|modes| {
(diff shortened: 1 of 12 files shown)


glguy:
The client shouldn't be validating nicknames sent by the server. If the server says something it is a nickname, it is.

On solanum when there's a collision between for the same nickname on two servers, one of the two will be renamed to their UID (which will always start with the server's 3 character unique identity which starts with a digit so that it can't collide with any nickname)

RealKindOne:
You can ```/nick 0``` on networks that run InspIRCd. 

```[00:55:29] * KindOne is now known as 2KNACNP1I```

**Comment by RealKindOne** [comment]
> On solanum when there's a collision between for the same nickname on two servers, one of the two will be renamed to their UID

Both get renamed.

**Comment by casperstorm** [comment]
Good information here, TIL.

**Comment by andymandias** [comment]
I searched through the code base to see exactly how `User::try_from` was being used, and found there were a variety with differing needs:
- The simplest were places where it was just a convenience function for making a `User` with a nickname given by a string literal (e.g. `User::try_from("foobar")?`).  In those cases, I simply replaced the call with the equivalent `User::from(Nick::from(&str))`.
- Similarly, there were server messages where I believe we can reasonably expect that we're receiving a bare nickname (`KICK`, `RPL_WHOISIDLE`, `RPL_WHOISSERVER`, `RPL_WHOISUSER`, `RPL_WHOISCHANNELS`, `RPL_WHOISACTUALLY`, `RPL_WHOISSECURE`, `RPL_WHOISACCOUNT`, `RPL_AWAY`, and `RPL_MONONLINE`).  I converted these to `User::from(Nick::from(&str))` as well.
- When storing who set a topic we were using a `String` and then converting to a `User` via `User::try_from`.  When setting the `TOPIC` we already have a `User` whose nickname we can access, and when receiving a `RPL_TOPICWHOTIME` we can expect ~~a bare nickname~~.  So, I changed it to store a `Nick` directly (which we can `User::from` directly).  Edit: `RPL_TOPICWHOTIME` warrants passing to `User::parse` since testing shows it gets sent with userhost information.

Then, there are three intertwined use cases for `User::try_from(&str)` `String::from(User)` that have slightly different requirements, which I decided to split into distinct functions:
- In `User::matches_mask` I think we don't want to have the match against the prefixes of the user (unlike the other two uses).  So, I made a new `UsernameFormat` variant that makes `User::formatted` create a `nickname!username@hostname` from a `User`.
- When receiving `RPL_NAMREPLY` we need to be able to parse `<prefix(es)><nickname><!username@hostname>` (where `<!username@hostname>` won't be present without `userhost-in-names`).  For this purpose I modified the existing `User::try_from` into `User::parse`;  instead of looking for the first character of the nickname within a set of characters, it uses the PREFIX ISUPPORT parameter to match any & all prefix characters, then anything after that is the nickname (and possibly the userhost).
- When serializing/deserializing a `User`.  We want to read/write something like what's in `RPL_NAMREPLY`, but we don't necessarily know what the PREFIX ISUPPORT parameter is in these cases.  For now I've set it to match any possible prefix when deserializing, but I think it might be preferable to move the serialization scheme to its own format which doesn't need to know the PREFIX ISUPPORT at all.

File: data/src/isupport.rs [hunk 1]
```diff
@@ -1236,7 +1242,9 @@ pub fn get_chantypes(isupport: &HashMap<Kind, Parameter>) -> &[char] {
 }
 
 // https://modern.ircdocs.horse/#modes-parameter
-pub fn get_mode_limit(isupport: &HashMap<Kind, Parameter>) -> Option<u16> {
+pub fn get_mode_limit_or_default(
+    isupport: &HashMap<Kind, Parameter>,
+) -> Option<u16> {
```
casperstorm, line 1248: Nit: i expect `*_or_default` fns to not return `Option<>` 🤔

data/src/user.rs
```diff
@@ -325,6 +267,16 @@ impl User {
...
+            UsernameFormat::Mask => format!(
+                "{}{}{}",
+                self.nickname(),
+                self.username()
+                    .map(|username| format!("!{username}"))
+                    .unwrap_or_default(),
+                self.hostname()
+                    .map(|hostname| format!("@{hostname}"))
+                    .unwrap_or_default()
+            ),
```
casperstorm, line 279: Perfect.

src/buffer/channel/topic.rs
```diff
@@ -45,77 +45,66 @@ pub fn view<'a>(
     prefix: &'a [isupport::PrefixMap],
     channel: &'a target::Channel,
     content: &'a message::Content,
-    who: Option<&'a str>,
+    who: Option<Nick>,
```
casperstorm: Could we use NickRef here? 🤔

data/src/isupport.rs
andymandias: The value itself is an `Option<u16>` (where `None` means unlimited), but this format is definitely confusing.  I'll add a comment to clarify.

casperstorm: Perfect a comment will do.

src/buffer/channel/topic.rs
andymandias: We certainly can :)

casperstorm (approved):
🚀

PR #1069, opened 2025-07-12
Refactor conditional logic to use && and let chains

casperstorm:
I made clippy nightly happy.

### DIFF
--- a/data/src/user.rs
+++ b/data/src/user.rs
@@ -238,7 +238,7 @@ impl User {
         self.username.as_deref()
     }
 
-    pub fn nickname(&self) -> NickRef {
+    pub fn nickname(&self) -> NickRef<'_> {
         NickRef(&self.nickname.0)
     }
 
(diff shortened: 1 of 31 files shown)


data/src/user.rs
```diff
@@ -440,11 +440,7 @@ impl AsRef<str> for NickRef<'_> {
 impl PartialOrd for NickRef<'_> {
     fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
         // TODO(pounce) casemapping
-        Some(
```
casperstorm, line 443: CC @4e554c4c, let me know if you agree with this.

4e554c4c: i disagree, it's a good approximation until we support additional casemappings
all casemappings say `A..Z` go to `a..z`, the only difference is {', '}', '|', and '^' and unicode characters

4e554c4c: Technically our current approach matches the "ascii" casemapping, which is the most conservative casemapping.

casperstorm: The section throws the following clippy error if you are curious: https://rust-lang.github.io/rust-clippy/master/index.html#non_canonical_partial_ord_impl

4e554c4c: ah I see, maybe we should change it then.

andymandias (approved):
Looks good to me!  :partying_face: `let` chains!

PR #1061, opened 2025-07-07
Document build deps, Tor proxy

KE0VVT:
I had to do some fiddling to get Halloy to build, and more fiddling to get it to work with Tor. I've documented how to do both. Hopefully this helps.

PR #1073, opened 2025-07-13
Tor support is now an optional feature

casperstorm:
After a few discussions on IRC we have agreed that the best approach forward is to move `tor` support to a optional feature. 
This way we also reduce the binary size by almost 20%.

### DIFF
--- a/irc/Cargo.toml
+++ b/irc/Cargo.toml
@@ -5,16 +5,31 @@ authors.workspace = true
 license.workspace = true
 edition.workspace = true
 
+[features]
+default = []
+tor = ["arti-client"]
+
 [dependencies]
 thiserror = { workspace = true }
 futures = { workspace = true }
-tokio = { workspace = true,  features = ["full"] }
+tokio = { workspace = true, features = ["full"] }
 bytes = { workspace = true }
 
-arti-client = { version = "0.26", default-features = false, features = ["rustls", "compression", "tokio", "static-sqlite"] }
-async-http-proxy = { version = "1.2.5", features = ["runtime-tokio", "basic-auth"] }
+arti-client = { version = "0.26", default-features = false, features = [
+    "rustls",
+    "compression",
+    "tokio",
+    "static-sqlite",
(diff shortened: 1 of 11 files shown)


irc/Cargo.toml
```diff
@@ -5,16 +5,31 @@ authors.workspace = true
...
 bytes = { workspace = true }
 
-arti-client = { version = "0.26", default-features = false, features = ["rustls", "compression", "tokio", "static-sqlite"] }
-async-http-proxy = { version = "1.2.5", features = ["runtime-tokio", "basic-auth"] }
+arti-client = { version = "0.26", default-features = false, features = [
+    "rustls",
+    "compression",
+    "tokio",
+    "static-sqlite",
+], optional = true }
```
andymandias: In light of #1061 should we add "onion-service-client" to the features list, since not having it can sometimes produce a run-time error?  Or, since [arti-client](https://docs.rs/arti-client/latest/arti_client/) says it's not as secure as C-Tor, add a feature for enabling it?  I don't really have any sense of what "not as secure" means in this context.

casperstorm:
@KE0VVT Could you take at this, and try run this and see if its all good? I am wondering if we need `default-features = true` as well 🤔 

```sh
cargo run --features tor
```

KE0VVT:
I did `git pull` but still couldn't get it to work.

```
***@***.***:~/.local/src/halloy$ cargo run --features tor
error: the package 'halloy' does not contain this feature: tor
help: there is a similarly named feature: dev
```

-- 
Caleb Herbert
https://calebh.top

casperstorm:
@KE0VVT You need to ensure you are on this branch:

```sh
git fetch --all
git checkout optional-feature-tor
```

KE0VVT:
@casperstorm It builds on Fedora 42. (It took a long time; I'm on a Chromebook.) Connects to Tor just fine.

andymandias (approved):
Looks good to me.

PR #1074, opened 2025-07-13
Dedicated scrollbar color to theme configuration

### LINKED ISSUE #947: [Docs] Custom themes : Indicate that horizontal_rule color also affects scrollbar
As found at:
https://halloy.chat/configuration/themes/index.html

And with a sample here:
https://github.com/squidowl/halloy/blob/main/assets/themes/ferra.toml

Mention, using a comment (?) that horizontal_rule also affects scrollbar:
horizontal_rule = "<string>" # Also affects scrollbar

There might be others that affect multiple items as well.

Celeo:
Fixes #947

I feel kinda bad about submitting a PR with such a small change, but the issue is still up, so hopefully it's still useful.

I'm not a theming wizard so I don't know any other rules that might also need a note. If anyone does, I'd certainly be happy to add more comments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,14 @@ Added:
 - Context menu item to server buffers to mark all messages on the server as read
 - Auto-accept file transfer option with support for nick and mask filtering
 
+Changed:
+
+- Scrollbars now use dedicated `scrollbar` color instead of `horizontal_rule` color
+
 Thanks:
 
 - Bug reports: @darienm, @mercster
-- Feature requests: @rossburton, death916
+- Feature requests: @rossburton, death916, @Celeo
 
 # 2025.6 (2025-06-14)
 
--- a/assets/themes/ferra.toml
+++ b/assets/themes/ferra.toml
@@ -1,6 +1,7 @@
 [general]
 background = "#2b292d"
 horizontal_rule = "#323034"
+scrollbar = "#323034"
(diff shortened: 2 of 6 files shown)


casperstorm:
I feel like we should simply extend the theme with a scroll bar option, and then make it fallback on the old one.

casperstorm:
@Celeo, I create a new `scrollbar` entry in the theme which the scrollbar in Halloy now uses rather than `horizontal_rule`.
Let me know what you think.

Note that with this, themes which doesn't add `scrollbar` will see it as trasnparent, we won't fallback to `horizontal_rule`.

Celeo:
Makes sense to me; hopefully that'll allow themers to have more control.

I wonder if all themes on <https://themes.halloy.chat/> would have to be updated?

casperstorm:
> I wonder if all themes on https://themes.halloy.chat/ would have to be updated?

They will, but i'll look into backfilling them and add `scrollbar` to them with same value as `horizontal_rule`

casperstorm:
> I don't think it needs to be addressed in this PR necessarily, but there were also some requests in #halloy a while back for customization of the scrollbar thickness. In particular, a way to make scrollbars thicker (for easier grabbing).

https://github.com/squidowl/halloy/pull/1075

andymandias (approved):
Looks good to me, tests well.

casperstorm (approved):

PR #1075, opened 2025-07-14
Add configurable scrollbar sizing

casperstorm:
@andymandias mentioned it had been requested.


andymandias (approved):
Looks good and works well for me.  Should we add similar settings for the sidebar?

PR #1077, opened 2025-07-15
initial scroll view page size of height / line_height

tarkah:
Instead of always loading 500+ messages, we now initially only load however many messages fit on the y axis and then lazily grow the scrollable from there.

Also fixes a bug where going the start of message history and then scrolling down, it was loading all messages vs anchoring to top + step count (lazy loading).

This should massively improve performance by reducing number of messages being loaded in view & layout & text caching by `open_buffers * (500 - (buffer_height  / line_height))` so say you have 2 side-by-side buffers open with a window height of ~768px and default line height of ~`(1.3 * 13)=16.9` we'd reduce the number of messages impacting view from `500 * 2 = 1000` to `2 * (500 - (768 / 16.9)) = 909` 909 messages! Each buffer will only have to load 45 messages in view vs 500 each. This should massively reduce CPU time spent on text caching / layout. 

Now users really only incur increasing cost when scrolling up quite a ways into the history, which is something that shouldn't happen often and typically only would happen one buffer at a time.

Edit: A picture speaks a thousand words. This shows me profiling this branch vs main w/ 4 buffers open in _debug_ mode:

| fix | before |
| - | - |
| image| image |

### DIFF
--- a/data/src/message.rs
+++ b/data/src/message.rs
@@ -1615,19 +1615,6 @@ pub enum Limit {
     Since(DateTime<Utc>),
 }
 
-impl Limit {
-    pub const DEFAULT_STEP: usize = 50;
-    pub const DEFAULT_COUNT: usize = 500;
-
-    pub fn top() -> Self {
-        Self::Top(Self::DEFAULT_COUNT)
-    }
-
-    pub fn bottom() -> Self {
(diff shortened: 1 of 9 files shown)


casperstorm (approved):
This is great. From my test the difference is HUGE 🚀

PR #1078, opened 2025-07-16
Case-Insensitive Sort in Nick List

andymandias:
Use `NickRef` for `User::key` so we get case-insensitive sorting in nick lists.


andymandias:
Seems like I can't officially request a review, but so unofficially please look it over when you have a chance @4e554c4c.  Not sure if you had other plans this would meaningfully conflict with.

PR #1080, opened 2025-07-18
preview adjustment

casperstorm:
- Added tooltip to hide button
- Ensure that the hide button always is visible, even when buffer is resized
- Updated style for preview cards

### DIFF
--- a/src/buffer/scroll_view.rs
+++ b/src/buffer/scroll_view.rs
@@ -26,6 +26,8 @@ use crate::widget::{
 };
 use crate::{Theme, font, icon, theme};
 
+const HIDE_BUTTON_WIDTH: f32 = 22.0;
+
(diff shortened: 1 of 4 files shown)


src/buffer/scroll_view.rs
```diff
@@ -1318,21 +1320,30 @@ fn preview_row<'a>(
...
                     theme::button::secondary(theme, status, false)
                 }),
-        )
+            config.tooltips.then_some("Hide Preview"),
+            tooltip::Position::Top,
+            theme,
+        ))
     } else {
-        None
+        container(horizontal_space().width(Length::Fixed(22.0)))
```
andymandias: Maybe we can have a const for the width so it can be explicitly shared with the `is_hovered` width?

PR #1083, opened 2025-07-22
Serde Struct Defaults

andymandias:
Utilizes [`#[serde(default)]`](https://serde.rs/container-attrs.html#default) on (de)serialized `structs` to avoid
- Having to use functions to define defaults (often not nearby where the field itself is defined)
- Having to define field defaults in multiple places

In a few cases I moved the `impl Default` for a struct closer to the struct's definition to make it easier to see what a field's serde default value will be.

In a few cases I did not convert field `#[serde(default)]` attributes to a struct attribute where it was either inconvenient to properly implement `Default` or the existing `Default` utilized deserialization.

### DIFF
--- a/data/src/appearance/theme.rs
+++ b/data/src/appearance/theme.rs
@@ -1,4 +1,5 @@
 use std::path::PathBuf;
+use std::sync::LazyLock;
 
 use base64::Engine;
 use iced_core::Color;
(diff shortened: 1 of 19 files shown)


data/src/appearance/theme.rs
```diff
@@ -14,6 +15,12 @@ const DEFAULT_THEME_NAME: &str = "Ferra";
 const DEFAULT_THEME_CONTENT: &str =
     include_str!("../../../assets/themes/ferra.toml");
 
+static DEFAULT_STYLES: LazyLock<Styles> = LazyLock::new(|| {
+    toml::from_str(DEFAULT_THEME_CONTENT).expect("parse default theme")
+});
+
+const DEFAULT_COLOR: Color = Color::TRANSPARENT;
```
casperstorm: Nit: I think i prefer naming this TRANSPARENT_COLOR or similar so that when reading it is obvious what color it is 🤔

andymandias: Makes sense to me.  I don't expect we'll be changing it often/ever, so we wouldn't get much out of the const.

Issue #1082, opened 2025-07-22
Instructions to finish creating executable Mac app after building from source


darienm:
[This could find its way into in the documentation]

Steps to create an executable Mac app after building from source:

1. Build from source per directions at this URL:
 https://halloy.chat/installation.html#build-from-source
 Skip the 'cargo run' step at the end
2. Open folder halloy/assets/macos
3. Right-click (ctrl-click) 'Halloy.app' and 'Show Package Contents'
4. Create new folder 'MacOS' inside of 'Contents'
4. Copy 'halloy' from halloy/target/release/ to folder 'MacOS'
5. Close the 'Halloy.app' window, and you should have an executable Mac app

darienm:
Note that there are some shell scripts in halloy/scripts that appear designed to assist in the building, signing, packaging steps, but they are undocumented. So an alternative would be to document the use of those scripts and instruct in their use.

4e554c4c:
i think it might be possible to use [cargo-bundle](https://crates.io/crates/cargo-bundle) for this.

it'd also be nice to put the halloy icon in the app. i forgot how to do this on macOS though.

casperstorm:
We should just documented the scripts we have. We recently did it for Linux as well :)

PR #1086, opened 2025-07-23
Add macOS build guide to book

casperstorm:
Fixes https://github.com/squidowl/halloy/issues/1082

### DIFF
--- a/book/src/guides/macos-application.md
+++ b/book/src/guides/macos-application.md
@@ -1,0 +1,51 @@
+# Building for macOS
+
+This guide explains how to build the Halloy `.app` for macOS.
+
+## Prerequisites
+
+- **macOS 11.0 or later**
+- **Latest Rust toolchain**
+- **Xcode Command Line Tools** (xcode-select --install)
+
+## Steps
+
+1. **Install Rust and Required Targets**
+
+   Make sure you have Rust installed. Then, add the macOS targets for both Intel and Apple Silicon:
+
+   ```sh
+   rustup target add x86_64-apple-darwin
+   rustup target add aarch64-apple-darwin
+   ```
+
+2. **Clone the Repository**
+
+   If you haven’t already, clone the Halloy repository:
+
+   ```sh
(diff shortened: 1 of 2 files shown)


book/src/guides/macos-application.md
```diff
@@ -0,0 +1,51 @@
+# Building for macOS
+
+This guide explains how to build the Halloy `.app` for macOS.
+
+## Prerequisites
+
+- **macOS 11.0 or later**
+- **Latest Rust toolchain**
+- **Xcode Command Line Tools** (xcode-select --install)
```
andymandias: Is it worth providing links here?
```md
- **macOS 11.0 or later***
- [**Latest Rust toolchain**](https://www.rust-lang.org/tools/install)
- **Xcode Command Line Tools** (`xcode-select --install` or <https://developer.apple.com/download/all/>)
```

book/src/guides/macos-application.md
```diff
@@ -0,0 +1,51 @@
+# Building for macOS
+
+This guide explains how to build the Halloy `.app` for macOS.
+
```
andymandias: Maybe we could add a sentence like
```md
If you want to install a release version of Halloy on macOS, then you should consult the [Installation](installation.md#macos) documentation instead.
```
to warn off people who might end up at this guide when they're only looking to install Halloy.

darienm:
I believe lines 18-19 of Step 1 are redundant with lines 26-27 inside build-macos.sh; additionally, it might be nice if the build script asks the user if they want Intel, Apple Silicon, or Universal, rather than always making the Universal.

casperstorm:
> I believe lines 18-19 of Step 1 are redundant with lines 26-27 inside build-macos.sh; additionally, it might be nice if the build script asks the user if they want Intel, Apple Silicon, or Universal, rather than always making the Universal.

Yes it's a good low hanging fruit to make it ask which architecture to use, and default to universal in CI.

PR #1093, opened 2025-07-26
Update iced and remove widget/collection.rs

casperstorm:


andymandias (approved):
Haven't tested, but looks good to me!

PR #1106, opened 2025-08-01
Casemapped Filtering

andymandias:
Fixes to ensure each server's casemapping is utilized when testing filters.  So, for example, adding a filter for `andymandias` will block the user `Andymandias`.

Includes a minor restructuring of how filter status is stored. Moves ignored status from a HashMap of message hashes to storing the status directly in the Message (not serialized when storing messages).


casperstorm (approved):
Good change!

PR #1108, opened 2025-08-01
Logs Notification Dot

andymandias:
Adds a dot to the sidebar menu when there are unread messages in logs.  Unread is changed to only trigger for logs when the log message's level is warn or error.  This somewhat re-contextualizes log messages with warn/error level, as they are now user-notifying.  Accordingly, a couple log messages have been downgraded in severity.

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -2084,13 +2084,24 @@ impl Client {
                             };
                         }
                         Err(error) => {
+                            // Last argument is :are supported by this server
                             if index != args_len - 1 {
-                                log::warn!(
-                                    "[{}] unable to parse ISUPPORT parameter: {} ({})",
-                                    self.server,
-                                    arg,
-                                    error
-                                );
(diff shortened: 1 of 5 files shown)


casperstorm:
I get quite a bit of errors related to failed parses. See below.

image

With above im just a little unsure if i would get annoyed by that because i can't really do anything about it. We could consider changing the error level of the preview error since theres no user action i can do to fix it 🤔

andymandias:
> I get quite a bit of errors related to failed parses.

Me too, that's one of the two log messages I downgraded in severity to info (the other is when an ISUPPORT parameter is unknown).

casperstorm:
> > I get quite a bit of errors related to failed parses.
> 
> Me too, that's one of the two log messages I downgraded in severity to info (the other is when an ISUPPORT parameter is unknown).

ah lol, didnt see that 🤦🏻

casperstorm (approved):

PR #1109, opened 2025-08-02
Remove Deprecated Use from Release Action

andymandias:
Replaces a use in the release GH action that was producing deprecation warnings with the equivalent GH CLI command.  Equivalence confirmed on my fork with a [test release](https://github.com/andymandias/halloy/releases/tag/test.5).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -80,8 +80,6 @@ jobs:
   create-release:
     needs: build
     name: Create Release
-    outputs:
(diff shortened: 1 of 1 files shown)


Celeo:
FYI that link 404s, at least for me.

andymandias:
Ah, right, draft releases aren't public.  Published now, thanks for the heads up!

casperstorm (approved):

PR #1113, opened 2025-08-02
Move View-Time-Invariant Filtering Out of `history_view`

andymandias:
Since there is now an existing structure for filtering/blocking messages at load/record, we can move server message filtering into the `block_message`/`block_messages` and avoid some repetitious filtering in `history_view`.  Internal messages filtering is left in `history_view` since the filtering may depend on the time of render.  In my rough profiling via comet there was ~10% speed up in view rendering, but view rendering is so dependent on system state that I'm not especially confident in that number.  Here are flamegraphs for `history_view` before
 
Screenshot 2025-08-02 at 15-13-07 halloy – Pop!_OS 22 04 LTS – 8_2_2025 10 03 27 PM UTC – Firefox Profiler

and after

Screenshot 2025-08-02 at 15-13-18 halloy – Pop!_OS 22 04 LTS – 8_2_2025 10 02 04 PM UTC – Firefox Profiler

### DIFF
--- a/data/src/history/manager.rs
+++ b/data/src/history/manager.rs
@@ -128,6 +128,7 @@ impl Manager {
         &mut self,
         message: Message,
         clients: &client::Map,
+        buffer_config: &config::Buffer,
     ) -> Option<Event> {
         match message {
             Message::LoadFull(kind, Ok(loaded)) => {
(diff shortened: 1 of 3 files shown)


casperstorm:
@tarkah this is for you when you have time

Issue #1111, opened 2025-08-02
Mark as Read When Closing While Already at the Bottom


deepspaceaxolotl:
Thank you for making this cool program! I love the customisability. :)

Would it be possible to make it so that closing a buffer does mark it as read if one is at the bottom of the channel? Currently, if I set it to not mark as read on close (which I would prefer in case I haven't finished backreading), if I'm in a channel as messages are arriving, I have to scroll up slightly then back down to have it marked as read because of scrolling to the bottom and then I can close it, or I have to manually mark it as read. It's quite inconvenient and contrary to how most non-IRC chat software behaves nowadays.

PR #1119, opened 2025-08-05
Mark as Read on Buffer Close Iff Scrolled to Bottom

andymandias:
Adds a setting to mark as read on buffer close only if the buffer is scrolled to the bottom (as requested in #1111).  The new setting is used as the default since I suspect it's the most common expectation these days.

Moves some of the mark as read logic out of history into dashboard, in order to have ready access to the UI state.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Added:
 
 - Notification indicator on sidebar menu for warning & error log messages
+- Ability to mark as read on buffer close only if scrolled to the bottom of the buffer (new `default behavior)
 
 # 2025.8 (2025-08-31)
 
--- a/book/src/configuration/buffer.md
+++ b/book/src/configuration/buffer.md
@@ -405,15 +405,15 @@ on_application_exit = false
 
 ### `on_buffer_close`
 
-When closing a buffer (a buffer is considered closed when it is replaced or if it is open when the application exits).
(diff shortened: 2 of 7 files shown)


book/src/configuration/buffer.md
```diff
@@ -405,15 +405,15 @@ on_application_exit = false
...
 
-When closing a buffer (a buffer is considered closed when it is replaced or if it is open when the application exits).
+When closing a buffer (a buffer is considered closed when it is replaced or if it is open when the application exits).  If set to `"if-scrolled-to-bottom"` then a buffer will only be marked as read if it is scrolled to the bottom when closing (i.e. if the most recent messages are visible).
 
 ```toml
 # Type: boolean
-# Values: true, false
-# Default: true
+# Values: true, false, "if-scrolled-to-bottom"
+# Default: "if-scrolled-to-bottom"
```
casperstorm: i would leave out the "if-" on this variant.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 Added:
 
 - Notification indicator on sidebar menu for warning & error log messages
+- Ability to mark as read on buffer close only if scrolled to the bottom of the buffer (new `default behavior)
```
casperstorm: small ` here.

casperstorm (approved):
Looks good to me! Good change.

Issue #1110, opened 2025-08-02
strfttime string causes panic with some custom formatting for [buffer.date_separators]


andar1an:
I am getting a panic when trying to create specific time formats based on strfttime for buffer date separators.

```toml
# https://pubs.opengroup.org/onlinepubs/007908799/xsh/strftime.html
[buffer.date_separators]
show = true
# works - default
format = "%A, %B %-d"
# works
format = "%Y-%m-%d"
# does not work, and it appears to be length related - but not sure why after checking buffer.rs briefly
format = "%Y/%m/%d, %T %Z"

```

I tested the same formats with buffer.timestamp, and all time strings above work there.
```toml
[buffer.timestamp]
# works - default
format = "%R"
# works
format = "%Y-%m-%d"
# works
format = "%Y/%m/%d, %T %Z"
```

checking logs I don't see anything specific mentioned as to why.

andar1an:
I am also not exactly sure what buffer date separators are yet, so please excuse if I am missing something simple. I believe it is a date separator between new days.

casperstorm:
> I am also not exactly sure what buffer date separators are yet, so please excuse if I am missing something simple. I believe it is a date separator between new days.

It is! I'll investigate this bug shortly. Thanks for reporting.

andar1an:
🎉🙏

PR #1120, opened 2025-08-08
Miscellaneous Fixes

andymandias:
A few minor fixes/improvements:
- Fix for #1110, by allowing time and timezone specification in date separators' strftime string.  Adds additional validation for strftime string settings in the configuration as well.
- Parse targets in monitor online messages with `User::parse` since they may be in `nickname!user@host` format.
- Vary the notification title between highlighting the user's nickname and highlighting a [match](https://halloy.chat/configuration/highlights.html#highlightsmatch) to more easily distinguish between the notification types.
- Downgrade the log level for log messages related to disconnection, to avoid triggering unread for log messages, since disconnection messages may be encountered in regular use and we already broadcast disconnection events to the user.
- Exclude the changelog from link checks.

### DIFF
--- a/.config/lychee.toml
+++ b/.config/lychee.toml
@@ -4,5 +4,5 @@ no_progress = true
 include_fragments = true
 
 # Exclude all non-Halloy GitHub links to avoid throttling
-exclude = [ '^https://halloy\.squidowl\.org/.*', '^https://github.com/.*', '^irc://.*', '^ircs://.*' ]
-include = [ '^https://github.com/squidowl/halloy/.*' ]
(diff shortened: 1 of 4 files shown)


book/src/configuration/buffer.md
```diff
@@ -234,7 +234,6 @@ Customize how date separators are displayed within a buffer
 ### `format`
 
 Controls the date format. The expected format is [strftime](https://pubs.opengroup.org/onlinepubs/007908799/xsh/strftime.html).  
-NOTE: The application will panic if a invalid format is provided.
```
casperstorm, line 237: 🗡️

Issue #1047, opened 2025-06-27
Select current target as default for commands `/join`, `/part`, etc.


4e554c4c:
In other IRC clients, the `/join`, `/part`, etc. commands usually refer to the current target. In Halloy a target must be provided no matter what, which is somewhat annoying.

andymandias:
You can tab-complete channel names which will always match the current channel first.  Granted, that is a few more keystrokes than typing nothing at all (specifically, `<space>#<tab>`), and isn't quite the same as this request.

4e554c4c:
i didn't know about the tab complete! thanks for that

casperstorm:
Agree, we should add this QOL feature.

andymandias:
I believe we've implemented this for all commands for which a targetless form currently makes sense.  If/when there is an interface where e.g. targetless `/join` is applicable then we should be sure to add that form.  For now, I'm going to close this, but please do re-open if there's a command that was missed.

PR #1124, opened 2025-08-14
Targetless Commands

andymandias:
This PR attempts to partially address issue #1047, with the following changes:

- `/part` (`/leave`) allows omission of its target argument when sent from a channel/query buffer (with the buffer being the target being the buffer it is sent from).  Queries targeted in this way will be closed, while channels will have the appropriate `PART` command sent then closed (previously only the `PART` command was sent).  Because queries are now allowed as targets, a reason argument cannot be provided without providing a target (since the first word of a reason could not be simply distinguished from a query).
- `/topic` allows skipping its target argument, when sent from a channel buffer.
- `/cleartopic` command has been added to allow a channel's topic to be cleared (`/topic` without text will request the topic of the target, not clear it).  Its target argument can be omitted in a channel buffer.
- `/mode` allows its target argument to be skipped.  When skipping the target argument in a channel it targets the channel for a channel buffer, the other user in a query, or Halloy's user in a server buffer.
- `/kick` allows the channel argument to be skipped, when sent from a channel buffer.
- `/ctcp` allows the nick argument to be skipped, when sent from a query buffer.

~~(I think we can make at least `/kick`, `/notice`, and `/ctcp` targetless as well, but figured it would be better to get some of the structures that support targetless commands settled before adding too many more lines.)~~

### DIFF
--- a/src/buffer/input_view/completion.rs
+++ b/src/buffer/input_view/completion.rs
@@ -596,13 +596,23 @@ impl Commands {
             },
             // CTCP
             {
+                let default = current_target
+                    .and_then(|target| target.as_query())
+                    .map(target::Query::to_string);
+
                 Command {
                 title: "CTCP",
                 args: vec![
                     Arg {
                         text: "nick",
-                        kind: ArgKind::Required,
-                        tooltip: None,
+                        kind: if default.is_some() {
+                            ArgKind::Optional { skipped: false }
+                        } else {
+                            ArgKind::Required
+                        },
+                        tooltip: default.map(|default| {
+                            format!("may be skipped (default: {default})")
+                        }),
                     },
                     Arg {
                         text: "command",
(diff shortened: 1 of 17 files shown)


src/buffer/input_view/completion.rs
```diff
@@ -1056,16 +1176,38 @@ impl Command {
 #[derive(Debug, Clone)]
 struct Arg {
     text: &'static str,
-    optional: bool,
+    kind: ArgKind,
     tooltip: Option<String>,
 }
 
+#[derive(Debug, Clone)]
+enum ArgKind {
```
casperstorm: A very very minior nit would be to make this enum name a little more descriptive to what it relates to. It could just be `Argument` or similar perhaps.

casperstorm:
Not related to this but i just found that setting a mode doesnt give any feedback.
So when i do `/mode +i` i cant see it worked, but doing `/mode` after confirms it was set.

We should add a message when it was changed. Doesn't have to be this PR.

casperstorm (approved):
Ive confirmed all is worked as expected!

PR #1128, opened 2025-08-18
Add display bounds check for window position

### LINKED ISSUE #1122: Opens to seconday monitor in a single-monitor setup
OS: Windows 11 24H2

How to reproduce:
Have two or more monitors connected. Move Halloy to eg. monitor 2.
Shutdown system and travel to a very VIP LAN party, but have only one monitor hooked up.
Open Halloy and it will now open to the not connected monitor. It will shortly show the animation of it flying offscreen from the primary monitor to the secondary monitor, if selected from the taskbar.

Halloy can't be moved back by using win+arrow keys or win+shift+arrow keys, properly because to the OS there is only one monitor connected. Halloy can "recovered" back to the primary screen via Windows Task manager -> Processes - Apps -> Halloy IRC Client -> Halloy -> Right click and select Maximize.

Video: [https://drive.google.com/file/d/1liP_3tFdvPKacSi97yY798N2csBNCYJh/view?usp=sharing](url)

casperstorm:
Introduces the 'display-info' crate to validate window positions against visible screen bounds. 

Fixes #1122

### DIFF
--- a/data/src/window.rs
+++ b/data/src/window.rs
@@ -43,7 +43,7 @@ impl Window {
         let size = size.max(MIN_SIZE);
         let position = position
             .filter(|pos| pos.y.is_sign_positive() && pos.x.is_sign_positive())
-            .filter(|pos| is_position_valid(*pos, size));
+            .filter(|pos| is_position_valid(*pos));
 
         Ok(Window { position, size })
     }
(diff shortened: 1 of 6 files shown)


data/src/window.rs
```diff
@@ -67,6 +68,34 @@ fn path() -> Result<PathBuf, Error> {
...
+            && position.y < (display.y + display.height as i32) as f32
+        {
+            return true;
+        }
+    }
+
+    // Window position is not within any display bounds
+    false
+}
+
```
andymandias: Maybe we could use [`from_point(position.x, position.y).is_ok()`](https://docs.rs/display-info/latest/display_info/struct.DisplayInfo.html#method.from_point) for position checking?  In particular, [they appear to do something non-intuitive for Wayland](https://github.com/nashaofu/display-info/blob/f781e0e2aa37c4da5fb9fe22eee0d696195abf51/src/linux/wayland.rs#L131) (I'm not currently running Wayland, so I haven't been able to double-check that their Wayland position checking is correct, but I would assume so).

andymandias (approved):
I've only done minimal testing;  can confirm it prevents an attempted off-screen placement in X11 (off-screen placement is prevented by the WM even when requested).

One minor thing I noticed while testing, which I don't think is new behavior, is that window position appears to be recorded as requested (0,0 when invalid) rather than where the WM ultimately places it (which Mutter happens to not put at 0,0 for some reason :shrug:).  As soon as the window is moved, then the position is recorded.  This could just be a quirk of how Mutter and/or X11 works, and regardless does not need to be addressed in this PR.

data/src/window.rs
casperstorm: Much better. Didnt see that one. Confirmed it also works.

Issue #983, opened 2025-05-08
Binary space partition split for Panes


seraxis:
This feature would add a "binary" configuration for the split axis for panes.

The base case would be a single pane belonging to the window root, every split operation would split the pane evenly on it's longest axis (implicitly biased to the right and down).

This would utilise window space more effectively regardless of window aspect ratio, and would save the user time performing tedious windowing operations.

PR #1135, opened 2025-08-24
Shorter & Largest-Shorter Split Axis Settings

andymandias:
Adds two new settings for `pane.split_axis`:
- `"shorter"` which takes the shorter dimension of the focused pane as the split axis.  E.g. if the pane is taller than it is wide, then the pane will be split horizontally.
- `"largest-shorter"` which will find the largest splittable pane (i.e. pane in the main window), and take its shorter dimension as the split axis.

I switched the default `split_axis` setting to `"shorter"` as I suspect it will be the most often preferred.

I believe one of these should address #983.


casperstorm (approved):
I like this, the naming is hard tho but i cant find any better names so all good for me :D

PR #1143, opened 2025-08-28
Hide Secondary Unread Indicators

andymandias:
Adds the fix and feature requested in #979.  Specifically:
- Messages that are hidden/blocked no longer cause the "Mark messages as read" button to activate.  (This utilizes the changes made in #1113, which is why this PR is branching off of that PR.)
- A setting to hide the backlog divider when all messages have been marked as read (similar to how Discord handles its backlog divider).  This has a potential for confusion (i.e. a user familiar with the current behavior could wind up looking for the divider for a while, not realizing it's hidden), so I am leaving the default setting as false.

### DIFF
--- a/book/src/configuration/buffer.md
+++ b/book/src/configuration/buffer.md
@@ -8,6 +8,7 @@ Buffer settings for Halloy.
    3. [Topic](#bufferchanneltopic) - Topic settings within a channel buffer
 2. [Chathistory](#bufferchathistory) - IRCv3 Chat History extension settings
 3. [Commands](#buffercommands) - Commands settings
+4. [Backlog Separator](#bufferbacklog_separator) - Customize when the backlog separator is displayed within a buffer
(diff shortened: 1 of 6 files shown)


book/src/configuration/buffer.md
```diff
@@ -227,6 +228,23 @@ Show or hide the description for a command
...
+Hide backlog divider when all messages in the buffer have been marked as read.
+
+```toml
+# Type: boolean
+# Values: true, false
+# Default: false
+
+[buffer.buffer_separator]
+hide_when_all_read = true
+```
```
casperstorm, line 246: Should this be true as default? 🤔

andymandias: I picked false as default because I thought it could be confusing for a user used to the current behavior.  That is, I was concerned they might go searching through the channel history for the backlog divider and never find it.  I don't feel strongly about it though, would be totally okay with defaulting to true.

casperstorm: Im good with your initial thought!

PR #1148, opened 2025-08-31
Add `libxcb-devel` to Fedora install docs/script

hashcatHitman:
`libxcb-devel` is needed to build on Fedora now. So, it has been added to the installation documentation and Linux install script.


andymandias (approved):
Looks good, thanks!

PR #1154, opened 2025-09-06
Add Normalized Representation to Nicknames

andymandias:
This PR has two main purposes:
- To fix a minor regression, where nicknames highlighting/coloring is case sensitive.  I.e. `Andymandias` would not be matched against the user `andymandias`.  The pre-regression (intended) behavior is for nickname matching to be case insensitive.
- A minor improvement to case insensitive matching, to use the server's casemapping (as supplied by ISUPPORT parameter).  The pre-regression behavior was for the casemapping used for case insenstiive matching is [`rfc7613`](https://modern.ircdocs.horse/#casemapping-parameter), regardless of received ISUPPORT casemapping parameter.

To those ends, the main change introduced by this PR is converting
```rust
pub struct Nick(String);
```
to
```rust
pub struct Nick {
    raw: String,
    normalized: String,
}
```
where `normalized` is specified from `raw` using `isupport::CaseMap`.  Accordingly, associated hasing and ordering Traits are modified to use `normalized`.  All other changes are downstream from the changes made to `Nick`.

This PR does not attempt to handle the possibility of a network's casemapping changing while connected, both to keep the scope of changes as small as possible and because it is not really an expected event.

### DIFF
--- a/data/src/history/manager.rs
+++ b/data/src/history/manager.rs
@@ -519,6 +519,7 @@ impl Manager {
                 user,
                 comment,
                 user_channels,
+                casemapping,
             } => {
                 let user_query =
                     queries.find(|query| user.as_str() == query.as_str());
@@ -529,6 +530,7 @@ impl Manager {
                     &user,
                     &comment,
                     config,
+                    casemapping,
                     sent_time,
                 )
             }
@@ -537,6 +539,7 @@ impl Manager {
                 new_nick,
                 ourself,
                 user_channels,
+                casemapping,
             } => {
                 if ourself {
                     // If ourself, broadcast to all query channels (since we are in all of them)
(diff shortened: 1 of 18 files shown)


data/src/history/manager.rs
```diff
@@ -756,7 +773,16 @@ impl Manager {
                                 let nick = match source.nick() {
                                     Some(nick) => Some(nick.clone()),
                                     None => message.plain().and_then(|s| {
-                                        s.split(' ').nth(1).map(Nick::from)
+                                        let casemapping = kind
+                                            .server()
+                                            .map(|server| {
+                                                clients.get_casemapping(server)
+                                            })
+                                            .unwrap_or_default();
```
casperstorm: Nit: Maybe we should have a fn to return this.

casperstorm (approved):
I did not test this, however changes look good. Good clean up as well i feel like.

PR #1158, opened 2025-09-07
Updated iced

casperstorm:
Updated iced to latest master.


andymandias (approved):
Only tested briefly, but everything looks good!

PR #1160, opened 2025-09-09
Fix Highlight Deserialization

andymandias:
Deserialization of highlights could fail due to a change in the `data::server::Server` `struct`.  This PR fixes backwards compatibility for highlights and adds a highlight to serde tests (and generates new test JSON).

Edit: also fixes dashboard deserialization.

### DIFF
--- a/data/src/message.rs
+++ b/data/src/message.rs
@@ -1970,6 +1970,8 @@ pub mod tests {
         parse_fragments_with_users,
     };
     #[allow(unused_imports)]
+    use crate::bouncer::BouncerNetwork;
+    #[allow(unused_imports)]
     use crate::config::Highlights;
     #[allow(unused_imports)]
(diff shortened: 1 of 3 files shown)


data/src/message.rs
```diff
@@ -2512,17 +2549,17 @@ pub mod tests {
...
-        for file in fs::read_dir("tests/message/").unwrap() {
-            let path = file.unwrap().path();
-            let file_string = fs::read_to_string(&path).unwrap();
-            let _messages: Vec<Message> =
-                serde_json::from_str(&file_string).unwrap();
-        }
-    }
+    // // Test Message deserialization from samples of messages serialized by
+    // // earlier versions (i.e. backward compatibility)
+    // #[test]
```
casperstorm: Is this something we want to remove or keep?

casperstorm (approved):

PR #1162, opened 2025-09-09
Fix Flathub links

Sporiff:
Looks like Flathub updated something and now Lychee isn't happy.

https://github.com/squidowl/halloy/actions/runs/17593468229/job/49979760622

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Join **#halloy** on libera.chat if you have questions or looking for help.
     <img src="https://repology.org/badge/vertical-allrepos/halloy.svg" alt="Packaging status">
 </a>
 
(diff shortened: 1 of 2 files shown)


englut:
Hm, https://flathub.org/apps/org.squidowl.halloy loads fine for me. It does redirect to the `/en/` URL. Not sure if that's what it was doing before, but likely, we want to preserve the user's language choice and not have english dictated to users.

Maybe it's just a hiccup.

I ran lychee locally:
```
> lychee --config ./.config/lychee.toml './book/**/*.md'

...

🔍 239 Total (in 6s) ✅ 161 OK 🚫 0 Errors 👻 78 Excluded
```

Maybe we just need to re-run those actions.

Sporiff:
@englut Yep, looks like it was a transient issue on Flathub's side. When I tested it yesterday those links were definitely down, but they're back in action now.

Issue #499, opened 2024-08-24
permit relative path or same-directory search for nick_password_file


dmd:
- I keep all my config files in a git repo. 
- I use the `password_file` functionality
- On one of the computers I want to use halloy on, my username is different from the others (at work it's my last name instead of my initials)

I can't use halloy on the work computer, because `password_file` [has to be an absolute path](https://github.com/squidowl/halloy/pull/381).

If shell expansion would be a lot of work, at least allow relative path - or even no path at all (i.e., if there is no path, assume the file is in the same directory as `config.toml`?

o-alquimista:
Path relative to the `config.toml` directory would be nice.

```toml
[servers.liberachat.sasl.plain]
username = "john_doe"
password_file = "passwords/liberachat"

[servers.oftc.sasl.external]
cert = "certs/oftc.cer"
key = "certs/oftc.key"
```

PR #1169, opened 2025-09-12
prefix relative paths with config dir

5GameMaker:
Addresses #499.

Allows using relative paths in any of the config parameters that accept a path (at least from the ones I've found).

### DIFF
--- a/data/src/config.rs
+++ b/data/src/config.rs
@@ -397,14 +397,14 @@ impl Config {
         let config_dir = Self::config_dir();
         let prefix_with_config_dir = |x: &mut PathBuf| {
             if x.is_relative() {
-                *x = config_dir.join(&mut *x)
+                *x = config_dir.join(&mut *x);
             }
         };
         let prefix_with_config_dir_opt = |x: &mut Option<PathBuf>| {
-            if let Some(x) = x.as_mut() {
-                if x.is_relative() {
-                    *x = config_dir.join(&mut *x)
-                }
+            if let Some(x) = x.as_mut()
+                && x.is_relative()
+            {
+                *x = config_dir.join(&mut *x);
             }
         };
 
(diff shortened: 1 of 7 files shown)


data/src/config.rs
```diff
@@ -194,6 +195,28 @@ where
...
+/// This is not panic-safe if panic occures after `map` drops the
+/// temporary value.
+fn apply<T>(x: &mut T, mut map: impl FnMut(T) -> T) {
+    unsafe {
+        let mut tmp: MaybeUninit<T> = MaybeUninit::zeroed();
+        tmp.as_mut_ptr().copy_from(x, 1);
+        tmp = MaybeUninit::new(map(tmp.assume_init()));
+        std::mem::swap(x, tmp.assume_init_mut());
+    }
+}
```
andymandias: I'm not sure I understand the intention behind this function.  Is there some reason paths cannot be prefixed without resorting to `unsafe` code?

5GameMaker: No, that was just the least annoying way to implement it (Rust does not allow moving out and in of mutable ref `x.field = map(x.field)`).

Albeit unsafe code can be avoided here by having `apply` be not generic.

(edit: nvm I was just doing it wrong apparently)

andymandias (approved):
Looks good to me and works well in my testing.  Just needs an entry in the CHANGELOG now I think.

5GameMaker:
`file_transfer.save_directory` is not affected by any expansions, is that an intended behavior?

andymandias:
> `file_transfer.save_directory` is not affected by any expansions, is that an intended behavior?

Good catch!  That is an oversight.  A `#[serde(deserialize_with = "deserialize_path_buf_with_path_transformations_maybe")]` would be appropriate for that setting (along with a footnote for the setting in `book/src/configuration/file_transfer.md`).

andymandias:
Nice work!  Thanks for the PR.

PR #1171, opened 2025-09-13
Page Up/Down Fixes

### LINKED ISSUE #1167: PgUp/PgDown scroll inconsistencies
Hitting PgUp or PgDown won't scroll by a screenful of messages. It won't even scroll by the same amount on each press, and the amount is different in different buffers too. It could range between half and 3.5 pages. No idea what additional info to include. Using v2025.08 from Flathub with soju v0.9.0

Edit:
- commenting out all `buffer.server_messages.*` customizations changes the amount of lines scrolled, but it's just different, still not correct.
- if I continue to scroll back, it will just refuse to do it after several presses. The only way to continue is to use mouse wheel, trackpoint, etc. At first I thought it is busy fetching the history from soju, but it works fine with the trackpoint or the trackpad.

andymandias:
Fixes #1167.

Two changes to scroll view behavior are made to address the reported behavior:
- The page up/down `scroll_by` requests are no longer clamped.  The clamping vector that was used rounds to the nearest pixel, so if the scroll view was within half a pixel from the top then the `scroll_by` request would be rounded to zero (and the top could never be reached).  Iced appears to handle the unclamped values fine, so I just left them unclamped rather than trying to improve the clamp bounds.
- Modified the load logic to ensure that a page up/down's target is always loaded.  I.e. that enough messages are loaded such that a page up/down can be performed without loading more messages.  When paging up/down past the loaded messages the scroll position was getting lost.  Attempting to persist the `scroll_by` across loading messages proved to be very complicated, so instead I opted to make simpler changes to ensure that paging past available messages doesn't happen.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,11 +29,12 @@ Fixed:
 - Do not activate the mark as read buffer when blocked/hidden messages are unread in the buffer
 - Nickname highlighting is case insensitive by default (and uses the server's specified casemapping)
 - Nickname changes properly broadcast in all channels user is in
+- Page up/down behavior moves by consistent page length, and will no longer get stuck at the top of a buffer
 
 Thanks:
 
 - Contributions: @Death916, @4e554c4c, @freakyy85, @hashcatHitman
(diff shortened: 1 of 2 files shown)


casperstorm (approved):
Very nice way to solve it.

PR #1172, opened 2025-09-14
Update Linux config paths in documentation

casperstorm:


apollo7321:
Looks good, thanks for the addition!

As said in IRC: perhaps using $HOME would fit better, because you already use $HOME for `$HOME/.config/halloy`

PR #1177, opened 2025-09-16
Add Flatpak Location for Log File

andymandias:
Adds the location of the log file for Flatpak to the documentation.


andymandias:
Flathub is the most popular place where one downloads Flatpaks.  Since our Flatpak can be downloaded from Flathub they both work, but I think Flatpak is technically more correct.

PR #1183, opened 2025-09-17
Notification delays per kind

Toby222:
Instead of timers matching on every field, just match on the notification kind.


data/src/config/notification.rs
```diff
@@ -10,7 +10,7 @@ pub struct Notification {
     pub show_toast: bool,
     pub show_content: bool,
     pub sound: Option<String>,
-    pub delay: Option<u64>,
+    pub delay: Option<i64>,
```
andymandias: This type defines how the field is read from the configuration file, so changing it implicitly changes what values are allowed to be specified.  I don't think we want to allow negative delays.  We can use a `u32` if we want to be sure it'll convert cleanly to an `i64` (the maximum ~49 day delay should be long enough to cover most people, I think).

src/notification.rs
```diff
@@ -178,14 +211,15 @@ impl Notifications {
         title: &str,
         body: impl ToString,
     ) {
+        let now = Utc::now();
+        let notification_kind = notification.into();
         let last_notification =
-            self.recent_notifications.get(notification).copied();
+            self.recent_notifications.insert(notification_kind, now);
```
andymandias: Inserting here instead of later means that the delay will be restarted even if we don't display a notification.  In a pathological case notifications could be delayed indefinitely.

src/notification.rs
```diff
@@ -11,8 +10,42 @@ use crate::audio;
...
+            }
+            Notification::MonitoredOnline(..) => {
+                NotificationKind::MonitoredOnline
+            }
+            Notification::MonitoredOffline(..) => {
+                NotificationKind::MonitoredOffline
+            }
+        }
+    }
+}
```
andymandias: I wonder if we should have `DirectMessage` include the user in it's kind still, to allow for direct messages notifications from multiple users in quick succession (but limit to one notification per user).  We could potentially add another field to `Highlight` that specifies match pattern that was matched, for the same purpose.  @quaff, if I recall correctly you worked on this feature originally, do you have a preference?

andymandias (changes requested):
Thank you for finding and fixing this!  There are a few changes I think we should make, but I think this is close.

src/notification.rs
Toby222: I considered that originally but as I then considered other notification types that may want to be delivered independently I remembered that the primary point of the delay is to not spam notifications at the user, and unless there's config options to more cleanly define what kind of message resets what timer, it seems prudent to only go off of message kind

englut: Good call out, @andymandias. I think the reasoning I had for just hashing a copy of the notification, was to keep track of the target (in this case, user, but can also apply to channels). With the previous logic, the delay would apply to a single user/channel. But with this updated logic, you're right, it would be applied to the notification type, and not context aware. I think I would prefer if the notification delay stayed context aware. So updating the Kind to include the target would be good!

englut: @Toby222 I'm not tied to the context aware delay. I do think it works better for my use-case, but if you and/or others would rather delays happen on notification type, then I am cool with that!

andymandias: > I considered that originally but as I then considered other notification types that may want to be delivered independently I remembered that the primary point of the delay is to not spam notifications at the user, and unless there's config options to more cleanly define what kind of message resets what timer, it seems prudent to only go off of message kind

Fair point.  We want to balance it against blocking too many notifications though, since it will also avoid getting a single notification from two separate users within a short duration as it currently is.  Maybe we could balance between these two concerns by having another check against `recent_notifications.len()`, blocking the notification if there have been above some number of notifications recently?

Toby222: I tried to do it properly, also to make it work better for MonitoredOnline/Offline with multiple users

src/notification.rs
```diff
@@ -39,44 +78,40 @@ impl Notifications {
...
-                        server,
-                    );
-                });
+                self.execute(
+                    &config.monitored_online,
+                    notification,
+                    "Monitored users are offline",
+                    &targets.iter().join(", "),
+                );
             }
```
andymandias, line 124: This is a nice update for potentially avoiding notification batches.  It's a little tedious, but let's take it one step further and customize the message based on the number of targets.  Specifically, pluralize user only if targets is greater than one, separate with commas only if the number of targets is greater than two, and add "and" between the second to last and last target.

andymandias (changes requested):
Looks good to me!  One minor extension I think would be nice to add, and also an entry in the CHANGELOG, then I think we'll be good to merge.

src/notification.rs
```diff
@@ -39,44 +78,48 @@ impl Notifications {
...
-                        notification,
-                        &format!("{} is online", target.nickname()),
-                        server,
-                    );
-                });
+                self.execute(
+                    &config.monitored_online,
+                    notification,
+                    if targets.len() == 1 {
+                        "Monitored user is online"
```
englut, line 106: I think this is a case where having the specific target nick shown would be good.

src/notification.rs
```diff
@@ -39,44 +78,48 @@ impl Notifications {
...
-                        notification,
-                        &format!("{target} is offline"),
-                        server,
-                    );
-                });
+                self.execute(
+                    &config.monitored_online,
+                    notification,
+                    if targets.len() == 1 {
+                        "Monitored user is offline"
```
englut, line 117: Same comment as before.

src/notification.rs
Toby222: Not sure I understand? The target nick is shown in the body, why move it to the title here?

englut: Ah. I see, right. The parameter below this one is the body. Apologies, I should have paid more attention to that!

andymandias (approved):
To not take up much more of your time I pushed the other two changes I requested from the last review.  If you're ok with those changes, then I think this is good to merge.

Toby222:
> To not take up much more of your time I pushed the other two changes I requested from the last review.  If you're ok with those changes, then I think this is good to merge.

Nice :3

andymandias:
Great!  @casperstorm, did you want to weigh in?

casperstorm:
All good from me

PR #1194, opened 2025-09-19
Update feature flags

4e554c4c:
The `rodio` crate, used by halloy to play sounds recently changed its decoders. I saw a whole lot of these `symphonia-*` dependencies being pulled into the build, however we only need the ogg vorbis decoder since halloy plays ogg files for sounds. This change should slightly speed up the build.

Test plan:

Was highlighted in a channel and heard a sound.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2099,12 +2099,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "extended"
-version = "0.1.0"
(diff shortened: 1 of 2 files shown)


andymandias:
Users are able to specify their own sound files for notifications ([notifications.sound](https://halloy.chat/configuration/notifications.html#sound)).  I believe this would effect the file types that would be supported with that feature.

4e554c4c:
I see! ok no use for this then.

Issue #1193, opened 2025-09-19
Channel Names Blank/Sidebar Too Narrow


deepspaceaxolotl:
Since the latest update, both with my own theme and the default one, the channel names are gone and the sidebar is really narrow (I've set the sidebar `max_width = 240`, more than the default, but even clearing my entire config file doesn't help). It still reacts like there are buttons there (to hovering or selection), but no text is visible.

<img width="114" height="815" alt="Image" src="https://github.com/user-attachments/assets/0334e9c4-fb9f-405a-a295-7a00841399d9" />

casperstorm:
Are you using the default font?

deepspaceaxolotl:
Oh, interesting, that seems to be it! Only saw the effects after restarting. But I didn't have this problem before when I first set this up.

ian-perry-mia:
Yeah, I'm also having this issue. Setting the font to default appears to help it. Running macOS 15.6.1.

ian-perry-mia:
Alright, on further look, the issue is not setting the font family, it's font size. Anything over 17 causes the problem, at least on my machine. And regardless, I can't see channels in the buffer at all.

deepspaceaxolotl:
Mine's set to 16.

dieperdev:
Same issue. It happens on font sizes larger than 17 (font size 17 works fine). Not sure if it is Windows only.

ian-perry-mia:
> Same issue. It happens on font sizes larger than 17 (font size 17 works fine). Not sure if it is Windows only.

Definitely not, happening on macOS for me.

PR #1196, opened 2025-09-20
Blank Sidebar Fix

andymandias:
When the text in the sidebar is bigger than the specified fixed height of the container it is in then it will fail to render (as seen in issue #1193).  This removes the fixed height of sidebar buttons, instead having sidebar buttons' height be determined by the text it contains.  Specifically, the size of the `Stack` that the button contains is determined by its first element (the text), so the icon does not change the button height.

Since the font size can get quite large I removed the cap on the icon size.  It is possible to get the icons to fail to render if the icon size is larger than the line height of the text, but I believe we're stuck with that behavior so long as we're using a `Stack`.

I tried to recreate the look of the existing sidebar as best as I could, but it is not exactly the same as before.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,11 @@
 
 Fixed:
 - Enforce delay between notifications
+- Larger fonts (and font sizes) can be used without blanking out the sidebar
 
 Thanks:
 - Contributions: @Toby222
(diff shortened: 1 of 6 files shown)


casperstorm (approved):
Tested and approved.

PR #1197, opened 2025-09-21
Sysinfo command to display system informations

### LINKED ISSUE #1181: [feature] Implement sysinfo stats
Can be possible to implement sysinfo as addon or build as optional?

casperstorm:
Fixes #1181 

Added `/sysinfo` command which displays the following:
CPU, Memory, GPU, OS, and Uptime.

They can be controlled by the following config.
They are all enabled (true) by default

```toml
[buffer.commands.sysinfo]
cpu = true/false
uptime = true/false
memory = true/false
gpu = true/false
os = true/false
```


casperstorm:
This simply uses iced (which uses winit) to pass sysinfo. I think this is good enough, however i was also looking at: https://crates.io/crates/sysinfo - but i think this is overkill.

andymandias (approved):
Looks good, works well in my testing.  Using an already included crate makes sense to me, and I don't see any important information missing.

Would be good to add to the table in `commands.md`, and a CHANGELOG entry.

PR #1199, opened 2025-09-21
Enable Primary Clipboard on Linux

### LINKED ISSUE #1189: Halloy doesn't honour primary selection
Put simply, one can't paste the primary selection in a halloy buffer with a mouse middle click. And there's no keyboard shortcut documented. Also one cannot paste the text selected in halloy elsewhere by the same method.

I saw a few other issues concerning text selection, like https://github.com/squidowl/halloy/issues/154 and https://github.com/squidowl/halloy/issues/212. As I don't know if they are related, I opened a new one instead of hijacking one of those.

Links:
- https://wayland.app/protocols/primary-selection-unstable-v1
- https://tronche.com/gui/x/icccm/sec-2.html
- https://www.jwz.org/doc/x-cut-and-paste.html

andymandias:
Enables primary clipboard on Linux.  Support in `TextInput` widgets is better done in iced itself, in my opinion, so I added a couple commits to our iced patch to do so.  Those patches allow pasting from the primary clipboard with the middle mouse button (with the paste inserted at the mouse position) and automatically write selections to the primary clipboard.  The changes in Halloy itself are to enable selections in `selectable_text` to be copied into the primary clipboard.

I believe this will fix #1189.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,11 @@ Added:
 Fixed:
 - Enforce delay between notifications
 - Larger fonts (and font sizes) can be used without blanking out the sidebar
+- The primary clipboard (with copy on selection & paste with middle click) is supported on Linux
 
 Thanks:
 - Contributions: @Toby222
-- Bug reports: @Toby222, @deepspaceaxolotl
+- Bug reports: @Toby222, @deepspaceaxolotl, @zhelezov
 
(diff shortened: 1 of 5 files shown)


zhelezov:
Tested on Fedora 42, this fixes https://github.com/squidowl/halloy/issues/1189

PR #1207, opened 2025-09-24
Added url context with copy url option

casperstorm:


src/buffer/context_menu.rs
```diff
@@ -63,140 +70,154 @@ impl Entry {
         server: &Server,
         prefix: &[isupport::PrefixMap],
         channel: Option<&target::Channel>,
-        user: &User,
+        user: Option<&User>,
         current_user: Option<&User>,
+        url: Option<&String>,
```
andymandias, line 75: Looking ahead to potentially putting context menus on more items (e.g. could be nice to have a context menu on timestamps to copy them in `YYYY-MM-DDThh:mm:ss.sssZ` format for use in chathistory commands), would it make sense to have an `Enum` that defines the context menu's... context?  With `Context::User` and `Context::Url` that contains the information needed/expected for that context?

andymandias (approved):
Looks good and works well for me.  Had one idea for restructuring the `view` interface slightly, but even if we do restructure it doesn't need to be done in this PR.

Also, noticed a couple long lines so I ran the `context_menu.rs` through `rustfmt` (no functional changes).

src/buffer/context_menu.rs
casperstorm: I think thats a good idea. Ill merge this now, and then if you get to add it before me feel free - otherwise ill tackle it later 👍🏻

PR #1208, opened 2025-09-25
Condensed Messages

andymandias:


casperstorm:
Great initial draft, @andymandias.
A few things from me:

image

1. Visually it is a bit overwhelming at the moment. I think it is the double signs (-+)
  a. Perhaps we should only show one sign: if they join, then leave we show "-nick" if they leave, then join we show "+nick"? 
2. Perhaps color the people who leave with the "offline" color would help the overview
3. Color the signs would help differenciate them
4. Perhaps we should remove the "dot" when doing condensed message.
5. For the config, perhaps we make it a struct instead so we further down the road could add the others in as well (change host, nickname, mode etc)

andymandias:
Some updates have been pushed.  I agree we should color the join/part/quit symbols, which is what I'll be looking at next.  I'd also like to add a setting to dim condensed messages (similar to senpai's rendering style).

Removed the dot as you suggested, I think it looks better (and helps distinguish a condensed and non-condensed server message).

Re: multi-signs.  I'm hesitant to condense `-+user` into `+user` since the latter implies (to me) that `user` was not in the channel before the condensed message.  What I've done currently is add a `buffer.server_messages.condense.format` setting which can be `"detailed"` or `"brief"`.  If `"brief"` is selected (its also the default), then `-+user` and `+-user` are just removed from the condensed message entirely (since the user's state does not change).

book/src/configuration/buffer.md
```diff
@@ -767,6 +767,60 @@ Note: Not all server messages uses this setting.
...
+
+The color and font style of the symbols is taken from the theme setting for that event type.
+
+```toml
+# Type: array of strings
+# Values: ["join", "part", "quit"]
+# Default: []
+
+[buffer.server_messages.condense]
+messages = ["join", "part", "quit"
```
casperstorm: Nit: missing `]`

casperstorm (approved):
This is a massive feature. Works really well.
I almost feel like we should add a guide to it somehow so people actually will see it.

PR #1211, opened 2025-09-27
Fix buffer link in pane documentation

apollo7321:
On https://halloy.chat/configuration/pane.html there is a link to https://halloy.chat/configuration//buffer.html which causes rendering issues. Removing the double slash fixes it.


casperstorm (approved):
Thanks for fixing this.

PR #1215, opened 2025-10-03
Anti-Flood Protection

andymandias:
Adds a simple token bucket style rate limiter for anti-flood prevention for user-sent messages.  I.e. to prevent the user from sending too many messages by accident and being kicked for flooding.  From what I've seen a common rate limiting scheme has an overall rate of one message per two seconds with an allowance for bursts of messages (e.g. from what I can tell from their codebases, soju and weechat both use that base rate).  So, this PR sets the limit to one message per two seconds with bursts up to eight messages.

If we want to incorporate messages sent by Halloy (WHO polls, MARKREADs, etc) into anti-flood measures, then WHO polling logic in particular should be revisited again.  Otherwise WHO polling could wind up being throttled in such a way that results in an ever-increasing backlog.  This PR sidesteps that by just considering user-sent messages.

If the ISUPPORT parameter [`soju.im/SAFERATE`/`SAFERATE`](https://codeberg.org/emersion/soju/src/branch/master/doc/ext/saferate.md) is present, then rate limiting is disabled.


data/src/client.rs
```diff
@@ -2998,6 +3002,10 @@ impl Client {
     }
 
     pub fn tick(&mut self, now: Instant) -> Result<()> {
+        if let Some(ref mut anti_flood) = self.anti_flood {
```
apollo7321, line 3005: I am not an expert in Rust, but putting anti_flood inside a Some seems a bit unecessary. You could probably use `matches!` here: `if matches!(self.anti_flood, Some(_)) {`


(I could be wrong here though)

casperstorm (approved):
Code make sense, and it works great in my testing.

casperstorm:
Perhaps we add a entry to CHANGELOG?

casperstorm (approved):

PR #1221, opened 2025-10-05
Filters Fixes

andymandias:
Minor bug fixes for filters:
- Filters are applied only to the servers they are specified for
- Filters are renormalized properly when ISUPPORT is updated (and open histories are renormalized properly when ISUPPORT is updated, so that updated filters can be applied)


casperstorm (approved):
Looks good

PR #1228, opened 2025-10-09
Support for CTCP query `USERINFO`

Bartkk0:
Adds support for the `USERINFO` CTCP query, disabled by default when the userinfo config option is left unset.

### DIFF
--- a/book/src/configuration/ctcp.md
+++ b/book/src/configuration/ctcp.md
@@ -73,6 +73,9 @@ Whether Halloy will respond to a [CTCP USERINFO](https://modern.ircdocs.horse/ct
 # Values: any string
 # Default: not set
 
+# Example usage: <nickname> (<realname>)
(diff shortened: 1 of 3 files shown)


book/src/configuration/ctcp.md
```diff
@@ -63,3 +63,16 @@ Whether Halloy will respond to a [CTCP VERSION](https://modern.ircdocs.horse/ctc
...
+
+Whether Halloy will respond to a [CTCP USERINFO](https://modern.ircdocs.horse/ctcp#userinfo) message. The response is enabled if this option is set to a string, which will be used as the reply.
+
+```toml
+# Type: string
+# Values: any string
+# Default: not set
+
+[ctcp]
+userinfo = true
```
andymandias: We should provide a valid example with a string.  Potentially `<nickname> (<realname>)` as on horsedocs, unless there's a more relevant example.

casperstorm:
Agree with, @andymandias - this is great. Thank you so much.

PR #1243, opened 2025-10-17
Automatically Request Channel Modes

andymandias:
Automatically sends a mode request when joining a channel (if the channel mode is not sent by server/bouncer shortly after joining).  The request response is used to populate channel metadata (to be able to show channel modes in the title bar and potentially elsewhere), and is otherwise hidden from the user.

Re-orders the tick logic a little bit to allow tick generated messages to enter the anti-flood message queue before it is processed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Added:
 - `/detach` command to detach from channels (soju feature)
 - Support for USERINFO CTCP query
 - Configurable anti-flood protection for servers that do not advertise SAFERATE
+- Channel modes will be retrieved on joining a channel (to be shown in the pane title bar)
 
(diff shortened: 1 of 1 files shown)


data/src/client.rs
```diff
@@ -2223,7 +2236,21 @@ impl Client {
                 ) && let Some(channel) =
                     self.chanmap.get_mut(&target_channel)
                 {
+                    let mode_request_response = if let Some(position) =
+                        self.mode_requests.iter().position(|mode_request| {
+                            mode_request.channel == target_channel
+                        }) {
+                        self.mode_requests.swap_remove(position);
```
casperstorm, line 2243: :chefkiss:

PR #1256, opened 2025-10-22
New docs structure

casperstorm:
Structured the docs a bit differently to make it easier to get a overview.

test it here: https://golden-mochi-30ae41.netlify.app/

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -27,7 +27,7 @@ jobs:
         id: lychee-interal
         uses: lycheeverse/lychee-action@v2
         with:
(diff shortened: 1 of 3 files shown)


casperstorm:
@andymandias something fishy with our internal link checker. Links does work, but keeps complaining.

andymandias:
> @andymandias something fishy with our internal link checker. Links does work, but keeps complaining.

The errors I'm seeing so far are it being overly strict by not allowing equivalent links for files. So, if the file is README.md it wants the link to be to README.md even if index.md or index.html will get routed appropriately.  I'll see if there isn't a way to relax that.

PR #1259, opened 2025-10-23
Renormalize Messages in `history::load`

andymandias:
Moves renormalization of messages (to ensure they match the server's casemap) into `history::load`, so it cannot be easily skipped.  Stops short of implementing `DeserializeSeed` to have stateful deserialization of messages, because I don't yet see a way to do so without manually implementing deserialization for every type that contains a `Nick` (resulting in a *lot* of boilerplate code to maintain).

Fixes an issue where `history::append` would not renormalize messages, potentially resulting in duplicate messages when the server's casemap is not RFC7613.

### DIFF
--- a/data/src/history.rs
+++ b/data/src/history.rs
@@ -179,8 +179,8 @@ pub async fn load(kind: Kind, seed: Option<Seed>) -> Result<Loaded, Error> {
     let mut messages = read_all(&path).await.unwrap_or_default();
 
     if let Some(seed) = seed {
-        // TODO: Utilize DeserializeSeed so proper normalization happens inside
-        // read_all ratheer than having to renormalize afterward
+        // TODO: Utilize DeserializeSeed (or equivalent) so proper normalization
+        // happens inside read_all, rather than having to renormalize afterward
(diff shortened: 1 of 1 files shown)


data/src/history.rs
```diff
@@ -168,15 +168,47 @@ pub struct Loaded {
...
+
+pub async fn load(kind: Kind, seed: Option<Seed>) -> Result<Loaded, Error> {
     let path = path(&kind).await?;
 
-    let messages = read_all(&path).await.unwrap_or_default();
+    let mut messages = read_all(&path).await.unwrap_or_default();
+
+    if let Some(seed) = seed {
+        // TODO: Utilize DeserializeSeed so proper normalization happens inside
+        // read_all ratheer than having to renormalize afterward
```
casperstorm:
```suggestion
        // read_all rather than having to renormalize afterward
```

casperstorm (approved):
All good, just a tiny nit. Tested and all runs as expected.

data/src/history.rs
andymandias: Thanks for catching that!

PR #1262, opened 2025-10-24
Prefix-Free User Deserialization

andymandias:
Adds a space between the access levels and names in `User` serialization in order to remove any need for knowing the server's PREFIX ISUPPORT parameter.  We can use space as the delimiter because it is not a valid access level, and I believe it is also forbidden in the nicknames, usernames, and hostnames.  If a space is not present in the serialization (i.e. it was serialized with an older version of Halloy), then deserialization falls back to parsing with the set of all prefixes (as is currently done).  Adds a new test JSON for future serde testing.

Also removes `Option` from the `CaseMap` argument in `User::parse` since `None` was not treated any different from `CaseMap::default()`, because I feel being a bit more explicit is preferable.

### DIFF
--- a/data/Cargo.toml
+++ b/data/Cargo.toml
@@ -17,7 +17,7 @@ required-features = ["message_tests"]
 [features]
 dev = []
 tor = ["irc/tor"]
-message_tests = []
+message_tests = ["iced/tokio"]
 
 [dependencies]
 thiserror = { workspace = true }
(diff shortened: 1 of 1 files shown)


data/Cargo.toml
```diff
@@ -44,7 +44,7 @@ flate2 = "1.0"
 hex = "0.4.3"
 iced_core = "0.14.0-dev"
 iced_wgpu = "0.14.0-dev"
-iced = { version = "0.14.0-dev", default-features = false, features = ["sysinfo"] }
+iced = { version = "0.14.0-dev", default-features = false, features = ["sysinfo", "tokio"] }
```
casperstorm: What is tokio used for?

PR #1272, opened 2025-10-28
Improve "Mark as Read on Message Sent" Behavior

andymandias:
Two minor improvements to how "mark as read on message sent" is handled:
- Show recently sent messages as read (show them above the backlog divider), while the read-marker may not yet be updated.
- When the server supports echoes, wait for echo from server to send MARKREAD (avoiding excess movement in read marker & reducing number of messages sent to the server).

### DIFF
--- a/src/buffer/scroll_view.rs
+++ b/src/buffer/scroll_view.rs
@@ -155,8 +155,8 @@ pub fn view<'a>(
     let Some(history::View {
         has_more_older_messages,
         has_more_newer_messages,
-        mut old_messages,
-        mut new_messages,
+        old_messages,
+        new_messages,
         max_nick_chars,
         max_prefix_chars,
         max_excess_timestamp_chars,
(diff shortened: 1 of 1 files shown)


casperstorm:
i wonder if we should consider

```toml
[buffer.backlog_separator]
hide_when_all_read = true
```

as default together with this?

andymandias:
> i wonder if we should consider
> 
> ```toml
> [buffer.backlog_separator]
> hide_when_all_read = true
> ```
> 
> as default together with this?

I don't have a strong preference one way or the other, but I think that would be in line with the current typical behavior in chat clients.  Will change.

casperstorm:
This works better than what we have now! However i see the backlog briefly appear, then disappear again:

https://github.com/user-attachments/assets/59691207-3ff4-4fb1-bc72-c8a997027fcc

andymandias:
Oh interesting, the round trip is longer than I expected.  ~~I'll increase the delay before showing reality.~~  I think my inclination that we don't update the read marker state without confirmation from the server is wrong, given how complicated it is to hide the jitter when doing so.  I've pulled in your solution instead.

Issue #1275, opened 2025-10-28
Performance regression on channel scroll


Erroneuz:
Halloy version: 2025.10
OS: Windows 11 25H2

Issue:
When scrolling back to former chat in a channel, the entire backlog is loaded at one point, which has a clear very negative impact on performance.

PR #1276, opened 2025-10-28
Scroll Performance Fix

andymandias:
Revert changes intended to fix issue when scrolling with multiple messages in history that have the exact same server time due to severe performance regression.  Will attempt another fix of that issue later.

Should address #1275.


casperstorm (approved):
Confirmed working

PR #1281, opened 2025-10-30
Improve Message Condensation

andymandias:
This PR is to improve message condensation by more thoroughly mapping nickname associations.  It also separates nickname association logic from the condensed message to condensed fragment conversion process to hopefully improve the legibility of the code.

Edit:  Also fixes a minor alignment error when using right-aligned nicknames.


casperstorm (approved):
More readable i like that:)
Tested and works.

PR #1284, opened 2025-10-30
Update to older version of Iosevka to reduce Halloy binary by 21%

casperstorm:
I've added a older version of iosevka to binary save size. We were previously using a 3 year old version of iosevka, but when i went to download the newest (33.3.3, as of writing this) i noticed that the size had grown to an absurd amount. I started digging into the difference because i don't believe we need anything special.

```
iosevka-term-regular.ttf from 2.3.3 (2019) is 410.7 kB.
iosevka-term-regular.tff from 33.3.3 (2025) is 9.9 MB.

```
From 2.3.3 to 33.3.3 it went from 3797 to 42809 glyphs. Size was increased by 2312% for JUST that font.
I've run a glyph test with a bunch of text, and i can't seem to find anything we are missing when running with version 2.3.3 (see below)

image

This PR shaves off roughly 21% of total Halloy binary size.
This PR also adds a light variant of iosevka


src/buffer/scroll_view.rs
```diff
@@ -211,7 +211,9 @@ pub fn view<'a>(
                 len,
                 usize::max(
                     max_excess_timestamp_chars.unwrap_or_default(),
-                    MESSAGE_MARKER_TEXT.chars().count(),
+                    4
+                    // TODO: get the width of the message marker text?
+                    // MESSAGE_MARKER_TEXT.chars().count(),
```
casperstorm: @andymandias @tarkah any suggestions what i can do here with this new change?

andymandias: I would be inclined to do a similar calculation as `font::width_from_chars`, but use the exact text/font used for the message marker.  Then assign it to a `static` and use here.

casperstorm: Agree, sounds like the right approach.

casperstorm: @andymandias take a look at latest commit and let me know what you think

src/buffer/scroll_view.rs
```diff
@@ -206,16 +206,14 @@ pub fn view<'a>(
...
+        let nick_width = font::width_from_chars(len, &config.font);
+        let timestamp_width = font::width_from_chars(
+            max_excess_timestamp_chars.unwrap_or_default(),
             &config.font,
-        )
+        );
+        let message_marker_width = font::width_of_message_marker(&config.font);
+
+        nick_width.max(timestamp_width).max(message_marker_width)
     });
```
andymandias, line 214: I'm not the original author, so I may be off base, but I think the original intention was to use max amongst the character counts to avoid having to call `width_from_chars` which I believe is a bit more heavyweight.  I think we can get away with it since it's a small number of calls, just wanted to note it.

While we're messing with this, it's probably past time to rename `max_nick_width`.  I think something like `right_alignment_width` is closer to its current purpose/value.  (I should've changed it when I was working on condensed messages.)

src/widget.rs
```diff
@@ -55,24 +58,25 @@ pub type Button<'a, Message> = iced::widget::Button<'a, Message, Theme>;
...
 pub fn message_marker<'a, M: 'a>(
     width: Option<f32>,
-    theme: &'a Theme,
+    config: &'a Config,
     style: impl Fn(&Theme) -> selectable_text::Style + 'a,
-    font_style: impl Fn(&Theme) -> Option<FontStyle>,
 ) -> Element<'a, M> {
-    let marker = selectable_text(MESSAGE_MARKER_TEXT);
+    let font_size =
+    config.font.size.map_or(TEXT_SIZE, f32::from) * 1.33;
```
andymandias: Should we make the scaling factor a `const` to ensure it stays the same in `font::width_of_message_marker`?

src/buffer/scroll_view.rs
casperstorm: True. Good point. Ive reduced the call to `width_from_char` now.

casperstorm:
> * Since we're shaving down the size on all our fonts, would it be worth spending those savings on more weights?  If we can't afford the full set, light-italic and semibold would cover the default pairings for light.

I've added lightitalic and semibold now. I think this is a good set of defaults. Do think we should add more?

andymandias:
> I've added lightitalic and semibold now. I think this is a good set of defaults. Do think we should add more?

I agree this is a good set of defaults.  I think thin will be the most common non-regular weight, and that'll support any bold/italic one might have in their theme.

PR #1294, opened 2025-11-03
Add border radius to images

kasperkronborg:
Border radius on both image previews and cards displaying images through meta tags


casperstorm (approved):
Perfect, thanks. Looks great.

PR #1296, opened 2025-11-04
Close all queries context menu

casperstorm:
This PR adds ability to close all active queries for a server. When doing so all query panes will also be closed.
Additionally i've sorted all menu items in context menu.

### DIFF
--- a/src/screen/dashboard.rs
+++ b/src/screen/dashboard.rs
@@ -380,9 +380,12 @@ impl Dashboard {
                 };
 
                 let (event_task, event) = match event {
-                    sidebar::Event::CloseAllQueries(server, queries) => {
-                        (self.leave_all_queries(clients, config, server, queries), None)
-                    }
+                    sidebar::Event::CloseAllQueries(server, queries) => (
+                        self.leave_all_queries(
+                            clients, config, server, queries,
+                        ),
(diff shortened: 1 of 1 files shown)


src/screen/dashboard.rs
```diff
@@ -2145,6 +2148,43 @@ impl Dashboard {
...
+        let history_tasks: Vec<_> = queries
+            .into_iter()
+            .filter_map(|query| {
+                self.history
+                    .close(history::Kind::Query(server.clone(), query), clients)
+                    .map(|task| Task::perform(task, Message::History))
+            })
+            .collect();
+
+        tasks.extend(history_tasks);
```
andymandias: It's not as efficient, but I wonder if it would be prudent to define tasks with something like:
```rust
        let tasks: Vec<Task<Message>> = queries
            .into_iter()
            .map(|query| {
                let buffer =
                    buffer::Upstream::Query(server.clone(), query.clone());

                self.leave_buffer(clients, config, buffer).0
            })
            .collect();
```
so that any changes to `leave_buffer` will be at least somewhat automatically reflected in `leave_all_queries`.

casperstorm: Awesome. This is used.

PR #1297, opened 2025-11-05
Not Sent Messages

andymandias:
This PR utilizes server echoes to confirm whether a server receives a sent message.  When connected to a server that supports echoes:
- Messages that are not confirmed are dimmed to indicate they are pending.
- A marker is added to unconfirmed messages after a timeout (currently 10s) to indicate the message has likely failed to send to the server.
- Clicking on the marker brings up a context menu, allowing the user to either delete or re-send the message (if they are connected to the server currently).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Added:
 - Exit application from user menu
 - Border radius on both image previews and cards displaying images through meta tags (configurable)
 - Context menu item added to server-wide buffers to close all queries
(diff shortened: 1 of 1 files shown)


casperstorm (approved):
Tested and this works good. I think it turned out great. I dont have much feedback other than it plugs nicely into what we have already. Its easy to read and i like the way you have structured the resend message fns.

This deserves a CHANGELOG entry.

PR #1300, opened 2025-11-06
Support unicode nicknames

mistivia:
While the original IRC specifications (RFC 2812) do not allow non-ASCII nicknames, some IRC servers (i.e. UnrealIRCd, Ergo, etc.) adopted optional unicode nick support for many years, making it a de facto standard in many communities.

Currently, Halloy does not fully support these unicode nicknames, likely due to restrictions in the existing parser that expects only ASCII characters for nicknames. This prevents users from connecting with Unicode nicknames, and also prevent users from receiving messages from other users with unicode nicknames.

Most modern IRC clients already support this feature, including:

- irssi
- WeeChat
- HexChat
- IRCCloud
- The Lounge
- kvIRC
- ...

With this PR, halloy would support unicode nicknames. But this PR is not thoroughly tested, and I'm not familiar with Halloy codebase, so there can be unexpected bugs. This is just a proof of concept.

### DIFF
--- a/irc/proto/src/parse.rs
+++ b/irc/proto/src/parse.rs
@@ -149,7 +149,7 @@ fn user(input: &str) -> IResult<&str, User> {
     let special = |input| one_of("-[]\\`_^{|}*/@")(input);
     // *( <letter> | <number> | <special> )
     let strict_nick = recognize(many1_count(alt((
-        satisfy(|c| c.is_ascii_alphanumeric()),
+        satisfy(|c| c.is_ascii_alphanumeric() || !c.is_ascii()),
         special,
     ))));
     // Used by things like matrix bridge
@@ -159,7 +159,7 @@ fn user(input: &str) -> IResult<&str, User> {
     let expanded_nick = verify(
         recognize(terminated(
             many1_count(alt((
-                satisfy(|c| c.is_ascii_alphanumeric()),
+                satisfy(|c| c.is_ascii_alphanumeric() || !c.is_ascii()),
                 special,
                 one_of(":."),
             ))),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,8 @@ Fixed:
 Changed:
 
 - By default the backlog separator is hidden when all messages in a pane have been marked as read (`buffer.backlog_separator.hide_when_all_read`)
-- Buffer context menu items in sidebar are sorted alphabetically 
+- Buffer context menu items in sidebar are sorted alphabetically
+- Unicode nicknames are now supported
 
 Thanks:
 
--- a/data/src/ctcp.rs
+++ b/data/src/ctcp.rs
@@ -66,7 +66,7 @@ pub fn parse_query(text: &str) -> Option<Query<'_>> {
         .strip_prefix('\u{1}')?;
 
     let (command, params) = if let Some((command, params)) =
-        query.split_once(char::is_whitespace)
+        query.split_once(|c| char::is_ascii_whitespace(&c))
     {
         (command.to_uppercase(), Some(params))
     } else {


irc/proto/src/parse.rs
```diff
@@ -149,7 +149,7 @@ fn user(input: &str) -> IResult<&str, User> {
     let special = |input| one_of("-[]\\`_^{|}*/@")(input);
     // *( <letter> | <number> | <special> )
     let strict_nick = recognize(many1_count(alt((
-        satisfy(|c| c.is_ascii_alphanumeric()),
+        satisfy(|c| c.is_ascii_alphanumeric() || c as u32 >= 128),
```
andymandias: Casting as a `u32` and testing it numerically is a bit open-ended.  I think it would need a fair amount of consideration and extensive testing.  I wonder if changing `c.is_ascii_alphanumeric` to `c.is_alphanumeric` would get us most of the way there, while keeping the scope still fairly well defined.  Would that work for your use case?

src/widget/selectable_text.rs
```diff
@@ -43,9 +43,6 @@ where
         Text {
             fragment: fragment.into_fragment(),
             format: Format {
-                #[cfg(debug_assertions)]
-                shaping: Shaping::Basic,
-                #[cfg(not(debug_assertions))]
                 shaping: Shaping::Advanced,
```
andymandias, line 49: The basic shaping setting being removed here should only affect debug builds.  I.e. building/running with the `--release` flag will get you advanced shaping.  I think for the purposes of this PR we can run with the release profile, so we can leave this as it was.

src/font.rs
```diff
@@ -123,7 +123,7 @@ pub fn width_from_chars(len: usize, config: &config::Font) -> f32 {
         font: MONO.clone().into(),
         align_x: text::Alignment::Right,
         align_y: alignment::Vertical::Top,
-        shaping: text::Shaping::Basic,
+        shaping: text::Shaping::Advanced,
```
andymandias: I'm curious what was the motivation for this change.  The characters being rendered here won't be Unicode (or visible), so I wouldn't expect advanced shaping to be useful in this context.  Is there specific behavior this change intends to correct?

ghost:
I have removed unnecessary modifications. It looks like only two lines of changes are needed to make everything work.

Here are some test cases:

- 中文名
- にほんじん
- 한국인
- Ἀλέξανδρος 
- Володи́мир

You can test this on ergo chat's testnet: [https://testnet.ergo.chat/](https://testnet.ergo.chat/) This server supports Unicode names, and Cyrillic, Greek, Chinese, Japanese, and Korean nicknames are all supported. It is already working for me:

image

I was unable to find a public IRC server that supports emoji nicknames, but on my private server, modified Halloy can properly display emoji nicknames, so do other clients like HexChat, kvIRC, etc.

To determine valid characters, I chose `c.is_ascii_alphanumeric() || !c.is_ascii()`.

Directly using `c.is_alphanumeric()` is feasible and works in most cases, but emoji characters are not supported, like:

- 🐶Woof🐶

If you agree, I hope to support emojis and leave the choice of valid nickname characters to the server. However, if you prefer to chooese a safer way and narrow the supported character range, then `c.is_alphanumeric()` is also acceptable. The decision is yours.

casperstorm:
Seems reasonable to me.

> Nicknames are non-empty strings with the following restrictions:
> 
> * They MUST NOT contain any of the following characters: space (' ', 0x20), comma (',', 0x2C), asterisk ('*', 0x2A), question mark ('?', 0x3F), exclamation mark ('!', 0x21), at sign ('@', 0x40).
> * They MUST NOT start with any of the following characters: dollar ('$', 0x24), colon (':', 0x3A).
> * They MUST NOT start with a character listed as a [channel type](https://modern.ircdocs.horse#channel-types), [channel membership prefix](https://modern.ircdocs.horse#channel-membership-prefixes), or prefix listed in the IRCv3 [multi-prefix Extension](https://ircv3.net/specs/extensions/multi-prefix).
> * They SHOULD NOT contain any dot character ('.', 0x2E).
> Servers MAY have additional implementation-specific nickname restrictions and SHOULD avoid the use of nicknames which are ambiguous with commands or command parameters where this could lead to confusion or error.

- https://modern.ircdocs.horse/#clients

At the end of the day, the server has the last say and we will get a `ERR_ERRONEUSNICKNAME` if anything goes wrong.

casperstorm (approved):

andymandias (changes requested):
Looks good.  The only thing I think we need to be careful about is splitting on Unicode whitespace.  That has been done in the past in Halloy, but it we got rid of most instances of it.  After review, it looks like the only remaining place that will need updating is in `data/src/ctcp.rs`, where
```rust
    let (command, params) = if let Some((command, params)) =
        query.split_once(char::is_whitespace)
    {
```
should be something like
```rust
    let (command, params) = if let Some((command, params)) =
        query.split_once(|c| char::is_ascii_whitespace(&c))
    {
```
With that change, and an entry in the CHANGELOG, I think this is good to go.

PR #1305, opened 2025-11-08
fuzzy matching in command bar

### LINKED ISSUE #942: [Feature Request]
If I have a server and a channel with the same name (lets say a server called `test` and a channel called `#test`) in the `Cmd+K` menu it will always prioritize the server over the channel. This is clearly not ideal behaviour in my opinion.

casperstorm:
Fixes #942 

This introduces fuzzy matching in command bar, and addresses a issue where servers would appear before channels if they had the same name. I've also renamed a few commands, and moved others.

Below is a example where i am on a server called "quakenet" and have joined channel "#quakenet"

image

### DIFF
--- a/src/screen/dashboard/command_bar.rs
+++ b/src/screen/dashboard/command_bar.rs
@@ -329,7 +329,7 @@ impl std::fmt::Display for Buffer {
                 if *maximize {
                     "Maximize buffer"
                 } else {
-                    "Restore buffer"
+                    "Restore buffer size"
                 }
             ),
             Buffer::New => write!(f, "New buffer"),


src/screen/dashboard/command_bar.rs
```diff
@@ -355,33 +323,37 @@ impl std::fmt::Display for Version {
...
-                    }
-                )
-            }
+            Buffer::Maximize(maximize) => write!(
+                f,
+                "{}",
+                if *maximize {
+                    "Maximize buffer"
+                } else {
+                    "Restore buffer"
```
andymandias: I think "Restore buffer size" would be a bit clearer, albeit a bit wordy.  Or, at least, I wasn't sure what the command would do until I understood the full context.

andymandias (approved):
Looks good, works well!

src/screen/dashboard/command_bar.rs
casperstorm: Agree!

PR #1309, opened 2025-11-11
ability to hide the indicator when a new version of Halloy is available 

casperstorm:
Requested by Ilya on IRC.

With this feature you won't see a indicator dot when a new version is available. This can be useful if you have installed Halloy through a package manager.

```toml
[sidebar.user_menu]
show_new_version_indicator = false
```


andymandias (approved):
Looks good to me!  Would it make sense to either:
- include this setting in the sidebar settings, or
- have a user_menu settings section, with the version setting and show_user_menu in it?

casperstorm:
> Looks good to me! Would it make sense to either:

Good call. I moved it under `[sidebar.user_menu]`.

PR #1311, opened 2025-11-12
Notifications Settings

andymandias:
Adds:
- `sound` setting for highlighted matches (used instead of the generic highlight notification sound, if specified)
- `[[notifications.channel."channel_name"]]` notifications, which will produce a notification for every message in a channel
- new syntax for `include`/`exclude` settings to allow for more complex logic (the old syntax is still supported)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ Added:
 - Fuzzy matching when searching in command bar
 - Setting to disable file transfers (DCC), removing it from the UI
 - Ability to hide nickname for consecutive messages from the same user
+- Ability to specify a distinct sound for each match highlight set
+- Setting to notify for every message in a channel
 
 Fixed:
 
@@ -32,7 +34,7 @@ Thanks:
 
 - Contributions: @kasperkronborg, @mistivia
 - Bug reports: @Frikilinux, ThePendulum, @crabbedhaloablution, @findus, @Darksecond
-- Feature requests: @cyrneko, ilya, abby
+- Feature requests: @cyrneko, ilya, abby, @clorteau, @RobertCraigie
 
 # 2025.11 (2025-10-27)
 


book/src/configuration/notifications/README.md
```diff
@@ -116,45 +127,43 @@ delay = 250
...
 
 [notifications.highlight]
-exclude = ["HalloyUser1", "#halloy"]
+exclude = { users = ["HalloyUser1", "#halloy"] }
 ```
 
 ### include
 
-Include notifications for nicks (and/or channels in `highlight`'s case).
+[Inclusion conditions](/configuration/conditions.md) in which you will be
```
casperstorm, line 151: When running the book locally these links doesn't work.

book/src/configuration/conditions.md
```diff
@@ -0,0 +1,45 @@
+# Inclusion/Exclusion Conditions
```
casperstorm, line 1: I wonder if we should put this into the guide section? 🤔

data/src/history/manager.rs
```diff
@@ -303,22 +308,17 @@ impl Manager {
         self.data.input.store_text(raw_input);
     }
 
+    // The message's blocked state should be determined prior to using this
+    // function.  In most cases, the best way to do that is by using the
```
casperstorm:
```suggestion
    // The message's blocked state should be determined prior to using this
    // function. In most cases, the best way to do that is by using the
```

Nit: extra space

data/src/message.rs
```diff
@@ -333,6 +346,102 @@ impl Message {
         })
     }
 
+    pub fn received_with_highlight<'a>(
```
casperstorm, line 351: Hmm curious about duplication of received here. I would expect a single fn, and then perhaps smaller fns to move some of the complex functionality away - i am afraid of having two so similar fns.

data/src/message.rs
```diff
@@ -2412,27 +2585,47 @@ fn content<'a>(
...
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct Highlight {
+    pub kind: HighlightKind,
+    pub channel: target::Channel,
+    pub user: User,
+    pub message: Message,
+}
```
casperstorm: We have src/message/* where we could do a message/highlight.rs file and then rename the `HighlightKind` into just `Kind` perhaps. Its a big file already. Just a thought, no required.

data/src/message.rs
andymandias: The reason I tacked this on to `received` is because we need to know exactly which highlight matched in order to pick the sound.  The only ways to do that with certainty I could see were (a) store that data with the message, which is a bunch of extra metadata we don't want or most of the time need, or (b) re-parse the message for highlights.  So I went with (b), but to avoid having to re-parse the highlights I made it returned by the original parsing.  We could potentially have `received` always return the highlight info and simply throw that info away in every place we don't need it.

book/src/configuration/notifications/README.md
andymandias: 🤦 will look into why lychee doesn't catch this

data/src/message.rs
andymandias: Ahh, like `data/src/message/broadcast.rs`?  I like this idea.

book/src/configuration/conditions.md
andymandias: Looks like we do need to link the file in `SUMMARY.md`, otherwise mdbook won't render it to html (even if it's linked from the body of another document).  I put it in guides as suggested, but it could potentially work under the config section as well (there are just so many subsections in the config section already).

PR #1322, opened 2025-11-19
apostrophe added to regex

### LINKED ISSUE #1317: Don't consider `word'<nick>` as a highlight
if someone has a nickname like `t`, words like `don't` and `can't` have the `t` at the end highlighted/clickable

example: 

Image

casperstorm:
New regex enforces that the character before the match must NOT be a letter, digit, underscore, or **apostrophe**.

image

Fixes #1317


casperstorm:
See https://github.com/squidowl/halloy/issues/1317#issuecomment-3557245568.

PR #1327, opened 2025-11-21
Expandable Condensed Messages

andymandias:
Allows condensed messages to be expanded/contracted by left-click.

Also adds `change-host` and `kick` as a supported server message types for condensation.

image


casperstorm (approved):
Tested, and confirmed all working. Great PR.

PR #1329, opened 2025-11-22
Add settings for backlog rule appearance

ignaloidas:
~~Stuff left to implement: adding `backlog_rule` to the theme editor - not sure how to do it properly because the old field from which I should pull the default is in a different struct from the new field.~~ Done 

Also, not sure on the naming of `text` option, or whether it should be turned into a boolean on whether to show the text or not - I do prefer it without text if the color is different, not entirely certain if the text truly needs to be customizable.

### DIFF
--- a/src/screen/dashboard/theme_editor.rs
+++ b/src/screen/dashboard/theme_editor.rs
@@ -459,9 +459,7 @@ impl Component {
             Component::Text(text) => text.color(styles),
             Component::Buffer(buffer) => buffer.color(styles),
             Component::Buttons(buttons) => Some(buttons.color(styles)),
-            Component::Formatting(formatting) => {
-                formatting.color(styles)
-            }
+            Component::Formatting(formatting) => formatting.color(styles),
         }
     }
 


data/src/buffer.rs
```diff
@@ -251,12 +251,14 @@ impl Default for Timestamp {
 #[serde(default)]
 pub struct BacklogSeparator {
     pub hide_when_all_read: bool,
+    pub text: String,
```
casperstorm: Lets do a enum here instead.

```rust
enum BacklogText { 
  Text(String) 
  Hidden
}
```

then you impl deserializer that handles boolean (`false`) => Hidden, and a string => Text(String).
(and perhaps boolean `true` => Text("backlog"))

src/buffer/scroll_view.rs
```diff
@@ -408,24 +408,41 @@ pub fn view<'a>(
...
-        ]
-        .padding(2)
-        .align_y(iced::Alignment::Center)
-    } else {
-        row![]
-    };
+    let divider =
+        if show_backlog_divier {
+            if config.buffer.backlog_separator.text.is_empty() {
+                row![
```
casperstorm: With it as a enum we could use a match here

```rust
let row = config.buffer.backlog_separator.text {
    Some(backlog) => row![ rule.width(fill), text(backlog), rule.width(fill) ],
    None => row![ rule.width(fill) ],
};
```

or combine everything into a single `row![]` since you can pass optionals to it.
so you could pass: rule, backlog_text.map(...), backlog_text.map(rule).

data/src/buffer.rs
```diff
@@ -247,16 +247,52 @@ impl Default for Timestamp {
...
 pub struct BacklogSeparator {
     pub hide_when_all_read: bool,
+    pub text: BacklogText,
 }
 
 impl Default for BacklogSeparator {
     fn default() -> Self {
         Self {
             hide_when_all_read: true,
+            text: BacklogText::Text("backlog".to_string()),
```
casperstorm: Impl Default for BacklogText like:

```rust
impl Default for BacklogText {
    fn default() -> Self {
        Self::Text(String::from("backlog"))
    }
}
```

And then just `BacklogText::default()` here.

data/src/buffer.rs
```diff
@@ -247,16 +247,58 @@ impl Default for Timestamp {
...
+        }
+
+        match T::deserialize(deserializer)? {
+            T::String(conf) => Ok(Self::Text(conf)),
+            T::Bool(val) => {
+                if val == false {
+                    Ok(Self::Hidden)
+                } else {
+                    Err(de::Error::custom(
+                        "true is not a valid value for buffer.backlog_separator.text",
```
casperstorm: I feel like we could support true and turn it into default value just to support both boolean values. What do you think? Feels better to always support both false AND true.

casperstorm: I feel like we could support `true` and turn it into default value just to support both boolean values. What do you think? Feels better to always support both `false` AND `true`.

src/buffer/scroll_view.rs
```diff
@@ -418,24 +408,40 @@ pub fn view<'a>(
...
+                    container(rule::horizontal(1).style(theme::rule::backlog))
+                        .width(Length::Fill)
+                        .padding(padding::right(6)),
+                    text(separator_text)
+                        .size(divider_font_size)
+                        .style(theme::text::secondary)
+                        .color(
+                            theme.styles().buffer.backlog_rule.unwrap_or(
+                                theme.styles().general.horizontal_rule
+                            )
```
casperstorm: This is not right. Here you are giving the text the rule color. 

Do we want to be able to color the RULE, the TEXT or BOTH of a backlog event?
And what about the date-change event?

A backlog (and date-change) event consists of two elements, as you now know:
1. The rule (using horizontal rule color from the theme)
2. The text (using secondary text color)

After thinking a bit about this, i feel we need to be a little more clear in what this is actually solving?
I think it's a bit dangerous, looking at this here, to start playing with the colors - a bit smaller, and easier if we just keep it to the text for now.

casperstorm (changes requested):

src/buffer/scroll_view.rs
ignaloidas: From my quick testing it did look better with the text using the same color as the rule, so that's why the change is here, but I am going to drop it for now, because as you correctly point out there are some extra complexities with it.

data/src/buffer.rs
ignaloidas: For me it felt a bit weird to have `true` as just a way to specify "use the default", did change it to that though.

ignaloidas:
Rebased to fix merge conflicts - is there anything else needed from my side for this to land?

src/buffer/scroll_view.rs
```diff
@@ -433,24 +433,35 @@ pub fn view<'a>(
...
-        .align_y(iced::Alignment::Center)
-    } else {
-        row![]
-    };
+    let divider =
+        if show_backlog_divier {
+            match &config.buffer.backlog_separator.text {
+                data::buffer::BacklogText::Hidden => row![
+                    container(rule::horizontal(1).style(theme::rule::backlog))
+                        .width(Length::Fill)
```
casperstorm:
```suggestion
                    container(rule::horizontal(1).style(theme::rule::backlog)).padding([2, 0])
                        .width(Length::Fill)
```

Let's add a bit of vertical padding. Then i think its good.

ignaloidas:
Added a changelog entry, and a bit of padding, also cleaned up the commits a bit to not have 6 commits for PR feedback fixes.

casperstorm:
@ignaloidas can you let me edit? there are a few conflicts still.

ignaloidas:
I do have it set to allow edits by maintainers, so should work?

casperstorm:
Thank you for the contribution, @ignaloidas :)

PR #1357, opened 2025-12-09
channel discovery (list)

casperstorm:
This PR adds a channel discovery mode which opens either by user menu > channel discovery, or when you write `/list`.
From here you can select one of your servers and it will fetch all the channels (through LIST)

image

Thanks to @kasperkronborg for assistance.


andymandias (approved):
A great addition, much nicer than trying to work with the output of `/list` (and `ELIST` extensions).

casperstorm:
Thanks for the help with it.

PR #1358, opened 2025-12-09
Send messages using numpad enter

csmith:
Currently pressing numpad enter inserts a new line into the input field (which then fails to send properly if you hit normal enter).

I'm not sure if this is an ideal long-term fix (maybe iced should handle this? or it should be treated as a keybind?), but for now just translate NumpadEnter keycodes into Named::Enter presses.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,9 +12,11 @@ Fixed:
 - No longer any delay in showing a tooltip after hovering
 - Ignored users will not be suggested by user auto-completion
 - Channel auto-completion works with all channel prefixes specified by the server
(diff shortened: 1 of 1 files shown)


casperstorm:
@csmith, do you mind updating CHANGELOG? :)

csmith:
> @csmith, do you mind updating CHANGELOG? :)

Done. I (rather presumptuously) added myself to the thanks section too :joy:

casperstorm:
> > @csmith, do you mind updating CHANGELOG? :)
> 
> Done. I (rather presumptuously) added myself to the thanks section too 😂

image

Issue #1354, opened 2025-12-04
Ignore hide_consecutive after previews


Erroneuz:
Feature request.

hide_consecutive is really nifty, but it would be nice with an option, so it doesn't do it after a preview where an image is show in chat.

casperstorm:
Fixed in main, you can now do:

```toml
[buffer.nickname.hide_consecutive]
enabled = true
show_after_previews = true
```

PR #1361, opened 2025-12-12
option to not hide consecutive if previous message was a image preview

casperstorm:
Fixes https://github.com/squidowl/halloy/issues/1354

### DIFF
--- a/book/src/configuration/buffer/nickname/hide-consecutive.md
+++ b/book/src/configuration/buffer/nickname/hide-consecutive.md
@@ -1,0 +1,45 @@
+# Hide Consecutive
+
+Hide nickname if consecutive messages are from the same user.  
+
+- [Hide\_Consecutive](#hide-consecutive)
+  - [Configuration](#configuration)
+    - [enabled](#enabled)
+    - [show_after_previews](#show_after_previews)
+
+
+## Configuration
+
+> ⚠️ `hide_consecutive` does not work in conjunction with `alignment = "top"` .
+
+
+### enabled
+
+If specified as `{ smart = integer }` then the nickname will be hidden for consecutive messages
+are from the same user and each is within `smart` seconds of each other.
+
+```toml
+# Type: boolean
+# Values: true, false, or { smart = integer }
+# Default: false
+
+[buffer.nickname.hide_consecutive]
+enabled = true
+
+# hide if the previous message was from the same user and sent within 2m of the current message
+[buffer.nickname.hide_consecutive]
+enabled = { smart = 120 }
+```
+
+### show_after_previews
+
+Show nicknames after messages with visible image or link previews.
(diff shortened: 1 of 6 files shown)


book/src/configuration/buffer/nickname/hide-consecutive.md
```diff
@@ -0,0 +1,45 @@
...
+
+### show_after_previews
+
+Show nicknames after messages with visible image or link previews.
+Note: has no effect when `enabled = false`.
+
+```toml
+# Type: boolean
+# Values: true, false
+# Default: false
```
andymandias, line 42: The default in the documentation (here) is listed as `false`, but `default_show_nickname_after_preview` is set to `true`.  The default of `false` makes more sense to me, since `true` potentially overrides the behavior defined in `enabled` (which may be surprising for a user).  But I don't feel strongly about it.

data/src/config/buffer/nickname.rs
```diff
@@ -59,35 +59,90 @@ pub enum ShownStatus {
...
+                enabled,
+                show_after_previews,
+            }),
+            Inner::Boolean(enabled) => Ok(HideConsecutive {
+                enabled: if enabled {
+                    HideConsecutiveEnabled::Enabled(None)
+                } else {
+                    HideConsecutiveEnabled::Disabled
+                },
+                show_after_previews: true,
```
andymandias: I think we should use `default_show_nickname_after_preview` here (and for the `show_after_previews` set for `Inner::Smart` as well).

src/buffer/scroll_view.rs
```diff
@@ -240,21 +286,31 @@ pub fn view<'a>(
...
-                                        (message::Source::User(user), message::Source::User(prev_user)) if user == prev_user
-                                    ) && duration.is_none_or(|duration| message.server_time - prev_message.server_time < duration)
-                                })
+                // Check if the previous message has a visible image preview
+                let prev_has_visible_preview = if config.buffer.nickname.hide_consecutive.show_after_previews {
+                    prev_message
+                    .is_some_and(|prev_msg| has_visible_preview(prev_msg, state, previews, &visible_for_source))
                 } else {
                     false
                 };
```
andymandias: Since we set `prev_has_visible_preview` to `false` based on the configuration option I feel like the name is slightly misleading.  But I'm not coming up with a better name (`show_after_prev_message_preview`?).  To avoid having to come up with a good name I'd probably put the logic inside the `prev_message.is_some_and` below :sweat_smile:.

book/src/configuration/buffer/nickname/hide-consecutive.md
casperstorm: Agree 👍🏻

PR #1365, opened 2025-12-12
control padding and spacing in sidebar and context menu

### LINKED ISSUE #1363: Feature request: Allow configuration of vertical space between sidebar entries
It could be nice to be able to configure the vertical space between the entries in the side bars (mostly channels in my case), possibly to make them more compact (e.g. similar to the nickname list).

Image

casperstorm:
Fixes #1363 

I've added settings to control both button padding and server spacing in sidebar:

```toml
[sidebar.padding]
buffer = [2, 5]

[sidebar.spacing]
server = 0
```

screenshot-2025-12-13_22-38-56

I've also added a new config setting for context menus:

```toml
[context_menu.padding]
entry = [2, 5]
```

screenshot-2025-12-13_22-38-11

### DIFF
--- a/book/src/configuration/context-menu/padding.md
+++ b/book/src/configuration/context-menu/padding.md
@@ -10,7 +10,7 @@ Adjust padding for context menu
 
 ### entry
 
-Controls padding for entries in a context menu
+Controls the padding around each entry in context menus.
 The value is an array where the first value is vertical padding and the second is horizontal padding. 
 
 ```toml
--- a/book/src/configuration/sidebar/spacing.md
+++ b/book/src/configuration/sidebar/spacing.md
@@ -11,7 +11,7 @@ Adjust spacing for sidebar
 
 ### server
 
-Controls the vertical spacing between servers
+Controls the vertical spacing between servers (i.e. between the last buffer for one server and the server buffer for the next server).
 
 ```toml
 # Type: number


lorenzosu:
@casperstorm thanks, finally managed to test this and it works very well IMHO, with the added benefit that users can really personalise this as they prefer (in my case reducing the padding, but some might like to have it wider :-) ) - The menu padding setting is also a nice addition.
Maybe I'm missing something, but I'm not 100% sure how the `[sidebar.spacing]` setting should work though? I tried using your example setting and changing numbers for the `server` parameter (including somewhat more 'extreme' values) but I can't seem to see any notable effect besides trying a negative number and getting a nice warning that the value should be u32 :-)

casperstorm:
> Maybe I'm missing something, but I'm not 100% sure how the `[sidebar.spacing]` setting should work though? I tried using your example setting and changing numbers for the `server` parameter (including somewhat more 'extreme' values) but I can't seem to see any notable effect besides trying a negative number and getting a nice warning that the value should be u32 :-)

image

Thanks for the feedback! The `[sidebar.spacing.server]` is the vertical space between each server section!

lorenzosu:
[...]
> Thanks for the feedback! The `[sidebar.spacing.server]` is the vertical space between each server section!

~~I thought so.. But can't seem to reproduce it. I compiled (following the installation instructions for source) on this branch. I'm on Linux btw, not sure if it's relevant~~

OK I don't seem to have a 'server' maybe because I have only one server configuered :-D

casperstorm:
> OK I don't seem to have a 'server' maybe because I have only one server configuered :-D

A sever = network, so if you are only on a single network it doesn't give you anything.

lorenzosu:
> > OK I don't seem to have a 'server' maybe because I have only one server configuered :-D
> 
> A sever = network, so if you are only on a single network it doesn't give you anything.

Gotcha! That makes totally sense, I was for some reason thinking network == server which is not the case... sorry for the noise

casperstorm:
> > > OK I don't seem to have a 'server' maybe because I have only one server configuered :-D
> > 
> > 
> > A sever = network, so if you are only on a single network it doesn't give you anything.
> 
> Gotcha! That makes totally sense, I was for some reason thinking network == server which is not the case... sorry for the noise

All good, and no need to say sorry! You were the reason i added this 😄

book/src/configuration/context-menu/padding.md
```diff
@@ -0,0 +1,22 @@
...
+- [Padding](#padding)
+  - [Configuration](#configuration)
+    - [entry](#entry)
+
+## Configuration
+
+### entry
+
+Controls padding for entries in a context menu
+The value is an array where the first value is vertical padding and the second is horizontal padding.
```
andymandias: How about "Controls the padding around each entry in context menus." as the description for this one?

Would it make sense to name the components (i.e. `horizontal` and `vertical`) rather than an array?  Might be easier for users to understand.

book/src/configuration/sidebar/spacing.md
```diff
@@ -0,0 +1,23 @@
...
+  - [Configuration](#configuration)
+    - [server](#server)
+
+
+## Configuration
+
+### server
+
+Controls the vertical spacing between servers
+
```
andymandias, line 15: Maybe "Controls the vertical spacing between servers (i.e. between the last buffer for one server and the server buffer for the next server)." to give a second description here?  It's a little tricky to understand without a picture, so I think an extra explanation might help.

PR #1371, opened 2025-12-19
Update Iced

andymandias:
Updates Iced, with two changes (at least) that warrant review:
- Auto-scroll functionality has been added to `scrollable`.  I believe this should operate similar to middle-click on Firefox that places an icon then the page's scroll movement is controlled by the cursor's current position relative to the icon.  I extrapolated styling options from the scroller's styling, but I can't get the auto-scroll functionality to work for more than a split-second to verify that the styling is reasonable.  (The auto-scroll functionality only shows for a split-second for me on the unpatched version of iced as well, so it may just be too early in its development to use.)
- [`hint_factor`](https://docs.iced.rs/iced/advanced/struct.Text.html#structfield.hint_factor) has been added to text.  For displayed text fields I've just taken the renderer's `scale_factor()` to use as the `hint_factor`, but I'm not certain that's right.  (Not-displayed text fields are the procedures in `font.rs` that measure the width of characters.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1271,8 +1271,7 @@ dependencies = [
 [[package]]
 name = "cosmic-text"
 version = "0.15.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "173852283a9a57a3cbe365d86e74dc428a09c50421477d5ad6fe9d9509e37737"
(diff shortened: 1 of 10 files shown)


casperstorm (approved):
LGTM. Curious if @hecrj can fill in on above?

hecrj:
I recommend you to update to `0.14` first.

The `hinting` changes have just landed in `master` and are still shaping up. Your fork is missing a couple of important fixes.

casperstorm:
LGTM

PR #1386, opened 2026-01-02
Prevent Search Query in Channel Discovery Flashing

andymandias:
Prevents the text_input in Channel Discovery from flashing when the pane is grabbed.
Also removes a redundant specification in primary theme for text_input.


casperstorm (approved):
Good solution

PR #1387, opened 2026-01-02
ability to hide the server icon

casperstorm:
Ability to hide server icon with:

```toml
[sidebar]
server_icon = "hidden"
```

or adjust the size like before:

```toml
[sidebar]
server_icon = 24
```

Another small change is that if we are not connected to a server we also color the title:

image

This serves two purposes:
1. I think it look better
2. If you hide server icon you can still see if server is not connected

### DIFF
--- a/book/src/configuration/sidebar/README.md
+++ b/book/src/configuration/sidebar/README.md
@@ -20,8 +20,6 @@ Sidebar settings for Halloy.
 
 Configure the server icon display.
 
-Note: If size is set larger than the line height of the specified [font](../font/) then the icon will not render.
-
 ```toml
 # Type: integer or string
 # Values: any positive integer or "hidden"
--- a/data/src/config/sidebar.rs
+++ b/data/src/config/sidebar.rs
@@ -61,7 +61,7 @@ where
             "hidden" | "none" => Ok(ServerIcon::Hidden),
             _ => Err(serde::de::Error::invalid_value(
                 serde::de::Unexpected::Str(string),
-                &"\"hidden\" or a size object",
+                &"\"hidden\" or a size (positive integer)",
             )),
         })
         .bool(|value| match value {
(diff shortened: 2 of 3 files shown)


book/src/configuration/sidebar/README.md
```diff
@@ -16,19 +16,26 @@ Sidebar settings for Halloy.
...
 ## Configuration
 
-### server_icon_size
+### server_icon
 
-Adjust server icon size.
+Configure the server icon display.
 
-Note: If set larger than the line height of the specified [font](../font/) then the icon will not render.
+Note: If size is set larger than the line height of the specified [font](../font/) then the icon will not render.
```
andymandias: We removed this limitation at some point in the past, so we can get rid of this note.

data/src/config/sidebar.rs
```diff
@@ -16,13 +16,62 @@ pub struct Sidebar {
...
+                    serde::de::Unexpected::Unsigned(value as u64),
+                    &"a positive integer",
+                ))
+            }
+        })
+        .string(|string| match string {
+            "hidden" | "none" => Ok(ServerIcon::Hidden),
+            _ => Err(serde::de::Error::invalid_value(
+                serde::de::Unexpected::Str(string),
+                &"\"hidden\" or a size object",
```
andymandias: How about `"\"hidden\" or a size (positive integer)"`, to be explicit?

src/screen/dashboard/sidebar.rs
```diff
@@ -811,33 +811,42 @@ fn upstream_buffer_button<'a>(
...
+                    } else {
+                        theme::text::primary
+                    }
+                } else {
+                    theme::text::error
+                })
+                .size(size),
+                size,
+            )),
+            data::config::sidebar::ServerIcon::Hidden => None,
```
andymandias: I think we should show the unread/highlight icon (as appropriate) in this case, as if it were any other upstream buffer.  Otherwise, unread status will not be indicated.

casperstorm: thats a great feedback, take a look at newest commit and let me know.

PR #1391, opened 2026-01-05
Handle messages that are terminated with '\n' instead of '\r\n'

424ever:
The IRC Server of [osu!](https://osu.ppy.sh/wiki/en/Community/Internet_Relay_Chat) sends QUIT messages that are terminated with `\n` instead of `\r\n`, which causes parsing to fail.

### DIFF
--- a/irc/src/codec.rs
+++ b/irc/src/codec.rs
@@ -16,19 +16,11 @@ impl Decoder for Codec {
         &mut self,
         src: &mut BytesMut,
     ) -> Result<Option<Self::Item>, Self::Error> {
-        let Some(pos) = src.windows(2).enumerate().find_map(|(i, b)| {
-            if b == [b'\r', b'\n'] {
-                Some(i + 2)
-            } else if b[0] == b'\n' {
-                Some(i + 1)
-            } else {
-                None
-            }
-        }) else {
+        let Some(pos) = src.iter().position(|&b| b == b'\n') else {
             return Ok(None);
         };
 
-        let bytes = Vec::from(src.split_to(pos));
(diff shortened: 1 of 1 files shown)


424ever:
I think the splitting in codec.rs doesn't quite work how I wrote it, I'll change it later today

tarkah:
Irc docs specifically state use of cr/lf, not lf on it's own.

https://modern.ircdocs.horse/#message-format

I'm afraid of the unintended consequences this change would cause. Why does this server not use carriage return?

424ever:
> I'm afraid of the unintended consequences this change would cause.

I understand that this doesn't follow the specification, but since this pattern being supported by *most* irc implementations I've checked, I suspect it wouldn't cause any issues.
 - [node-irc](https://github.com/martynsmith/node-irc/pull/280)
 - [the irc-proto crate](https://github.com/aatxe/irc/commit/321bbec7ad3d88c1a9ec23d2786780bd780dd67b)
 - [HexChat](https://github.com/hexchat/hexchat/blob/master/src/common/server.c#L352) (this one looks like it might be accidental, to be fair)

As seen in the node-irc pr, [rfc1459](https://www.rfc-editor.org/rfc/rfc1459.html#section-8) also recommends splitting messages on different line endings instead of requiring CR-LF.

> Why does this server not use carriage return?

This server is primarily used as the in-game chat, and the lead developer has previously said that adhering to the specification and supporting specific clients is not a priority

424ever:
I guess I'll just keep this in a fork to use it, since it *is* non-standard

Issue #374, opened 2024-05-08
request: /connect or /server


vidje:
From what I can tell, so far the only way to connect to a new server is via config.
Would it be possible to add /connect or /server type of command to open a new server buffer from within the client, without needing to add/keep servers in the config?

auronandace:
Taking the example of weechat it has a separate main buffer where one would issue such commands. Would something similar be needed in halloy or could it be done through the command bar?

casperstorm:
An alternative is doing single pane mode: https://halloy.chat/guides/single-pane.html

PR #1396, opened 2026-01-09
Connect Command

andymandias:
Adds a `/connect` command to temporarily connect to servers (using the mechanism for opening irc/ircs URLs).

Partially addresses issue #374.


casperstorm (approved):
This looks good. Should we in the commands.md mention that it is only temp. and to persist a connection you have to edit the config file, or just by pointing to a configuration page?

andymandias:
> This looks good. Should we in the commands.md mention that it is only temp. and to persist a connection you have to edit the config file, or just by pointing to a configuration page?

I think this is a good idea, I can add the warning.  Maybe we should put a similar warning for join as well?

Issue #1369, opened 2025-12-18
Gaps in menu icon rendering differ in size


Alexendoo:
The gap between the first two bars is larger than the gap between the second & third

<img width="30" height="35" alt="Image" src="https://github.com/user-attachments/assets/91b63880-7915-4dce-8b58-1869fb0a4ecc" />

Halloy 2015.12, Windows 11 10.0.26200 Build 26200

casperstorm:
This is a icon called `menu` from Font Awesome from https://fontello.com/.
We could consider another icon in there instead.

PR #1409, opened 2026-01-16
Adjust menu icon

casperstorm:
Fixes https://github.com/squidowl/halloy/issues/1369

Previous:
image

Now:
image


andymandias (approved):
Looks good to me!  The icon looks nice and even as well.  The slightly larger size is welcome change for me as well.

PR #1410, opened 2026-01-16
Fix GoToMessage

andymandias:
Fixes an issue where clicking on a message link in the highlights buffer would only successfully navigate to the message if the message's buffer is already open (otherwise the buffer would only open).


casperstorm (approved):
Great way to fix this. Good work.

PR #1414, opened 2026-01-20
Only-Emojis Font Size

andymandias:
Adds a setting for the font size of messages that contain only emojis.  The intention/expectation is that users can set a larger size for messages that contain only emojis.

Also included in the PR:
- Upgrade emojis to Unicode v17.0 (crate to 0.8.0).
- Fix in documentation for `bold_weight` font setting.


casperstorm (approved):
Fun setting, works great.

Issue #325, opened 2024-04-06
Unable to build with LTO enabled


txtsd:
When compiling with LTO enabled, the build process fails.

See log:
https://bpa.st/V4KQ

(Couldn't paste the log directly because it's longer than GitHub's maximum character limit)

k1gen:
+1

tarkah:
I cannot reproduce on Linux with `cargo build --release --config profile.release.lto=\"on\"`

k1gen:
`cargo build --release --config profile.release.lto=\"on\"` also works for me, but this doesn't: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=halloy

k1gen:
> Seems like an issue in the AUR packaging environment then. What linker is being used? Are other rust flags enabled behind the scenes? Which rust version?

```
$ cat .cargo/config.toml
[profile.release]
strip = true
lto = "thin"

[build]
rustc-wrapper = "sccache"
$ rustc --version
rustc 1.77.2 (25ef9e3d8 2024-04-09)
```
I have no idea what linker is being used, the only one I have installed is the `GNU ld`

txtsd:
There's a known bug that LTO does weird things if you
1. include some C in your rust code
2. compile with GCC (not clang)
3. use lto.

Not doing just one of those three appears to "fix" it

See these:
https://github.com/gyscos/zstd-rs/issues?q=is%3Aissue+undefined+reference+to

PR #1418, opened 2026-01-21
Release Profile

### LINKED ISSUE #361: Add `profile.release` to enable link time optimizations

andymandias:
Adds a packaging profile using in build/install scripts and GH release action.  Packaging profile adds ["fat" LTO](https://doc.rust-lang.org/cargo/reference/profiles.html#lto) and [codegen-units](https://doc.rust-lang.org/cargo/reference/profiles.html#codegen-units) similar to #367.

Significantly increases compile time, optimization benefits need testing, and may need testing with AUR/Nix (#325).

If accepted, will close #361.


andymandias:
Updated to move the `packaging` profile to `release` (recent experience suggests it could be overlooked as a non-standard profile).  Minor tweak to the `debug`/`dev` profile to provide a less-customized experience.

PR #1421, opened 2026-01-24
fix buffer not showing history until resize

### LINKED ISSUE #1353: buffer contents don't show up until view is resized
I often observe the following bug:

1. Start out with an empty halloy view (no buffer selected)
2. Click on a buffer (channel)
3. The buffer view contains no messages or only the most recent message.
4. When I change the size of the buffer window by resizing the whole halloy window or opening a second buffer, now the other messages appear.

Instead of appearing only on resize, those messages should appear immediately. The bug is somewhat random. Sometimes the messages show up correctly on first try.

I'm on Arch Linux on Wayland (Sway). Happens both with the official release binaries and building from source at most recent commit.

e00E:
Halloy reuses an existing buffer pane when creating a new one in some cases. In this case, the new pane gets zero size in Pane::new. A comment there says

> Will get set initially via `Message::Resized`

That comment is incorrect. When creating a new pane, a resize doesnt' happen automatically. This results in that pane staying at zero size.

I fix this by updating the size to the size of the replaced pane.

It might be a good idea to go through all references to `Pane::new` and make sure that the size is correct. Maybe move the size to the arguments so that you can't not set it. I didn't include that here because I'm unfamiliar with halloy internals. Feel free to do it in a follow up change.

fixes #1353

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 Fixed:
 
 - Version number reported in AppData file
+- Buffer not showing history until resize
 
 # 2026.1 (2026-01-20)
 


casperstorm:
Thank you for the fix, @e00E.

src/buffer/scroll_view.rs
```diff
@@ -1201,7 +1201,7 @@ impl Status {
 fn step_messages(height: f32, config: &Config) -> usize {
     let line_height = theme::line_height(&config.font);
 
-    (height / line_height) as usize
+    (height / line_height).min(8.0) as usize
```
e00E, line 1204: This unfortunately breaks something again. With this change, I have a similar bug to the one mentioned in this PR. When I open the channel buffer, it shows some messages but not enough, and I cannot scroll up. Without this change but including the other changes from this PR, everything is fine.

casperstorm: Right should have been `.max()` 😓

e00E: Everything looks good with the max() fix.

PR #1429, opened 2026-01-29
Fix Some Filter Effects Being Applied to All Servers

andymandias:
Fix for query panes (and direct message notifications) being hidden on all servers due to filters that were specified for a single server only.


casperstorm:
LGTM, Andy.

PR #1430, opened 2026-01-29
Introduce font.line_height

Hummer12007:
Introduce the font.line_height config value.

Wire it up for all text in the main interface. It probably doesn't matter that much for other views but idk.

Also two minor refactorings to make the commits nice and clean.

### DIFF
--- a/src/widget.rs
+++ b/src/widget.rs
@@ -76,7 +76,7 @@ where
 
     let def_line_height = 1.3;
     let cfg_line_height = config.font.line_height.unwrap_or(def_line_height);
-    let line_height_delta = cfg_line_height / def_line_height;
+    let line_height_ratio = cfg_line_height / def_line_height;
 
     let (text, font_size, line_height) = match marker {
         Marker::Dot => (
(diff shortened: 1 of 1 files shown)


Hummer12007:
Want to wire it up in a few more places.

book/src/configuration/font/README.md
```diff
@@ -47,6 +48,19 @@ Font size.
...
 ```
 
+### line_height
+
+Line height (relative to the font size).
+
+```toml
+# Type: number
+# Values: any positive float
+# Default: not set (resolves to LineHeight::default()) in iced, which is 1.3
```
casperstorm:
```suggestion
# Default: 1.3
```

book/src/configuration/font/README.md
```diff
@@ -47,6 +48,19 @@ Font size.
 size = 13
 ```
 
+### line_height
+
+Line height (relative to the font size).
```
casperstorm: We should note that a restart of the client is nessecary in order for this to take effect.

src/buffer/input_view.rs
```diff
@@ -408,13 +409,18 @@ pub fn view<'a>(
                 |title: Text<'a>,
                  keybind: Option<data::shortcut::KeyBind>,
                  message: Option<Message>| {
+                    let title =
+                        title.line_height(theme::line_height(&config.font));
                     button(
                         row![
                             title,
```
casperstorm: Let's move the `line_height(..)` to line 416. No need for the `let title...`

book/src/configuration/font/README.md
Hummer12007: There is a disclaimer on top, though. That's why I didn't mention it here

casperstorm (approved):

PR #1437, opened 2026-01-29
Transparent background support

Hummer12007:
⚠️  **DISCLAIMER**: When setting an RGBA background color, beware that iced may (or may not) use premultiplied alpha for RGBA. This means that you should manually multiply the colors by the alpha if they feel odd for correct rendering.

**Example**: To get #181818CC you should use #131313CC in your theme
(CC corresponds to alpha of 0.8, and 0x18 * 0.8 = 0x13)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Added:
 - Setting to adjust inner spacing and outer padding for panes (`pane.gap.inner`, `pane.gap.outer`)
 - Setting to style unread query buffers as highlights in the sidebar (`sidebar.unread_indicator.query_as_highlight`)
 - Settings to configure card and imagge preview dimensions 
+- Support for transparent background.
 
 Fixed:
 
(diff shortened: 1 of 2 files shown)


casperstorm:
@hecrj do iced always use pre-multiplied alpha? wondering if my https://github.com/squidowl/halloy/pull/1437/commits/c8a60dbc40916233e24cefae1b35c6af69d4f6fc makes sense.

hecrj:
This isn't something that should be fixed from userspace.

Pre-multiplying manually will cause incorrect alpha blending, since you'd be mixing pre-multiplied and post-multiplied colors.

Hummer12007:
@hecrj aha, so it needs to be fixed in iced?

hecrj:
https://github.com/iced-rs/iced/commit/f1fc77dfb2d728ead65a5c28175f059db7ffb3e5 should make things consistent. No need to ask your users to pre-multiply.

Hummer12007:
Dropped the disclaimer after we fixed it in iced.

PR #1439, opened 2026-01-30
add request_attention to notification

### LINKED ISSUE #1034: [Feature Request] Urgency hint
Is it possible to set the UrgencyHint flag for the client window if someone highlights you or sends a private message?
I don't know about wayland-protocol but sway supports urgency for native wayland clients

Hummer12007:
fixes #1034

### DIFF
--- a/book/src/configuration/notifications/README.md
+++ b/book/src/configuration/notifications/README.md
@@ -102,8 +102,12 @@ show_toast = true
 
 ### request_attention
 
-Notification should request user attention for the window (aka urgency).
-Triggers only when the window is not focused.
+Notification should request user attention for its window (aka urgency).  Exact
+behavior is platform specific:
+
+  - macOS: Bounces the dock icon once.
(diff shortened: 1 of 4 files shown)


Hummer12007:
@casperstorm do you think show_toast should imply this by default?

andymandias (approved):
Thank you for tackling this @Hummer12007!

I made a few changes to utilize some of the existing structures in Halloy, but just one functional change:  the toast and request_attention settings have been decoupled.  In testing it seems that request behavior behavior varies pretty wildly by platform, and in some cases it can be quite disruptive (changing virtual desktops to the one that has the window, then bringing it into focus).  And in other cases it's arguably less disruptive than a toast (flashing in the task bar or bouncing a dock icon).  So, we think it's best to allow users to activate one and not the other.

Anything you want to add @casperstorm?

casperstorm (approved):
Tested and works great

PR #1444, opened 2026-01-30
Sidebar nicklist

Hummer12007:
Disabled by default. Should keep things as is when not enabled.


Hummer12007:
Needs to be improved a little first.

casperstorm:
First of all, I like the feature. Good job implementing it. However, I don’t like the way we are “hammering” more complexity into the current side menu. The side menu is already quite a sore point for me. For once the `position` makes it feel a little awkward.

I have talked with @andymandias about a way forward, and we came up with a new system which also makes the entire thing a little more "halloy-ish". It will, however require quite a big refactor. A refactor which you shoudn't feel forced to do, i don't mind picking it up.

The main idea is ~~simply~~ to add two new `pane_grid`. A vertical one, and a horizontal one. For lack of better name we could call them `pane_grid_horizontal` and `pane_grid_vertical` here. Their purpose is to display different panes;
 
* pane1: external buffers: servers, channels, queries
* pane2: internal buffers: logs, highlights, file transfers
* pane3: nicklist
* pane4: ...

We can persist position and pane types to the `dashboard.json` file like the current pane_grid of buffers. Then the user is free to choose, resize and move the panes around as they want.

Each pane, eg external buffer, needs to answer a question of "can i be displayed both horizontally or vertically?" and if so implement that logic. For those panes that doesnt work horizontally, eg nicklist, we can then display a "This pane isn't supported in horizntal mode"

Then at the end of the day we end up with something like (pesudo code)

```
pane_grid_vertical
position: left
panes: nicklist, internal buffers

pane_grid_horizontal
position: bottom
panes: external buffers
```

Problems with this as i can see is:

- We might run down a rabbit hole just to figure out it isnt worth it
- When you click a pane we "select" it - that is not always something we want eg when switching channels.

Benefits:

- Easy to expand with new panes, eg network meter for ping to server etc.

This is up for discussion. 

Untitled-2026-02-05-1009(1)

casperstorm:
I'll prototype a bit to see how it turns out, and to see if this is a viable way to go.

Hummer12007:
An important thing I wanted to account for here:
2 ways to display the nicklist:
- Inside a single scrollable with the buflist (for folks with a small number of channels wanting a neat look)
- A fixed split (e.g. 2/1) for folks idling in a lot of channels like me which don't even fit on a screen

casperstorm:
Not quite ready for this here.

PR #1446, opened 2026-01-31
Improved SAFELIST Utilization

andymandias:
Ensures SAFELIST parameter is checked regardless of method used to open the Channel Discovery pane.  To make it simpler to ignore the lack of SAFELIST, provide a button to send a LIST command regardless of SAFELIST status.

Also includes a couple minor visualization updates to the top row button in scroll_view, to better match button theming elsewhere.


casperstorm:
Works great for me. A small ux suggestion would be to maybe just drop the tooltip and update title and button to something like:

Title
> This server lacks SAFELIST support. Requesting channels may disconnect you.

Button
> Fetch channels

PR #1460, opened 2026-02-10
deduplicate clippy lints in Cargo.toml

auronandace:
Instead of specifying the lints for the workspace and repeating them we can just use the workspace lints with a simple single line in the `Cargo.toml`.


casperstorm (approved):
Nice! Thanks a lot.

Issue #202, opened 2023-09-08
Feature request: Shift-Return to create a newline


jtbx:
Hey, nice client. On some other IRC clients you can hold down Shift and press Return to create a newline in your message. When you press Return to send the message, each line will be individually sent as a new message. I think Halloy should have something like this as it allows a user to send multiple messages at once without being interrupted in a busy channel.

casperstorm:
> i heard this is possible now in iced, right?

Yes! This is possible with https://github.com/iced-rs/iced/pull/2123

nc7s:
It would also be great to be able to Ctrl/Cmd-Enter to send ;)

Issue #1032, opened 2025-06-18
Support IRCv3 Multiline


cyrneko:
Not to be confused with #613, support for IRCv3's [Multiline Messages](https://ircv3.net/specs/extensions/multiline) would be very nice to have, just as a quality-of-life thing

That's all :)

4e554c4c:
i would really like to see this

PR #1464, opened 2026-02-11
Enable multi-line send

### LINKED ISSUE #614: Feature request: paste multiline text
Following https://github.com/squidowl/halloy/discussions/613
Creating a Feature request as suggested by @casperstorm.

Hi! Like this app. I'm using it on MacOS latest version via package.

Can't find a way to send multi line text. E.g. I want to send a small piece of code of 4 lines, I have to send each line separately. In other clients there are options like paste.

### Suggested feature: ability to send/paste multiline text

There could be numerous implementations. What could be done:

1. Implementation of `input box` (small pop-up dialog) where one could paste the text and send it. Which can be called via key combination or via `right click` on `Send message...` input box.
2. An option with fetching current system buffer and populate an input box (`paste window`) with it.

Снимок экрана 2024-10-15 в 11 50 56

3. Option to send either `msg` or `notice`.

Снимок экрана 2024-10-15 в 11 50 36

4. Pasted text then could be sent by Halloy via independent messages:
Снимок экрана 2024-10-15 в 11 54 55

Thanks!

casperstorm:
Fixes #614, #202.

Adds multi line input support: `Shift+Enter` adds a new line, and `Enter` sends each line one by one. You can add a delay in between if needed.

### DIFF
--- a/data/src/capabilities.rs
+++ b/data/src/capabilities.rs
@@ -22,13 +22,13 @@ pub enum Capability {
     InviteNotify,
     LabeledResponse,
     MessageTags,
+    Multiline,
     MultiPrefix,
     ReadMarker,
     Sasl,
     ServerTime,
     Setname,
     UserhostInNames,
-    Multiline,
 }
 
 impl FromStr for Capability {


lc-guy:
Is there a desire to support [+draft/multiline](https://ircv3.net/specs/extensions/multiline), if the server supports it?

4e554c4c:
> Is there a desire to support [+draft/multiline](https://ircv3.net/specs/extensions/multiline), if the server supports it?

I would like to support this, probably after this lands. see #1032

src/buffer/input_view.rs
```diff
@@ -1647,6 +1392,520 @@ impl State {
...
+                            multiline_batch_kind = Some(line_batch_kind);
+                        }
+
+                        multiline_byte_count += line_bytes;
+                        multiline_byte_count > multiline.max_bytes
+                    } else {
+                        true
+                    }
+                })
+                .map_or(max_lines, |position| position.max(1))
```
casperstorm, line 1439: do we need to count `\n` as well?

andymandias: In place of counting line breaks I'm relying on `lines` to contain single lines (no line breaks) and `take` to limit the maximum number of lines we will batch.

PR #1466, opened 2026-02-11
create a shared text wrapper to always add line_height from config

casperstorm:
this was planned for #1430 but i forgot it...
idea is just to have a shared `widget::text` now so we don't forget the `line_height` moving forward.


andymandias (approved):
Haven't tested but looks good to me. Approach will need modification when font reload support is added, but we can cross that bridge when we come to it.

PR #1474, opened 2026-02-14
Check for Updates via Proxy

andymandias:
Use the default proxy (if configured) to check for updates.

Also renames `sidebar.user_menu.show_new_version_indicator` to `check_for_update_on_launch` to make it clear that setting it to true will disable the check itself.

### DIFF
--- a/book/src/configuration/check-for-update-on-launch.md
+++ b/book/src/configuration/check-for-update-on-launch.md
@@ -2,7 +2,7 @@
 
 Controls whether Halloy will check the Halloy repository on launch for a new version of Halloy.  When a new version is found a dot indicator will appear on the user menu button and a menu item to open the release webpage will be added to the user menu.
 
-This can be useful if you rather want to rely on a package manager.
+This can be useful if you would rather rely on a package manager.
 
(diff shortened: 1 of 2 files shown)


book/src/configuration/check-for-update-on-launch.md
```diff
@@ -0,0 +1,25 @@
+# Check for Update on Launch
```
casperstorm, line 1: I wonder if we should make a small guide entry as well regarding upgrading Halloy?
Then we could link to this as well. In the guide we could write that we recommend using Halloy through a package manager, etc.

andymandias: Sounds good to me except the only advice I can think of at the moment is "use whichever package, or building from source, appeals to you".  Or should we link to the installation guide?

someone2037492034: Well, I personally think that it would be useful to have a section on updates, as some users might just install Halloy via their package manager and not want it to check for updates.

casperstorm (approved):

Issue #1449, opened 2026-02-02
Version 2026.2 hard crashes Windows 10 on install.


arcadepro:
At the end of the installation process, usually when Halloy .exe is launched, my desktop PC hard crashes (locks the entire machine). This didn't happen with any previous installations.

After a reboot, the 'halloy.exe' is present in it's install directory. Launching said executable results in an error. The crash occurs on an upgrade install or from a clean slate (no user data present).

The log file from the crashed installation doesn't get to the point of enumerating the GPU info.

The GPU info from the working 2026.1 installation is as follows.

```21:29:12.604:INFO -- Available adapters: [
    AdapterInfo {
        name: "NVIDIA GeForce GTX 970",
        vendor: 4318,
        device: 5058,
        device_type: DiscreteGpu,
        driver: "NVIDIA",
        driver_info: "582.28",
        backend: Vulkan,
    },
    AdapterInfo {
        name: "NVIDIA GeForce GTX 970",
        vendor: 4318,
        device: 5058,
        device_type: DiscreteGpu,
        driver: "32.0.15.8228",
        driver_info: "",
        backend: Dx12,
    },
    AdapterInfo {
        name: "Microsoft Basic Render Driver",
        vendor: 5140,
        device: 140,
        device_type: Cpu,
        driver: "6.2.19041.5794",
        driver_info: "",
        backend: Dx12,
    },
    AdapterInfo {
        name: "NVIDIA GeForce GTX 970/PCIe/SSE2",
        vendor: 4318,
        device: 0,
        device_type: Other,
        driver: "",
        driver_info: "4.6.0 NVIDIA 582.28",
        backend: Gl,
    },
]
21:29:12.821:INFO -- Selected: AdapterInfo {
    name: "NVIDIA GeForce GTX 970",
    vendor: 4318,
    device: 5058,
    device_type: DiscreteGpu,
    driver: "NVIDIA",
    driver_info: "582.28",
    backend: Vulkan,
}
```

Currently reverted to running 2026.1 which installed with no problems.

casperstorm:
@KaiKorla, do you have any idea?

casperstorm:
Strange, I booted into Window 11 (10.0.26100 Build 26100) and upgraded without a issue.
So it's perhaps a Windows 10 issue... not entirely sure how to test that.

arcadepro:
Since the log file from the crashed installation doesn't get to the point of enumerating the GPU info, I updated the first post with some GPU information.

[...]

PR #1475, opened 2026-02-14
feat(config): allow forcing wgpu backend and power preference via config

KaiKorla:
This adds a small escape hatch for users experiencing wgpu/driver specific issues. (#1449)

Since iced::daemon currently does not expose renderer backend selection, we apply the official wgpu override mechanism via environment variables early during startup.

The change is fully backwards compatible and opt-in.


casperstorm:
Thanks, @KaiKorla, while i appreciate the work here I would like us to wait for support from iced.
Iced will (soon) have runtime renderer configuration. I'll close this in favor for that.

Issue #1478, opened 2026-02-14
IRCv3 reactions


4e554c4c:
This feature request is a more specific version of #550, which pertains only to ircv3 reactions.

PR #1481, opened 2026-02-14
reacts: the third

4e554c4c:
Implement reactions in halloy

UI demo:
image

message (de)serialization backwards compatibility has been tested. This PR stores reactions on messages instead of separately in history metadata, which makes things easier. Unreactions are still stored, since the IRC chathistory protocol means that a reaction might be fetched _after_ its unreaction.

Resolves: #1478
Supersedes:  #1039

### DIFF
--- a/data/src/history.rs
+++ b/data/src/history.rs
@@ -1101,16 +1101,16 @@ pub fn find_reaction_target<'a>(
     // server_time is after a react
     let position = messages
         .iter()
-        .take(start_index + 1)
+        .take(start_index)
         .rev()
         .position(|m| m.id.as_deref() == Some(id))
-        .map(|position| start_index - position)
+        .map(|position| start_index - 1 - position)
         .or(messages
             .iter()
-            .skip(start_index + 1)
+            .skip(start_index)
             .rev()
             .position(|m| m.id.as_deref() == Some(id))
-            .map(|position| messages.len() - position));
+            .map(|position| messages.len() - 1 - position));
 
     position.and_then(|position| messages.get_mut(position))
 }
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -651,9 +651,6 @@ impl Client {
         }
 
         match &message.command {
-            _ if is_reaction(&message) => {
-                return Ok(vec![Event::Reaction(message)]);
-            }
             Command::BATCH(batch, params) => {
                 let mut chars = batch.chars();
                 let symbol = ok!(chars.next());
@@ -808,6 +805,9 @@ impl Client {
                         vec![]
                     } else {
                         match &message.command {
+                            _ if is_reaction(&message) => {
+                                vec![Event::Reaction(message)]
+                            }
                             Command::NICK(_) => batch_target
                                 .as_channel()
                                 .map(|channel| {
@@ -967,6 +967,9 @@ impl Client {
                         source,
                     )]);
                 }
+            }
+            _ if is_reaction(&message) => {
+                return Ok(vec![Event::Reaction(message)]);
             }
             // Reroute whois, whowas, and user mode responses
             Command::Numeric(
--- a/src/main.rs
+++ b/src/main.rs
@@ -710,24 +710,17 @@ impl Halloy {
                     );
 
                     if let Screen::Dashboard(dashboard) = &mut self.screen {
-                        let mut commands = vec![];
-
-                        for event in events {
-                            handle_client_event(
-                                &server,
-                                event,
-                                dashboard,
-                                &mut commands,
-                                &mut self.clients,
-                                &self.config,
-                                &mut self.notifications,
-                                &mut self.servers,
-                                &mut self.controllers,
-                                &self.main_window,
-                            );
-                        }
-
-                        return Task::batch(commands);
+                        return handle_client_events(
+                            &server,
+                            events,
+                            dashboard,
+                            &mut self.clients,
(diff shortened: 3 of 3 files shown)


4e554c4c:
this is tested with other clients (like the unreaction branch of goguma) and ready for review! not sure what's going on with the checks, I can't replicate the error.

data/src/config/buffer/channel.rs
```diff
@@ -14,10 +14,20 @@ pub struct Channel {
...
 pub struct Message {
     pub nickname_color: Color,
+    pub show_emoji_reacts: bool,
+}
+
+impl Default for Message {
+    fn default() -> Self {
+        Self {
+            nickname_color: Color::default(),
+            show_emoji_reacts: true,
```
casperstorm, line 28: Let's keep default as `false`.

data/src/command.rs
```diff
@@ -43,6 +43,12 @@ pub enum Irc {
     Nick(String),
     Quit(Option<String>),
     Msg(String, String),
+    React {
+        target: String,
+        msgid: message::Id,
+        text: String,
+        unreact: bool,
+    },
```
casperstorm, line 56: Maybe just split it out into React, Unreact to make it clear tbh.

data/src/message.rs
```diff
@@ -896,12 +928,68 @@ pub fn condense(
...
         None
     }
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct Reaction {
+    pub sender: Nick,
+    pub text: String,
+    pub unreact: bool,
+}
```
casperstorm: i think we should move Reaction, and ReactionContext into its own file under data/src/message/reaction.rs

data/src/message.rs
```diff
@@ -896,12 +928,68 @@ pub fn condense(
...
+
+#[derive(Debug)]
+pub struct ReactionContext {
```
casperstorm: Not a huge fan of "Context" in general, but right now i don't have a better word. Once moved to reactions.rs we could drop Reaction in front at least.

src/buffer/message_view.rs
```diff
@@ -324,6 +325,27 @@ impl<'a> ChannelQueryLayout<'a> {
...
+                    // SAFETY(pounce): our message must have an ID as it has reactions which we are
+                    // tracking
+                    msgid: message.id.as_ref().unwrap().clone(),
+                    text: text.to_owned(),
+                    unreacted: false,
+                },
+                |text| Message::Reacted {
+                    msgid: message.id.as_ref().unwrap().clone(),
+                    text: text.to_owned(),
+                    unreacted: true,
```
casperstorm: not happy about these unwraps.

src/buffer/scroll_view.rs
```diff
@@ -56,6 +58,11 @@ pub enum Message {
     ContentResized(Size),
     PendingScrollTo,
     HeightsCollected(Vec<(message::Hash, f32)>),
+    Reacted {
+        msgid: message::Id,
+        text: String,
+        unreacted: bool,
```
casperstorm: Again, we could split it to Reacted, Unreacted.

data/src/config/buffer/channel.rs
4e554c4c: I'm generally against hiding information by default, especially if it's intended communication (joining a server that has reacts enabled and then someone reacts to your message)

data/src/command.rs
4e554c4c: I agree that the boolean is unclear, but also having two different structs would mean duplication everywhere
maybe an enum with two variants would be better?

src/buffer/message_view.rs
4e554c4c: The unwraps are safe, but I can probably change the callbacks to have an Option<Message>

I think this would be a better choice in general for refactors in the future.

src/widget/reaction_row.rs
```diff
@@ -0,0 +1,127 @@
...
+                    .collect::<BTreeSet<&str>>(),
+            )
+        })
+        .filter(|(_, set)| !set.is_empty())
+        .collect();
+
+    let row = Row::from_iter(m.iter().map(|(reaction_text, nicks)| {
+        // we reacted to this message already
+        let selected =
+            our_nick.is_some_and(|nick| nicks.contains(&nick.as_str()));
```
casperstorm, line 64: could this fail due to casemapping?

data/src/history.rs
```diff
@@ -798,6 +799,30 @@ impl History {
...
+            | History::Full {
+                messages,
+                last_updated_at,
+                ..
+            } => {
+                let Some(message) =
+                    messages.iter_mut().find(|m| m.id.as_deref() == Some(&*id))
+                else {
+                    return;
+                };
```
casperstorm, line 881: i’m a bit concerned about the O(messages) lookup here 🤔

src/widget/reaction_row.rs
4e554c4c: no, since the `PartialEq` impl checks nicks that are normalized under the casemapping

4e554c4c:
~BLOCKING ISSUE: currently reactions don't show up if they occur after the backlog indicator~ fixed

casperstorm:
Ill merge this now, and then we can fine tune as things come. Great job on this @4e554c4c.

PR #1487, opened 2026-02-14
Allow cycling unread buffers with no current buffer

chamlis:
Treat this case as if an invisible 'zeroth' buffer is selected.

Not the prettiest index maths, and I considered using `rotate_{left,right}()` on `all` but decided to favour a less invasive change. I'm open to suggestions about potentially cleaner ways of expressing this.

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,10 +16,11 @@ Added:
 Fixed:
 
 - Channel mode in pane headers not updating immediately after mode changes
+- Cycle next/previous (unread) buffer shortcuts work from an empty pane
 
 Thanks:
 
-- Contributions: @Hummer12007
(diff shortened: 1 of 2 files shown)


andymandias (approved):
Good idea!  Looks like a good solution to me.  I expanded it to the non-unread versions of next/previous buffer as well.

Thank you!

chamlis:
Thanks for the quick review! For what it's worth, I think the non-unread versions worked before (the closure returned `None` due to `current?`, and then the `or_else` kicked in).

andymandias:
Ah yes, I believe you're right :facepalm:  Didn't think to test their behavior before modifying them to match your changes.  But I did test their behavior afterwards :sweat_smile:

PR #1500, opened 2026-02-17
Scripting

casperstorm:
Early draft of scripting for Halloy.
Initial version checklist:

- [X] Callbacks
  - [X] `on_connect` (when you connect to a given server)
  - [X] `on_join` (when someone joins a channel)
  - [X] `on_part` (when someone parts a channel)
  - [X] `on_message` (when a messages arrivies)
  - [X] `on_mode` (when a mode was changed)
  - [X] `on_start` (when script was loaded)
  - [X] `on_nick` (when a nick was changed)
- [X] Actions
  - [X] Command
  - [X] Log
- [x] UI
  - [x] Scripts buffer to start/stop  
- [X] Config
  - [X] Ability to autorun a script
- [ ] Return functions
  - [ ] Block message
  - [ ] Edit message
  - [ ] Pass message

```lua
-- reply with world! when someone writes hello in channel #foobar
function on_message(ctx, server, target, user, text, is_notice)
  if target ~= "#foobar" then return end
  if text:lower() == "hello" then
    ctx:command(string.format("PRIVMSG %s world!", target))
  end
end
```

### DIFF
--- a/data/Cargo.toml
+++ b/data/Cargo.toml
@@ -74,6 +74,7 @@ mime_guess = "2.0.5"
 any_ascii = "0.3.3"
 idna = "1.1.0"
 unicode-security = "0.1.2"
+mlua = { version = "0.11.4", features = ["lua54", "vendored", "send"] }
 
 [dev-dependencies]
 serde_test = "1.0"
--- a/data/Cargo.toml
+++ b/data/Cargo.toml
@@ -75,6 +75,7 @@ mime_guess = "2.0.5"
 any_ascii = "0.3.3"
 idna = "1.1.0"
 unicode-security = "0.1.2"
+mlua = { version = "0.11.4", features = ["lua54", "vendored", "send"] }
 
 [target.'cfg(target_os = "macos")'.dependencies]
 apple-native-keyring-store = { version = "1", default-features = false, features = [
--- a/data/Cargo.toml
+++ b/data/Cargo.toml
@@ -15,8 +15,16 @@ path = "src/bin.rs"
 required-features = ["message_tests"]
 
 [features]
+default = ["lua54", "lua-vendored"]
 tor = ["irc/tor"]
 message_tests = ["iced/tokio"]
+lua51 = ["mlua/lua51"]
+lua52 = ["mlua/lua52"]
+lua53 = ["mlua/lua53"]
+lua54 = ["mlua/lua54"]
+luajit = ["mlua/luajit"]
+luajit52 = ["mlua/luajit52"]
+lua-vendored = ["mlua/vendored"]
 
 [dependencies]
 thiserror = { workspace = true }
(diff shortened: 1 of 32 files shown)


classabbyamp:
Based on my hexchat scripts, here are some notes on what I'd like the scripting API to be able to do.
Some of this may already be implemented.

- ability to hook on all messages
    - or specific messages, like `on_message("005")`
    - ability to get all protocol message attributes (tags, source, command, args, etc)
- ability to get a list of:
    - networks
    - channels (in a network, in all networks)
    - users in a channel
    - this can tie in well with a context object, that allows you to get things like users in the *current* channel easily
- it might be nice to have a mechanism for script configuration, but this could be done through something like `require("myconf.lua")` easily (like neovim configs)
- ability to dynamically add/remove hooks
    - example: send command to nickserv and use a temporary hook to parse the response
    - maybe there's a better way to do this, especially with `labelled-response` or `reply` support, like allow sending a message and having a callback for the response (`ctx:command("PRIVMSG NickServ INFO foo", function() ... end )` perhaps...)
- ability to print a message to a buffer that isn't a protocol message
- ability to eat the message that triggered the hook (for everything, for following plugins, for display only)

- ability to register custom `/` commands
- ability to pre-fill the input buffer
    - example: generate a command that might be destructive, but don't automatically do it so the user can inspect before sending
- ability to create non-channel buffers
- ability to redirect messages to certain buffers

Additionally, something to think about:
- if external commands are run (with `io.popen()` or similar), does it block halloy? other plugins?
- if a script wants to send an irc command, interpret the response, then act, does it block halloy? other plugins?

B4reMetal:
This is an exciting pull request to see. Thanks for your continued work on this client. Scripting support would very likely help me and others dump our old unmaintained clients.

I think the previous comment sums up about everything that would be great to have in a scripting api.

It would also be nice if plugins/scripts were able to create and manage "mIRC style" dialog windows natively through iced. Basically a widget box with text inputs, buttons, etc. I'm sure this would add a lot of complexity however, especially if it is done safely. Just something to add to a future roadmap maybe.

Thanks again.

classabbyamp:
I'm not sure about the utility of having to manually list all the scripts you want to autoload. That seems a little backwards from how clients i'm familiar with load scripts.

data/Cargo.toml
```diff
@@ -74,6 +74,7 @@ mime_guess = "2.0.5"
 any_ascii = "0.3.3"
 idna = "1.1.0"
 unicode-security = "0.1.2"
+mlua = { version = "0.11.4", features = ["lua54", "vendored", "send"] }
```
classabbyamp: from a distro perspective, forcing `vendored` is majorly annoying. It'd also be nice to be able to use luajit (which means lua 5.1 or 5.2)

PR #1510, opened 2026-02-18
Condensed server message icon defaults to none

casperstorm:
Adds a new setting to change condense server message icon. It now defaults to "none".

image

### DIFF
--- a/src/buffer/message_view.rs
+++ b/src/buffer/message_view.rs
@@ -67,7 +67,27 @@ pub struct ChannelQueryLayout<'a> {
 }
 
 impl<'a> ChannelQueryLayout<'a> {
-    fn condensation_marker(&self, marker: Marker) -> Marker {
+    fn condensation_marker(
+        &self,
+        expanded: bool,
+        has_condensed: bool,
+    ) -> Marker {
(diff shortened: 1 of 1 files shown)


src/buffer/message_view.rs
```diff
@@ -67,6 +67,14 @@ pub struct ChannelQueryLayout<'a> {
...
 
 impl<'a> ChannelQueryLayout<'a> {
+    fn condensation_marker(&self, marker: Marker) -> Marker {
+        match self.config.buffer.server_messages.condense.icon {
+            CondensationIcon::None => Marker::None,
+            CondensationIcon::Chevron => marker,
+            CondensationIcon::Dot => Marker::Dot,
+        }
+    }
+
```
andymandias: Maybe we could have this take in `message.expanded` and `message.condensed.is_some()` instead of marker, then have all of the logic in this function?  Where the arguments could be hardcoded as `false` and `true` inside `format_condensed_message`.

andymandias (approved):
Looks good to me, and works well.  Had one idea for potentially tidying things up a little, but feel free to ignore if you like it better as-is.

PR #1511, opened 2026-02-19
Connect & Disconnect

andymandias:
Restructures how connection/disconnection are handled, to:
- Allow users to manually connect to disconnected servers via sidebar context-menu
- Allow servers to be disconnected without triggering auto-reconnect or being removed from the sidebar
- Allow users to specify servers that should not be connected to on launch (or when added via configuration reload)

Additionally:
- Adds context-menu entry for removing servers from the sidebar 
- Improves the handling of disconnect on SASL authentication failure
- Re-initiates connections when the connection details are changed as part of a configuration reload

### DIFF
--- a/data/src/stream.rs
+++ b/data/src/stream.rs
@@ -458,7 +458,8 @@ async fn _run(
                                 );
                                 state = State::Disconnected {
                                     autoconnect: updated_config.autoconnect,
-                                    retry: time::interval(
+                                    retry: time::interval_at(
+                                        Instant::now() + Duration::from_secs(1),
                                         config.reconnect_delay,
                                     ),
                                 };


src/screen/dashboard/sidebar.rs
```diff
@@ -708,16 +718,18 @@ impl Menu {
...
-    Leave,
     MarkAsRead,
     MarkServerAsRead,
     NewPane,
     Popout,
     Replace,
     Swap(window::Id, pane_grid::Pane),
+    Detach,
+    Leave,
+    Remove,
```
casperstorm, line 734: I think we should rename `Leave` to `Disconnect`.

casperstorm: Follow up. I think its bit strange to see both `Leave (Disconnect)` and `Remove`.
Perhaps should consider only showing `Remove` when they are `Disconnected`.

We could also use the seperator like we have on a user context:

Connected:

```
ergo <-- name of server to keep it similar with user context
----
Close all queries
Mark server buffer as read
Mark entire server as read
Open in new pane
Open in new window
Replace current pane
Disconnect <--- Connect if disconnected
Remove <--- Only shown if disconnected
```

PR #1519, opened 2026-02-19
Show/Hide previews context

casperstorm:
This adds ability to show/hide previews in url context.
I also modified the opengraph parsing as there were a bug where we parsed both property and name, but name overwrote property, so we treated it as twitter:image instead of og:image which resulted in... no image.

Can be tested with the following url:
https://nyheder.tv2.dk/samfund/2026-02-18-i-aarhus-er-eskimovej-snart-fortid-men-det-er-det-ikke-andre-steder-i-landet 
This wont give a preview in latest release, but will in this PR.

### DIFF
--- a/data/src/preview.rs
+++ b/data/src/preview.rs
@@ -470,11 +470,11 @@ mod tests {
         let meta = parse_meta_tag_properties(html).expect("should parse");
 
         assert_eq!(
-            meta.canonical_url.as_ref().map(|u| u.as_str()),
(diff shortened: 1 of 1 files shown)


andymandias (approved):
Looks good to me.  I don't know the details of the meta tag well enough to be able to say much about its parsing, but I don't see any issues in my testing.  Being able to show hidden previews is good QoL improvement.  Maybe we could add some tests with example tags we want to ensure we can parse?

Issue #1491, opened 2026-02-16
How to disable history in %appdata%/halloy/history?


c727:
How can I disable this as I dont need it? Thanks

PR #1520, opened 2026-02-19
Add configuration option for max_messages

melocene:
Resolves https://github.com/squidowl/halloy/issues/1491

This adds a new configuration option `max_messages` which can be used to disable storing history locally which increases user privacy. The setting is opt-in and still uses the 10,000 message limit originally set as a constant. The truncation count was not changed so that logic all remains the same. A small change was made as I noticed a bug in the logic that could potentially result in an error if `max_messages` was too small. Now if appropriate after truncation there is a max of 1 message. The new option was also added to the documentation in the appropriate section.

With manual testing, on Windows 11, I did not notice files written to disk with `max_messages` set to zero confirming the expected behavior. I didn't see unit tests for the config so I didn't add any here.

Claude Opus 4.6 was used to help me understand the codebase and get up to speed before making changes. I also used Claude to verify my logic and assumptions as I worked on the code. These changes were not vibe-coded, though Claude did help with with some wording and minor code tweaks.

### DIFF
--- a/book/src/configuration/buffer/chat-history/README.md
+++ b/book/src/configuration/buffer/chat-history/README.md
@@ -30,7 +30,7 @@ Set the number of messages recorded to the history files on disk
 
 ```toml
 # Type: integer
-# Values: any positive integer
+# Values: 0 or any positive intege
 # Default: 10000
 
 [buffer.chathistory]


casperstorm:
Thank you for the contribution, will focus on the upcoming release and then ill test this.

melocene:
@casperstorm I just rebased this onto main which should have resolved the conflicts due to other added features and config options.

PR #1524, opened 2026-02-20
Add more padding to theme editor on macOS

### LINKED ISSUE #1521: MacOS: Window controls overlap selector widget in theme editor
I noticed a potential issue on MacOS where the window controls overlap the selection widget in the left upper corner of the theme editor. See screenshot.

Image

It might be specific to Tahoe, but I am not sure.

Thanks for this awesome IRC client. :) Cheers!

casperstorm:
Fixes #1521

Screenshot 2026-02-20 at 21 30 00

### DIFF
--- a/src/screen/dashboard/theme_editor.rs
+++ b/src/screen/dashboard/theme_editor.rs
@@ -335,8 +335,11 @@ impl ThemeEditor {
             })
         });
 
-        let platform_specific_padding =
-            platform_specific::popped_out_window_padding(config);
+        let default_padding: u16 = 8;
+        let top_padding = platform_specific::popped_out_window_padding(config)
+            + u32::from(default_padding);
+        let padding =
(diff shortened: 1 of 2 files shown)


src/screen/dashboard/theme_editor.rs
```diff
@@ -347,10 +347,16 @@ impl ThemeEditor {
         ]
         .spacing(8);
 
+        let padding = if cfg!(target_os = "macos") {
+            Padding::new(8.0).top(30.0)
+        } else {
```
andymandias: Should we reference `config.platform_specific.macos.content_padding` here?

andymandias (approved):
Can't test myself, but the result looks good to me.

src/screen/dashboard/theme_editor.rs
casperstorm: We should!

Issue #1506, opened 2026-02-18
Add an ability to sort channels by the full name


Jagget:
Currently, case-insensitive alphabetical ordering  of the channels in the left panel ignores the prefix, but for some users that might seem inconvenient. 

Please, add the ability to NOT ignore the prefix and use the full channel name in alphabetical ordering.

andymandias:
I believe PR #1525 closed this ordering. Other orderings can be addressed with new issues.

PR #1525, opened 2026-02-21
Config option to include chantypes when sorting channels in the sidebar

englut:
Addresses #1506 

New configuration option:

```toml
[sidebar]
order_channels_by = "name-and-prefix"
```

Default is `"name"` (sorts without the prefixes), but when set to `"name-and-prefix"`, channels in the sidebar will include prefixes (chantypes) when sorting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@ Changed:
 
 Thanks:
 
-- Contributions: @Hummer12007, @chamlis
+- Contributions: @Hummer12007, @chamlis, @englut
 - Feature requests: @megumintyan, @4e554c4c, @RoboDanjal
 - Bug reports: @someone2037492034, @deepspaceaxolotl
 


data/src/client.rs
```diff
@@ -1693,12 +1693,18 @@ impl Client {
...
+                        config::sidebar::OrderChannelsBy::Name => false,
+                        config::sidebar::OrderChannelsBy::NameAndPrefix => true,
+                    };
                     let _ = self.chanmap.insert_sorted_by(
                         target_channel.clone(),
                         Channel::default(),
                         |c1, _, c2, _| {
                             compare_channels(
                                 &chantypes,
+                                include_chantypes,
```
andymandias: You could pass `config.sidebar.order_channels_by` into `compare_channels` here if you want, but we don't need to do that now.

We run everything through `rustfmt +nightly` to try and keep style consistent and non-formatting changes out of the commit log.  I think that's the only thing we need here.

andymandias (changes requested):
One minor change to get this passing tests then I think it's good to go.  Thank you for tackling this!

data/src/client.rs
englut: Oh! I gotta get in the habit of running this. Thank you! I've pushed up the fix and the change :)

PR #1530, opened 2026-02-21
perf: reduce allocations for history

4e554c4c:
Currently, two large multi-part (many append) allocations are needed for both writing to and reading history files. One allocation happens in `compression.rs` and one happens externally. The `Read` and `Write` traits exist so that this is not necessary. We can `read()` directly from a file and serialize it into the history struct.

Unfortunately not all allocations could be removed. Since it is usually called from an async context, and serde does not support `AsyncWrite`, we need a single allocation for writing to files. This is tricky to fix and I don't want things to get too complicated, so I am leaving this to a future PR.

My profiling has shown that approximately 20% of halloy's allocation time is spent in these two functions. Hopefully this speeds things up.

Test plan:
send and receive a message on a channel, close hallloy and reopen.

Test result:
history still works and is backwards compatable with previous history.

### DIFF
--- a/data/src/compression.rs
+++ b/data/src/compression.rs
@@ -7,20 +7,14 @@ use flate2::write::GzEncoder;
 use serde::Serialize;
 use serde::de::DeserializeOwned;
 
-pub fn compress<T: Serialize>(value: &T) -> Result<Vec<u8>, Error> {
-    let bytes = serde_json::to_vec(&value).map_err(Error::Encode)?;
-    let mut encoder = GzEncoder::new(Vec::new(), Compression::fast());
-    encoder.write_all(&bytes).map_err(Error::Compression)?;
-    encoder.finish().map_err(Error::Compression)
+pub fn compress<W: Write, T: Serialize>(w: W, value: &T) -> Result<(), Error> {
+    let encoder = GzEncoder::new(w, Compression::fast());
+    serde_json::to_writer(encoder, &value).map_err(Error::Encode)
 }
 
-pub fn decompress<T: DeserializeOwned>(data: &[u8]) -> Result<T, Error> {
-    let mut bytes = vec![];
-    let mut encoder = GzDecoder::new(data);
-    encoder
-        .read_to_end(&mut bytes)
-        .map_err(Error::Decompression)?;
-    serde_json::from_slice(&bytes).map_err(Error::Decode)
+pub fn decompress<R: Read, T: DeserializeOwned>(rdr: R) -> Result<T, Error> {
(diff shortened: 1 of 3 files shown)


data/src/history.rs
```diff
@@ -256,8 +258,8 @@ pub async fn delete(kind: &Kind) -> Result<(), Error> {
 }
 
 async fn read_all(path: &PathBuf) -> Result<Vec<Message>, Error> {
-    let bytes = fs::read(path).await?;
-    Ok(compression::decompress(&bytes)?)
+    let file = File::open(path)?;
```
casperstorm: why not `tokio::fs::File::open(path)`?

4e554c4c: because `tokio::fs::File` only implements `AsyncRead`, which isn't compatible with serde

casperstorm: correct, but i was thinking

```rust
let file = tokio::fs::File::open(path).await;
let std_file = file.into_std().await;
tokio::task::spawn_blocking(move || compression::decompress(std_file)).await
```

🤔

4e554c4c: ah nice, i didn't see `into_std` only `try_into_std`

data/src/compression.rs
```diff
@@ -7,20 +7,14 @@ use flate2::write::GzEncoder;
...
 use serde::de::DeserializeOwned;
 
-pub fn compress<T: Serialize>(value: &T) -> Result<Vec<u8>, Error> {
-    let bytes = serde_json::to_vec(&value).map_err(Error::Encode)?;
-    let mut encoder = GzEncoder::new(Vec::new(), Compression::fast());
-    encoder.write_all(&bytes).map_err(Error::Compression)?;
-    encoder.finish().map_err(Error::Compression)
+pub fn compress<W: Write, T: Serialize>(w: W, value: &T) -> Result<(), Error> {
+    let encoder = GzEncoder::new(w, Compression::fast());
+    serde_json::to_writer(encoder, &value).map_err(Error::Encode)
```
casperstorm, line 12: last thing, @4e554c4c - why drop `encoder.finish()` here?

4e554c4c: `serde_json::to_writer` consumes the `encoder`, so it's not possible to call `finish` anymore. When the encoder is dropped at the end of use, it will call `finish` as per its [drop method](https://docs.rs/flate2/latest/src/flate2/gz/write.rs.html#161-167)

casperstorm (approved):

casperstorm:
Great PR thanks a bunch

Issue #1492, opened 2026-02-16
Show messages from services in the server window instead of separate ones


c727:
Would be great if messages from services like `NickServ` or `HostServ` could be shown in the server window to avoid clutter

4e554c4c:
silly question: but is there any way to know whether something "is a service"?
afaik this hasn't been standardized in irc, and not all networks have them.

c727:
I think it would be good if you can configure them for each server. For example I'd also like to add `Global` which is used to send (useless) logon/random news

From xchat: `NickServ,ChanServ,InfoServ,N,Q` https://github.com/hexchat/hexchat/blob/b544ac3350e85d4cc41fe3414cbdb82d75ce5d7a/src/common/cfgfiles.c#L881

c727:
`BotServ,MemoServ`

casperstorm:
@c727 will be able to tackle it in #1532.

PR #1532, opened 2026-02-22
Reroute private messages

### LINKED ISSUE #142: feature request: config to choose where NOTICEs appear

casperstorm:
Fixes #142, #1492.

Ability to reroute private messages. Good for when talking with bots.
Below is an example where all messages from BotTheo will be rerouted to `#foobar`, including your own messages to him.

```toml
[servers.quakenet.reroute]
private_messages = [
  { user = "BotTheo", target = { channel = "#foobar" }}
]
```

### DIFF
--- a/data/src/config/server/filters.rs
+++ b/data/src/config/server/filters.rs
@@ -1,0 +1,109 @@
+use fancy_regex::{Regex, RegexBuilder};
+use serde::{Deserialize, Deserializer};
+
+#[derive(PartialEq, Eq, Debug, Clone, Deserialize, Default)]
+#[serde(default)]
+pub struct Filters {
+    pub ignore: Vec<Ignore>,
+    #[serde(deserialize_with = "deserialize_fancy_regexes")]
+    pub regex: Vec<FancyRegex>,
+}
+
+#[derive(PartialEq, Eq, Debug, Clone)]
+pub enum Ignore {
+    User(String),
+    UserInChannel { user: String, channel: String },
+    Regex { regex: FancyRegex },
+    RegexInChannel { regex: FancyRegex, channel: String },
+}
+
+impl<'de> Deserialize<'de> for Ignore {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        #[derive(Debug, Clone, Deserialize)]
+        #[serde(untagged)]
+        pub enum Inner {
+            User(String),
+            UserInChannel { user: String, channel: String },
+            Regex { regex: String },
+            RegexInChannel { regex: String, channel: String },
+        }
+
+        match Inner::deserialize(deserializer)? {
+            Inner::User(user) => Ok(Ignore::User(user)),
+            Inner::UserInChannel { user, channel } => {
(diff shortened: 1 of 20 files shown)


docs/configuration/reroute.md
```diff
@@ -0,0 +1,35 @@
...
+private and are not visible to other users in the channel.
+
+```toml
+# Type: array
+# Default: []
+
+[buffer.private_messages]
+reroute = [
+  { user = "Q", target = { channel = "#noc" } },
+  { user = "ChanServ", target = { server = "libera" } },
```
c727: Imo the rerouting should be limited within a server. This sounds like messages from all servers are rerouted to this server

casperstorm: Make sense.

casperstorm: Perhaps we should simply move it under each server. That would make more sense.

data/src/config/server/filters.rs
```diff
@@ -0,0 +1,109 @@
+use fancy_regex::{Regex, RegexBuilder};
+use serde::{Deserialize, Deserializer};
+
+#[derive(PartialEq, Eq, Debug, Clone, Deserialize, Default)]
+#[serde(default)]
+pub struct Filters {
```
casperstorm, line 6: Moved filters now that i created a new folder for reroute. Seemed fitting.

c727:
Maybe the config could be simpler. Something like `private_messages = [user1 = "#channel1", user2 = server]`

But I'll only configure it once, so I won't care that much 🤪

casperstorm:
> Maybe the config could be simpler. Something like `private_messages = [user1 = "#channel1", user2 = server]`
> 
> But I'll only configure it once, so I won't care that much 🤪

I am mainly trying to ensure future maintainability with the proposed config.

casperstorm:
It should be in a good state to test, if you have any feedback @c727.

data/src/message.rs
```diff
@@ -86,6 +86,96 @@ pub mod formatting;
...
+                reroute_private.has_server_reroute_rule_for(raw_target, server)
+            } else if let Source::User(user) = source {
+                reroute_private
+                    .has_server_reroute_rule_for(user.as_str(), server)
+            } else {
+                false
+            }
+        }
+        _ => false,
+    }
```
andymandias: It's a bit tedious but I think we should do casemap normalization for these comparisons.

4e554c4c:
Is there any reason this can't include supporting the [`channel-context`](https://ircv3.net/specs/client-tags/channel-context) client tag?

andymandias:
> Is there any reason this can't include supporting the [`channel-context`](https://ircv3.net/specs/client-tags/channel-context) client tag?

None that I can see.  This is currently languishing because of me, but I will be returning to it soon.  Thanks for the reference, it does clarify what needs to be done for this PR.

andymandias (approved):

PR #1538, opened 2026-02-23
Ensure window position is always saved on exit

melocene:
Halloy was returning before the code executed to save the window position meaning it was not reliably updated. In this PR I just added it to the cleanup tasks when the window is being closed so it should save the data before closing cleanly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Unreleased
 
+Fixed:
+
+- Window state is saved even if altered shortly before quitting
+
+Thanks:
+
+- Contributions: @melocene
+
 # 2026.3 (2026-02-24)
 
 Added:


src/main.rs
```diff
@@ -856,14 +856,22 @@ impl Halloy {
...
-                                    .map(Message::Dashboard);
+                                return Task::batch(vec![
+                                    save,
+                                    dashboard
+                                        .exit(&mut self.clients, &self.config)
+                                        .map(Message::Dashboard),
+                                ]);
                             } else {
-                                return iced::exit();
+                                return Task::batch(vec![save, iced::exit()]);
```
andymandias: `Task::batch` doesn't give us any guarantee on the order of execution, so I think we should do `save.chain(iced::exit())` here to ensure the save is done before exiting.  And by the same logic I think the same should be done for the return in the if branch above (although I think it's much less likely to be an issue for that return).

melocene: Should be good to go now. I swapped both instances of `Task::batch` for the `save::chain`. Hopefully that is what you meant.

PR #1546, opened 2026-02-24
Mark focused buffers as read on incoming messages

### LINKED ISSUE #1544: Don't Show Backlog Line If Buffer Is Open and Read
When a buffer is open and has been read (i.e. doesn't already have a backlog separator from having had unread messages before opening it, that hasn't been dismissed by marking the channel as read or scrolling to the bottom with that setting on), new messages make the backlog separator show up even when Halloy is open and the focused window. I would expect new messages arriving in an open buffer that is in focus to not be treated as if they're unread.

At the moment, dismissing the backlog separator in this scenario requires either marking messages as read using the button, scrolling up and then down with the setting to mark messages when buffer is scrolled to the bottom on, or switching back and forth between channels (with the relevant setting to mark a buffer as read when closing it set, I believe). Or sending a message. All are some level of annoying, so improvement here would be welcome!

casperstorm:
Fixes #1544.

When you have a focused a buffer, and are at the very bottom you don't trigger any backlog indicators.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -771,19 +771,22 @@ impl Halloy {
 
                         let mut bouncer_network_commands = vec![];
 
-                        for (bouncer_network, events) in bouncer_network_events {
-                            bouncer_network_commands.push(handle_client_events(
-                                &bouncer_network,
-                                events,
-                                dashboard,
-                                &mut self.clients,
-                                &self.config,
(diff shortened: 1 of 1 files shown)


andymandias (approved):
Cleanly executed, works great.

PR #1558, opened 2026-02-25
Record window size if maximized or not

melocene:
When recording the window data into `window.json` I noticed it didn't have a way to record if the window was maximized or not. The `fullscreen` option seems different, at least on Windows, and doesn't meet the same goal I think.

I was able to at least verify this functionality on Windows 11 and confirmed that closing and reopening `halloy` will maximize the window if that is its state when closed, whereas previously it did not. I tried to keep the changes as minimal as possible so feel free to propose any additional changes and I am happy to make them. Existing data is still written and respected as it was previously. This is just one new field defaulting to not maximized if unset or not found.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Added:
 
 - Support for the IRCv3 [`react`](https://ircv3.net/specs/client-tags/react.html) client tag.
+- Correctly record window size if window is maximized
 
 Changed:
 


andymandias (approved):
Works well for me on GNOME/Pop!_OS.  Haven't tested Windows/macOS.

I think it warrants an entry under fixed in the changelog.

casperstorm (approved):

PR #1563, opened 2026-02-26
Add support for notification subtitle on supported OSes

bmndc:
Halloy currently show notification titles that are too long and often be truncated by macOS. notify_rust supports displaying notification subtitle on macOS and Windows. It might be better to display channel and server info there when show_content is enabled for certain types of notifications. Heavily inspired by Discord notification template :)

On Linux, the subtitle line is wrapped in parentheses and appended to the title line.

TODO: might need extra testing on Windows because adding the subtitle line might display less message content.

Halloy notification for a new channel message
Halloy notification for a new direct message

### DIFF
--- a/src/notification.rs
+++ b/src/notification.rs
@@ -265,10 +265,7 @@ impl Notifications {
                         self.execute(
                             &config.highlight,
                             notification,
-                            &format!(
-                                "{} {description}",
-                                user.nickname()
-                            ),
+                            &format!("{} {description}", user.nickname()),
                             Some(if cfg!(target_os = "linux") {
                             	format!("{channel}, {server}")
                             } else {


bmndc:
Glad this is making its way to upstream. Thank you both! :3

src/notification.rs
```diff
@@ -266,7 +266,7 @@ impl Notifications {
                             &config.highlight,
                             notification,
                             &format!("{} {description}", user.nickname()),
-                            Some(format!("{channel}, {server}")).as_deref(),
+                            Some(format!("{channel} ({server})")).as_deref(),
```
bmndc: On Linux, since there's no support for the subtitle line, it is wrapped in parentheses and appended to the title line. My idea was something like `casperstorm (#halloy, liberachat)`

This will make it look like `casperstorm (#halloy (liberachat))` which some may find a bit too extra.

Just a quick note, no big pressure!

andymandias: Thanks for the note.  Maybe we use that style for Linux and the other for non-Linux then?  Mainly looking to keep the styling as consistent as possible throughout.

bmndc: With the current code, that has to be done on the notification types themselves. This was done with my minimal understanding of Rust, but I think we might be able to carry over what's done in toast.rs to the notification code. I'll test if that works...

bmndc: Screenshot of a Halloy chat notification

Haven't tested on Linux, but seems to work.

PR #1573, opened 2026-02-28
Code Spans

andymandias:
Updates to encoding and parsing with respect to code spans:
- Some fixes in how markdown parsing is performed
- Update the code span escapes to be a bit closer to Commonmark
- When receiving code spans, ignore non-code formatting codes inside the code span
- When receiving code spans, do not try to parse links within the code spans

The last two items are chosen because there is no existing functionality to have multiple font styles within a code block (and it looks like current iced capabilities make it exceedingly difficult).  So, code span integrity is given priority over respecting all formatting codes.  When that capability is available we should probably revisit that decision.

### DIFF
--- a/data/src/message/formatting/encode.rs
+++ b/data/src/message/formatting/encode.rs
@@ -6,7 +6,8 @@ use nom::branch::alt;
 use nom::bytes::complete::tag;
 use nom::character::complete::{anychar, char, satisfy};
 use nom::combinator::{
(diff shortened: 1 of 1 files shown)


andymandias:
Currently supports 1-3 backticks (with or without spaces) for delimiting code spans;  converted to draft to see if it can be simply restructured to support an arbitrary number as requested in #1578.

casperstorm (approved):
Tested and works great. Also really nice simplification.

casperstorm:
PS i rebased for you :)

PR #1574, opened 2026-02-28
New about screen

casperstorm:
Adds a new about screen with a bit of extra information

screenshot-2026-02-26_22-20-54


src/modal/about.rs
```diff
@@ -0,0 +1,170 @@
...
+                logo,
+                column![
+                    item("Version", &self.version, Field::Version),
+                    item("Commit", &self.commit, Field::Commit),
+                    item("Contact", data::environment::EMAIL, Field::Mail),
+                    container(rule::horizontal(1))
+                        .padding([6, 0])
+                        .width(Length::Fill),
+                    item("GPU Backend", gpu_backend, Field::GpuBackend),
+                    item("GPU Adapter", gpu_adapter, Field::GpuAdapter),
```
andymandias: Maybe the labels could be clickable too?  I think that could be convenient, but maybe it could be confusing?

I also think it'd be nice to have a "Copy All" button that would include labels and fields, so users could easily copy all relevant info into an issue.

src/modal/about.rs
```diff
@@ -0,0 +1,170 @@
...
+    copied: Option<Field>,
+}
+
+impl About {
+    pub fn new(
+        version: String,
+        commit: String,
+        system_information: Option<iced::system::Information>,
+    ) -> Self {
+        let logo_bytes = include_bytes!("../../assets/logo.png").to_vec();
```
andymandias: For me this winds up being a little crunchy with the scale to 128px.  The pre-scaled version `../../assets/linux/icons/hicolor/128x128/apps/org.squidowl.halloy.png` looks smoother.  Maybe we could bless a pre-scaled 128px wide version of the logo specifically for the about screen?

casperstorm: Make sense to me. Looks much better!
Also changed it in the welcome screen.

src/modal/about.rs
casperstorm: Added!

Issue #1576, opened 2026-02-28
Setting to lowercase channel names in the sidebar.


RealKindOne:
A small fractions of channels start with a uppercase letter, end with uppercase letters, or start and end with uppercase. 

 Libera.Chat examples: `##Llamas`  `#idleRPG` `#RevolutionIRC`

What about adding a setting to visually lowercase them in the sidebar? I think it would make the channel sidebar look better.

Suggested setting:

Lowercase letters A-Z in the channel name in the channel sidebar.
```
# Type: boolean
# Values: true, false
# Default: false

[sidebar]
lowercase_channel = true

```

RealKindOne:
Can the channel name in the channel pane also be lowercased with this setting or should it use a different setting?

englut:
> Can the channel name in the channel pane also be lowercased with this setting or should it use a different setting?

I think it should be a separate config option in `[buffer]`

```
# Type: boolean
# Values: true, false
# Default: false

[buffer]
lowercase_channels = true
```

Alternatively, could have it as an enum with something like this (same for sidebar of course):

```
# Type: string
# Values: "original", "lowercase", "uppercase"
# Default: "original"

[buffer]
channel_case = "lowercase"
```

casperstorm:
Fixed by https://github.com/squidowl/halloy/pull/1581.
API will be:

```
[buffer.channel]
channel_name_casing = "lowercase"

[sidebar]
channel_name_casing = "lowercase"
```

Then in future we can expand on the functionallity better if we want.

PR #1581, opened 2026-03-01
feat(config): add optional lowercase channel titles for sidebar and pane

KaiKorla:
Adds support for requested features in #1576

### DIFF
--- a/data/src/config/buffer/channel.rs
+++ b/data/src/config/buffer/channel.rs
@@ -12,6 +12,8 @@ pub struct Channel {
     #[serde(alias = "topic")] // For backwards compatibility
     pub topic_banner: TopicBanner,
     pub message: Message,
+    #[serde(default)]
+    pub lowercase_channel: bool,
 }
 
 #[derive(Debug, Clone, Deserialize)]
--- a/data/src/config/buffer/channel.rs
+++ b/data/src/config/buffer/channel.rs
@@ -4,6 +4,7 @@ use super::NicknameClickAction;
 use crate::buffer::Color;
 use crate::channel::Position;
 use crate::config::buffer::{AccessLevelFormat, Away};
+use crate::isupport;
 
 #[derive(Debug, Clone, Default, Deserialize)]
 #[serde(default)]
@@ -12,8 +13,21 @@ pub struct Channel {
     #[serde(alias = "topic")] // For backwards compatibility
     pub topic_banner: TopicBanner,
     pub message: Message,
-    #[serde(default)]
-    pub lowercase_channel: bool,
+    pub channel_name_casing: Option<ChannelNameCasing>,
+}
+
+#[derive(Debug, Copy, Clone, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub enum ChannelNameCasing {
+    Lowercase,
+}
+
+impl ChannelNameCasing {
+    pub fn apply(&self, name: &str, casemapping: isupport::CaseMap) -> String {
+        match self {
+            Self::Lowercase => casemapping.normalize(name),
+        }
+    }
 }
 
 #[derive(Debug, Clone, Deserialize)]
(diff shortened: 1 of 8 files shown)


src/screen/dashboard/pane.rs
```diff
@@ -64,7 +64,13 @@ impl Pane {
         let title_bar_text = match &self.buffer {
             Buffer::Empty => String::new(),
             Buffer::Channel(state) => {
-                let channel = state.target.as_str();
+                let channel = config
+                    .buffer
+                    .channel
+                    .lowercase_channel
+                    .unwrap_or(false)
+                    .then(|| state.target.to_string().to_lowercase())
```
4e554c4c: is there any reason this can't lowercase by the casemapping?

data/src/config/buffer/channel.rs
```diff
@@ -12,6 +12,7 @@ pub struct Channel {
     #[serde(alias = "topic")] // For backwards compatibility
     pub topic_banner: TopicBanner,
     pub message: Message,
+    pub lowercase_channel: Option<bool>,
```
4e554c4c: it's probably better to have this be a `bool` with a default value (like `show_emoji_reacts` below)

data/src/config/sidebar.rs
```diff
@@ -25,6 +25,7 @@ pub struct Sidebar {
     pub padding: Padding,
     pub spacing: Spacing,
     pub order_channels_by: OrderChannelsBy,
+    pub lowercase_channels: Option<bool>,
```
4e554c4c: ditto

data/src/config/buffer/channel.rs
4e554c4c: this also needs a configuration entry in the book (https://github.com/squidowl/halloy/blob/main/book/src/configuration/buffer/channel/README.md) see the rest of the book for examples

4e554c4c (changes requested):

src/screen/dashboard/pane.rs
KaiKorla: Hmm, this is purely a presentation layer transformation. it's not changing stored identifiers or comparison behavior, just how the channel name is displayed in the view. the internal logic continues to use the original name and canonical IRC casemapping where needed.

data/src/config/buffer/channel.rs
KaiKorla: Done & done.

data/src/config/sidebar.rs
KaiKorla: Done.

src/screen/dashboard/pane.rs
4e554c4c: my main worry is from a user perspective. If you see a channel in the list and do `/msg #channel` it should work. If your channel is `#Über` and the casemapping is anything but rfc7613, then this will show up as `#über` in the list but messaging `/msg #über` refers to a different channel.

KaiKorla: ~~Hmm, ~~you mean this edge-case for rfc7613, i think.~~. Other way around: So we should allow this setting if the server is reporting rfc7613 for channel names?~~

I think i got it: If the server is reporting CASEMAPPING != rfc7613 for channel names and also the channel name contains non-ASCII, then it should be ignored.

4e554c4c: Afaik rfc7613 is pretty dang uncommon (although i think ergo uses it) so that would make things fairly useless 
`to_ascii_lowercase` also works for all casemappings but it might not be as pretty

I'd be happy with either that or a warning in the book

KaiKorla: Lowercasing is now restricted to ASCII-only channel names unless the server advertises rfc7613 casemapping, preventing any semantic discrepancies.

4e554c4c: it seems easier to do `casemapping.normalize(str)` instead of this (since it has the same effect most of the time)

but up to you. im happy with this now either way.

casperstorm (approved):
Changed to a enum so we can in the future support othe casings if we want.
Otherwise looks good thanks.

PR #1593, opened 2026-03-03
Implement channel monitor Buffer/pane

### LINKED ISSUE #1338: Feature Request: Channel monitor like Weechat AdiIrc etc.
Feature:  A buffer mode/pane like the highlight window that shows all messages from joined channels. Examples of this are AdiIrc channel monitor and weechat chanmon plugin. [https://wiki.adiirc.com/projects/adiirc/wiki/Monitor_Panels](url) [https://weechat.org/scripts/source/chanmon.pl/](url)

Death916:
Inspired by https://weechat.org/scripts/source/chanmon.pl/ and https://dev.adiirc.com/projects/adiirc/wiki/Monitor_Panels_Options

https://github.com/squidowl/halloy/issues/1338a
#1338


casperstorm:
I took it for a spin and my first thought was; could simplify the code by re-using `buffer::highlights`?
It is technically the same view, and the only difference is the data: highlights vs all messages. Both gets data from across all connected servers and channels. I think we could rename it and reuse it. Perhaps we call it `internal_monitor` or just `monitor` with a `enum Kind` like:

```rust
#[derive(Debug, Clone, Copy)]
pub enum Kind {
    Highlights,
    Messages, // Feels like a better name than ChannelMonitor?
}
```

Death916:
> I took it for a spin and my first thought was; could simplify the code by re-using `buffer::highlights`? It is technically the same view, and the only difference is the data: highlights vs all messages. Both gets data from across all connected servers and channels. I think we could rename it and reuse it. Perhaps we call it `internal_monitor` or just `monitor` with a `enum Kind` like:
> 

IM not particularly attached to the name just used the name from AdiIrc which it was inspired by. Also monitor is already a feature so i didnt want to have another variable named same thing for clarity.  If the simpler name seems better sounds good to me. 

Copied form my irc message:
do you mean use the existing highlights.rs and and add an enum there? cuz i already did reuse the highlight code for the main bulk of what i added
 "I think we could rename it and reuse it." is what i did pretty much.

\Sorry if im misunderstanding.

casperstorm:
> do you mean use the existing highlights.rs and and add an enum there? cuz i already did reuse the highlight code for the main bulk of what i added
> "I think we could rename it and reuse it." is what i did pretty much.

I am thinking we could have a single file, which is a way to represent messages across multiple channels/servers. I kinda see we could then also use that same path for when we add searching, perhaps @andymandias can weigh in here but yes a enum in there was what i was thinking.

Death916:
@casperstorm  I looked into combining them a bit, and in my opinion it would clutter up highlights.rs as it is and need logic added for which view to show (chanmon/highlights) especially if both are open. This way  I think is more clean.

But, I see what you mean about using a combined version to make searching easier but wouldn't that just be more of an extension on the existing history stuff? You kknow the codebase so I could be way off. If you did want to combine it Id love to hear some ideas on best way to do it , as it would be a much bigger architecture change than Id want to do, alone at least lol. I played around with implementing it that way but decided to pause for a bit for input.

casperstorm:
I'd like to help with this here next. Can you rebase it and ill take a look.

Death916:
@casperstorm missed your comment but ya let me get back up to date and re send

Death916:
how it looks

assets/fontello/config.json
andymandias, line 1: Maybe this can be reverted since we're using the SVG for channel monitor?  Okay by me if we want to keep a record of what was pulled from Fontello, but I think we wouldn't want to generate a new font file from this config (to keep the font file size as small as possible).

andymandias (approved):

casperstorm (approved):

PR #1597, opened 2026-03-03
Setting to limit reaction text length

casperstorm:
Added a default cap of 5 characters to reactions, but made it configurable.

### DIFF
--- a/data/src/reaction.rs
+++ b/data/src/reaction.rs
@@ -68,9 +68,11 @@ pub fn truncate_text(text: &str, max_chars: usize) -> String {
         return text.to_string();
     }
 
-    UnicodeSegmentation::graphemes(text, true)
+    let mut truncated = UnicodeSegmentation::graphemes(text, true)
         .take(max_chars)
-        .collect()
+        .collect::<String>();
+    truncated.push_str("...");
+    truncated
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


data/src/reaction.rs
```diff
@@ -60,6 +63,16 @@ impl Reaction {
...
 }
 
+pub fn truncate_text(text: &str, max_chars: usize) -> String {
+    if UnicodeSegmentation::graphemes(text, true).count() <= max_chars {
+        return text.to_string();
+    }
+
+    UnicodeSegmentation::graphemes(text, true)
+        .take(max_chars)
+        .collect()
```
andymandias: Should we add an ellipsis to indicate the reaction was truncated?

data/src/reaction.rs
```diff
@@ -38,6 +40,7 @@ impl Reaction {
             (None, Some(s)) => (s.clone(), true),
             _ => return None,
         };
```
4e554c4c, line 42: I would honestly just reject instead of truncating

4e554c4c:
LGTM!
we'll have to change it when custom emojis come around but i think this is a good solution. Afaik goguma currently rejects anything that's not a single grapheme

andymandias (approved):
Looks good to me, even if it means we may not find out what happens next in the Bee movie.

PR #1598, opened 2026-03-03
fix(emojipicker): click-through and mouse selection

KaiKorla:
- prevent click-through on completion picker overlays
- allow mouse selection for emoji and command picker entries

### DIFF
--- a/src/appearance/theme/button.rs
+++ b/src/appearance/theme/button.rs
@@ -125,6 +125,25 @@ pub fn secondary(theme: &Theme, status: Status, selected: bool) -> Style {
     button(foreground, background, background_hover, status)
 }
 
+pub fn picker(theme: &Theme, status: Status, is_selected: bool) -> Style {
+    let foreground = theme.styles().text.primary.color;
+    let button_colors = theme.styles().buttons.primary;
+
+    let background = if is_selected {
+        button_colors.background_hover
+    } else {
+        button_colors.background
+    };
+
+    let background_hover = if is_selected {
+        button_colors.background_selected
+    } else {
+        button_colors.background_hover
+    };
+
+    button(foreground, background, background_hover, status)
+}
+
 pub fn bare(_theme: &Theme, status: Status) -> Style {
     match status {
         Status::Active | Status::Pressed | Status::Hovered => Style {
--- a/src/buffer/input_view/completion.rs
+++ b/src/buffer/input_view/completion.rs
@@ -17,7 +17,7 @@ use data::user::{ChannelUsers, Nick, NickRef};
 use data::{Config, mode};
 use iced::Length;
 use iced::widget::text::Shaping;
-use iced::widget::{column, container, row, text_editor, tooltip};
+use iced::widget::{button, column, container, row, text_editor, tooltip};
 use irc::proto;
 use itertools::{Either, Itertools};
 use strsim::jaro_winkler;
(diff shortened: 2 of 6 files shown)


src/buffer/input_view/completion.rs
```diff
@@ -2863,16 +2886,22 @@ impl Emojis {
...
-                                .style(if selected {
-                                    theme::container::primary_background_hover
-                                } else {
-                                    theme::container::none
-                                })
-                                .padding(6)
-                                .center_y(Length::Shrink),
+                                .padding(0)
+                                .style(theme::button::bare)
+                                .on_press(on_select_emoji(*index)),
```
andymandias: We can improve the mouse feedback here, and simplify things a bit, by adding a new theme to `src/appearance/theme/button.rs` and using it here.  Let's call it `picker`, and it should work similar to `primary` except I think we want `button_colors.background_hover` and `button_colors.background_selected` for `background` and `background_hover` if `selected`.  If we apply that theme to the button, and give it the 6 padding, then we can place `content` directly in the button and get rid of the `row` altogether.

src/buffer/input_view/completion.rs
```diff
@@ -125,6 +125,14 @@ impl Completion {
...
     }
 
+    pub fn select_emoji(
+        &mut self,
+        index: usize,
+        config: &Config,
+    ) -> Option<Entry> {
+        self.emojis.select_at(index, config).map(Entry::Emoji)
+    }
+
```
andymandias: Since we're adding mouse interaction to the emoji picker, we should add it to the command picker as well.  We can make this `select_at` and have it operate similarly to `select` above (with a new `select_at` function added to `Commands`).  Similarly for `SelectEmoji`, etc.

andymandias (changes requested):
I think this is a nice addition to the picker overlay.  A few small changes I think we should make, but overall it's looking good.

src/appearance/theme/button.rs
```diff
@@ -125,6 +125,25 @@ pub fn secondary(theme: &Theme, status: Status, selected: bool) -> Style {
...
+        button_colors.background_selected
+    } else {
+        button_colors.background_hover
+    };
+
+    let background_hover = if selected {
+        button_colors.background_hover
+    } else {
+        button_colors.background_selected
+    };
```
andymandias: Stylistically I think this would look a little nicer:
```rust
    let background = if is_selected {
        button_colors.background_hover
    } else {
        button_colors.background
    };

    let background_hover = if is_selected {
        button_colors.background_selected
    } else {
        button_colors.background_hover
    };
```

src/buffer/input_view/completion.rs
```diff
@@ -183,15 +194,19 @@ impl Completion {
...
 
-    pub fn view<'a, Message: 'a>(
+    pub fn view<'a, Message: Clone + 'a>(
         &self,
         input: &str,
         server: &Server,
         config: &Config,
         theme: &'a Theme,
+        on_select_command: impl Fn(usize) -> Message + Copy + 'a,
+        on_select_emoji: impl Fn(usize) -> Message + Copy + 'a,
```
andymandias: Since these are coming in as the same function we can collapse this to one argument.  If we need to differentiate them later we can do that then 🙂

andymandias (approved):
Looks good, thank you!

PR #1599, opened 2026-03-03
fix(modal): block interaction and hover behind open overlays

KaiKorla:
- stop forwarding mouse/keyboard/touch input to base content while a modal is open
- prevent cursor/hover feedback from underlying widgets behind the modal
- keep existing modal-close behavior via outside click and Escape

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Fixed:
 
 Thanks:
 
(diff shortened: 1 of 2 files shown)


andymandias (approved):
Looks good to me, tested with URL prompt and About Window and everything is working as intended.

PR #1600, opened 2026-03-04
Security and operational improvements

melocene:
This PR is for a handful of security and operational improvements for the project, of which I use as well in most projects. I made sure each was a separate commit so maintainers can cherry-pick certain changes while leaving out others as desired. A couple workflows did end up with minor tweaks for quoting or spacing due to my editor but these are not functional changes, just stylistic and will not affect how Actions parses the workflows.

- Add `dependabot.yml` to help keep workflows and dependencies up to date. This will help with keep updated with security and other performance fixes that may come through. I set this to monthly for Rust dependencies and weekly for GitHub actions. This was chosen based on potential and reasonable churn without being too noisy.
- Ensure caching is utilized in all possible locations. The cache steps were missing the fallback key restoration pattern so could end up in a state where nothing was restored from cache. I updated this and split the cache into debug/test and release to share the cache sensibly for best utilization.
- Added a security scanning workflow which utilizes CodeQL and `cargo audit`. CodeQL can help help catch code level security issues, such as cleartext logging or storage of sensitive data, various injection attacks, and more. CodeQL can also trace through code and data flows for potential issues. `cargo audit` can capture and report on higher level security issues with dependencies. Neither of these should block by default unless they become required checks in the repository settings.

**Note:** By default the `cargo audit` job will open issues for its findings. It will not auto-close them. This can be tweaks via options to the job.

- I added a release profile to `Cargo.toml` which I commonly use on other projects. The settings I chose help enable as many optimizations as possible while minimizing binary size and maximizing performance. This does add to the build time however.
- `cargo-edit` was being built each time via CI consuming runner resources and adding to build and check times. This was swapped for another action which can download a precompiled binary to speed up those steps.
- The Linux build was not using `--locked` which I noticed was present for the other OS so I went ahead and just added that for consistency.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -63,3 +63,4 @@ jobs:
       - uses: rustsec/audit-check@69366f33c96575abad1ee0dba8212993eecbe998 # v2.0.0
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
+          ignore: RUSTSEC-2023-0071 # rsa: Marvin Attack - no fix available, transitive via arti-client


.github/workflows/build.yml
```diff
@@ -29,46 +29,48 @@ jobs:
...
+            pkg-config \
+            mold \
+            clang \
+            libdbus-1-dev \
+            libudev-dev \
+            libxkbcommon-dev \
+            libfontconfig1-dev \
+            libasound2-dev \
+            libx11-xcb-dev
+      - uses: actions/cache@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5.0.3
```
classabbyamp, line 37: using commit shas instead of tags can be an [attack vector](https://cloudisland.nz/@xssfox/115335067424247437)

melocene: I was following the recommendation from the documentation [here](https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions). Happy to switch this bag to a tag, especially for immutable actions if desired. I did verify these SHA and repository combinations manually before submitting this PR.

The `github/codeql-action` does have immutable actions (tags cannot have their SHA reset after publication) so this can be swapped readily to the tag without issue. The other two do not currently have released immutable actions. Tooling from `zizmor` can be addedto attempt auto-detection of fraudulent commits in those other cases.

classabbyamp: I leave that up to the maintainers, I just wanted to point that out

.github/workflows/build.yml
```diff
@@ -29,46 +29,48 @@ jobs:
...
+        run: cargo clippy --locked --profile ci --workspace --all-targets -- -D warnings
 
-    - uses: taiki-e/install-action@v2
-      with:
-        tool: cargo-nextest
-    - name: Test
-      run: cargo nextest run --locked --profile ci --workspace --all-targets
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-nextest
```
casperstorm: Should we SHA-pin this as well, or do tag?

.github/workflows/build.yml
```diff
@@ -29,46 +29,48 @@ jobs:
...
+            git \
+            pkg-config \
+            mold \
+            clang \
+            libdbus-1-dev \
+            libudev-dev \
+            libxkbcommon-dev \
+            libfontconfig1-dev \
+            libasound2-dev \
+            libx11-xcb-dev
```
casperstorm: We have near identicial apt package installs both here; in release.yml and security.yml. Perhaps we could consolidate to make it easier to maintain?

.github/workflows/build.yml
casperstorm: I am good with either here, as long as we keep it consistent. Personally i lead towards SHA.

KaiKorla: I would recommend pinning to a commit SHA. It is safer than using `latest` or a tag, because those references are mutable. This prevents most common supply-chain attacks. That said, SHA pinning only guarantees the exact Git commit — it does not secure the entire supply chain.
The attack vector mentioned in the thread is real, but it can be mitigated by verifying that the referenced commit is actually part of the repository’s visible history. Trust is good, but verification is better.

.github/workflows/build.yml
KaiKorla: Actions should be verified and have SHA-pin anyway.

.github/workflows/build.yml
melocene: @casperstorm @KaiKorla I went ahead and pushed a commit pinning all actions to SHAs for the exact version already specified. No additional updates were performed at this time. Let me know if you would like this applied, or wait for dependabot to take care of it.

melocene:
While swapping all Actions to SHAs I noticed the following:

- **peaceiris/actions-mdbook@v1** — Last released Oct 2022, stuck on Node 16. This can be swapped to `taiki-e/install-action` similar to what was done with `cargo edit` to install `mdbook`.
- **peaceiris/actions-gh-pages@v3** - This has an available update to a more supported version, or perhaps using the official `actions/deploy-pages` may be a good alternative.
- **jmgilman/actions-generate-checksum@v1** - Has not had commits in years, nor a release. It also uses Docker for Actions which may or may not remain supported. A simpler alternative could be just using the following:

```
- name: Generate checksum
  run: sha256sum h* > checksums.txt
```

.github/workflows/build.yml
melocene: https://github.com/squidowl/halloy/pull/1600/changes/78cf6a713780866cad15b95767055510ac632d47 Should now consolidate the shared list of packages while maintaining the individuals needed by the `build` workflow.

scripts/build-windows.sh
```diff
@@ -7,7 +7,6 @@ TARGET="x86_64-pc-windows-msvc"
 HALLOY_VERSION=$(grep -q '\..*\.' VERSION && cat VERSION || echo "$(cat VERSION).0")
 
 # update package version on Cargo.toml
-cargo install cargo-edit
```
KaiKorla, line 10: cargo-edit is needed for cargo set-version:
- https://crates.io/crates/cargo-edit
- https://crates.io/crates/cargo-edit#cargo-set-version

melocene: Correct, this should be installed in the runner by this section: https://github.com/squidowl/halloy/pull/1600/changes/BASE..f7376c2814a157b2cc026cf1ed2779c81d16c0ee#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R56-R60

The line in the script compiles and installs `cargo-edit` where the new step downloads a precompiled binary for use. The MacOS and Linux versions do not use this same mechanism so `cargo edit` is not installed for those OS.

KaiKorla: Hmm, ok. Then i have to keep this in mind, when i testing the windows build and windows installer on local. 👍🏻

melocene: Good call out. I just pushed a small change to the script. In Actions, or if a user already has `cargo-edit` installed then no install happens. Otherwise it will go ahead and install the binary like it did before. How does that sound to you @KaiKorla as a good middle ground?

.github/workflows/build.yml
casperstorm: Sounds good.

casperstorm:
Unsure if we can do anything about the warnings `cargo audit` throws at this moment. 
Should we `cargo audit --ignore` them or what is the best action to take?

KaiKorla:
> Unsure if we can do anything about the warnings `cargo audit` throws at this moment. Should we `cargo audit --ignore` them or what is the best action to take?

There are possible drop-in replacements in the log. IMO give them a try.

melocene:
> > Unsure if we can do anything about the warnings `cargo audit` throws at this moment. Should we `cargo audit --ignore` them or what is the best action to take?
> 
> There are possible drop-in replacements in the log. IMO give them a try.

This is an option, especially if drop-in it would be minimal work to swap them out. Updating some dependencies may also help. If the update isn't semver compatible though it might need some investigation to integrate and get a successful build.

There is also the option of temporarily disabling that job and re-enable it later when the issues are resolved. Or leave it which won't block any PRs or merges if not made a required check. Just could look awkward for users who are not familiar with the context.

I didn't want to make this PR too intrusive up front which is why I did not perform any updates to dependencies or actions at this time. My plan was to follow this on with additional PRs to spread out the code and review work while chipping away at findings or features.

casperstorm (approved):
Looks good to me. Let's address the `cargo audit` in a follow up.

Cargo.toml
```diff
@@ -164,6 +164,13 @@ opt-level = 1
 [profile.dev.package."*"]
 opt-level = 3
 
+# Enable a large amount of optimizations and minimize filesize in release profile.
+[profile.release]
+lto = true
+panic = "abort"
+strip = true
+codegen-units = 1
+
```
andymandias: `release` profile is frequently used for debugging that doesn't need full `debug` profile capabilities.  PR #1418 (following up on suggestions from PR #367) introduced a `packaging` profile to allow the `release` profile to continue to serve that purpose.  From recent experience with `--all-features` being included in packages I'm not sure we can expect packagers to use a `packaging` profile, so moving the contents of the `packaging` profile settings from that PR to `release` here is probably the right call from that perspective. But, we should have a dev profile that's less overhead than `debug`.

melocene: This commit can be reverted if you would like. Looking at the profiles to me they seem to be reasonable as is. For release (distribution) builds this change results in the most performant and minimally sized binary. This profile is what I also use when building `halloy` for my daily use.  I am not totally clear on the use case of the `packaging` profile as I typically see `release` used for both packaging and end user consumption (also noted in the [Cargo documentation](https://doc.rust-lang.org/cargo/reference/profiles.html#release)) Likewise for using `release` for debugging without the symbols, but as noted I am fine to revert this commit either way.

andymandias: Let's revert this for now, I'll update PR #1418 with our requirements and we can follow this PR with it.

melocene: Reverted, and I can help keep an eye on https://github.com/squidowl/halloy/pull/1418

casperstorm:
@melocene do you wanna tackle the `cargo audit` that needs to be resolved? then i think it would make the most sense to make a new branch on top of this, and then when ready we merge both of them. This here is ready to go once `cargo audit` has been addressed.

melocene:
@casperstorm Looking at the results the check might not be able to turn green until `rsa` publishes a new release. There is [v0.10.6-rc.16](https://github.com/RustCrypto/RSA/releases/tag/v0.10.0-rc.16). Even with doing a `cargo upgrade && cargo update` to get all dependencies to the latest semver compatible version I got it down to 1 vulnerability and 4 warnings. Even when swapping some of the unmaintained dependencies as suggested by @KaiKorla only the warnings will go down, but `rsa` will still report as vulnerable, meaning the check will remain red.

If it isn't desired to have a red check then disabling the job temporarily, or excluding the finding would be the call. It is suggested this then turn into an issue to not lose track of a re-enable the job or finding later when fixed. Alternatively, since it is not required it won't block any merges and can serve as a reminder for keeping an eye on the dependencies for updates.

Also note: `bincode` is no longer maintained and will not be again by the original authors (Ref: https://crates.io/crates/bincode). This also means there will also be that warning until any crate using `bincode` replaces it. As this is a warning it will allow the check to remain green but will be present in the logs, and the workflow will open a single issue about the crate.

andymandias:
According to the arti folks [the RSA vulnerability doesn't impact them](https://gitlab.torproject.org/tpo/core/arti/-/issues/1141).  I believe they're the sole source of the `rsa` crate, so if we trust them the issue should not impact us.

casperstorm:
+1 for "ignoring" the `rsa`. We don't build tor feature by default as well.

melocene:
@casperstorm The exception was added for the `rsa` vulnerability. This PR looks ready to go now except for the compilation optimizations being worked on in another PR.

PR #1601, opened 2026-03-04
Include Relay Bytes in Message Length Verification

andymandias:
For messages that are expected to be relayed (`PRIVMSG` & `NOTICE`) the server will need to prepend `:nick!user@host ` to the message (and some servers will also add a trailing colon to messages even if there are no spaces in the message).  This includes those bytes in the calculation to determine whether a message is overlong.  From what I've found there are three mechanisms that I believe should be reliable for determining the user's username and hostname:
- `PRIVMSG`/`NOTICE` echoed via `echo-message` when it is available ("If servers apply any modifications to these messages, they MUST send the final version of the message back to the originating client.")
- `CHGHOST` messages, when `chghost` is available (a `CHGHOST` on connection cannot be expected)
- `RPL_NAMREPLY`, when `userhost-in-names` is available

From what I've observed `RPL_WHOREPLY`, `RPL_WHOSPCRPL`, and `RPL_WHOISUSER` may return non-cloaked values to the user and can't be relied upon.

So, this patch attempts to update the username/hostname as available, and falls back to ISUPPORT parameters (if available, if not then a conservative estimate is used).

I did some restructuring to try to consolidate capabilities management because we were starting to have far too many `supports_*` fields.

(Edit to note:  I consider this state of username/hostname detection well less than ideal, but it is the most reliable conglomeration I found.  And having messages bounce back truncated or denied is a poor experience.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ Added:
 
 - Setting to limit reaction text length (`buffer.channel.message.max_reaction_chars`)
 
+Fixed:
+
+- Improved message length calculation (when known the bytes required to relay a message are included in the calculation, when unknown an estimate is used)
+
+Thanks:
+
+- Bug reports: chmod222, @whitequark
+
 # 2026.4 (2026-03-03)
 
 Added:


data/src/client.rs
```diff
@@ -139,12 +139,234 @@ struct ChatHistoryRequest {
...
+
+        if self.pending.contains("soju.im/bouncer-networks")
+            && !self.acknowledged(Capability::BouncerNetworks)
+        {
+            requested.push("soju.im/bouncer-networks");
+        }
+
+        if self.pending.iter().any(|cap| cap.starts_with("sasl")) {
+            requested.push("sasl");
+        }
```
casperstorm: Should we gate with `!self.acknowledged(Capability::Sasl)` to avoid a unexpected re-auth? I am unsure if it can happen.

Issue #1557, opened 2026-02-25
Opening URL Schemes should detect current instance


sebbu2:
When clicking one of the link from [URL Schemes](https://halloy.chat/guides/url-schemes.html), it should detect if a halloy instance is already running or not instead of launching a new one.

Then it should detect if I'm already connected to the server or not : checking map or links or the server urls at the start of the connection before motd (I might be connected to another linked server).

Then it should detect if I'm already on the channel or not, join it if necessary, and focus on it.

KaiKorla:
~~This likely requires a decision by @casperstorm / @andymandias / @tarkah on whether this special behavior should be implemented.~~

~~**Reason:** The Windows URL scheme opens a new instance of a program each time. Currently, Halloy is not a single-instance application, and I assume this is unlikely to change. There are possible solutions, for example using IPC to hand over the URL to an already running instance. However, I’m not sure this is a good fit for a multi-instance application. Some users may prefer this behavior, while others may not.~~

~~In my opinion, the URL scheme is more of a convenience feature than a core one. Given that, it might be worth considering how often it would actually be used before investing significant effort into a more complex solution.~~

KaiKorla:
@sebbu2 It seems we already have this IPC mechanism — I wasn't aware of it. 🙂

I tested it on my Windows machine with Halloy 2026.4 and it worked flawlessly, so I can't reproduce the issue at the moment. Could you provide some more context (OS, OS version, Halloy version, etc.)?

sebbu2:
Windows 11 x64, halloy 2026.3 and 2026.4.
Chrome last version, public release.
[screenshot](https://x0.at/PLyP.png). I specifically set autoconnect to false in my config for the new test.

PR #1619, opened 2026-03-06
fix(ipc): handle IRC URL routes in running instance

KaiKorla:
Support for #1557

Reuse active IRC connections when processing IPC URL events by preferring
ISUPPORT NETWORK matching and falling back to hostname matching. If no active match exists, create a new connection from the URL host/port, open or join the target channel, and focus its buffer without duplicates.

@andymandias & @casperstorm 
As always: feedback welcome. Let me know if this is a proper implementation. tyvm


assets/networks/networks.toml
```diff
@@ -0,0 +1,29 @@
...
+
+[[networks]]
+network = ["Libera.Chat"]
+hosts = [
+    "irc.libera.chat",
+    "irc.au.libera.chat",
+    "irc.eu.libera.chat",
+    "irc.ea.libera.chat",
+    "irc.ipv4.libera.chat",
+    "irc.ipv6.libera.chat"
```
sebbu2: nicely forgotten comma 😁

PR #1621, opened 2026-03-08
More prominent searchbar in Book

englut:
I've been seeing a bunch of questions from users about configuration options, only to be surprised there's a search feature on the Book!

I've taken it upon myself to create a (albeit hacky) method of forcing the searchbar to be:

1. always open
2. moved to the top of the sidebar

Here's a screen recording of me playing with it:

[halloy-book-searchbar.webm](https://github.com/user-attachments/assets/36f0fb70-8c89-4727-b7b1-1d5bec9d29ba)

To closely mimic production, I tested via `mdbook build && serve book/`. The video is made on the built static files rather than `mdbook serve` :pray:

### DIFF
--- a/book/scripts/move-searchbar.js
+++ b/book/scripts/move-searchbar.js
@@ -4,30 +4,33 @@ window.addEventListener("load", () => {
   const searchNode = document.querySelector("#mdbook-search-wrapper");
   const searchToggleNode = document.querySelector("#mdbook-search-toggle");
 
+  // move search bar to the sidebar
+  if (sidebarNode && searchNode) {
+    sidebarNode.prepend(searchNode);
+  }
+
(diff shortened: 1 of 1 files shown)


englut:
Alternatively, if we don't want to mess about too much with the sidebar of mdbook, I could just force the search bar to be always displayed!

Screenshot_20260309_135501

casperstorm:
Thanks @englut. I am currently researching if there is any other good projects instead of mdbook. I feel like something like this should be a default feature 🙃

casperstorm:
@englut https://github.com/squidowl/halloy/pull/1664

englut:
> @englut #1664

Nice!! I like that a lot more. Will close this, thanks for addressing this @casperstorm :pray:

PR #1631, opened 2026-03-09
Sidebar channel & nickname sorting

englut:
Added a new `"config"` option to `sidebar.order_channels_by`. Allows the channels to be sorted by user's config.
```toml
[sidebar]
order_channels_by = "config"

# Example: When using "config", channels appear in this exact order:
[servers.liberachat]
channels = ["#rust", "#halloy", "#halloy-test"]
# Result: #rust → #halloy → #halloy-test → (any other channels are sorted by "name")
```

This also allows custom user sorting of the bouncernetwork channels.

Other fixes:

- I added a case insensitive sort for query names in the sidebar.
- Configuration reload now works for sidebar channel sorting.

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -290,6 +290,7 @@ impl Client {
         from_modal: bool,
     ) -> Vec<Event> {
         if self.registration_step == RegistrationStep::Complete {
+            // TODO: allow from_modal when modal matching to bouncer networks is added.
             if !self.server.is_bouncer_network() {
                 self.join(
                     &config


src/screen/dashboard/sidebar.rs
```diff
@@ -557,7 +570,9 @@ impl Sidebar {
                             }
 
                             // Queries from the connected server.
-                            let queries = history.get_unique_queries(server);
+                            let mut queries =
+                                history.get_unique_queries(server);
+                            queries.sort_by_key(|q| q.as_str().to_lowercase());
```
andymandias: We should use `as_normalized_str()` here instead of `as_str().to_lowercase()`.  And in `src/screen/dashboard.rs`.

Would be nice to pre-sort this as well, but I think the added complication will not be worth it at the moment.

data/src/client.rs
```diff
@@ -1786,19 +1786,7 @@ impl Client {
...
-                        |c1, _, c2, _| {
-                            compare_channels(
-                                &chantypes,
-                                config.sidebar.order_channels_by,
-                                c1.as_normalized_str(),
-                                c2.as_normalized_str(),
-                            )
-                        },
-                    );
+                    self.chanmap.entry(target_channel.clone()).or_default();
```
andymandias: The reason this was originally sorted here, rather than in `view` functions, is because `view` functions are called very frequently.  We could probably get away with it in this case because the sorting is going to be relatively small, but unless there's a reason we should sort the values beforehand.

That does mean we will have to re-sort in `update_config` (which I believe we neglect to do currently).

englut: For sure. I attempted to go down this route at first, and attempt to resort in `update_config`, but realized how far up the chain I'd have to pass config::Config instead of just config::Server.

I can return to this method if you think that is better. I was worried that the view is too frequently updated as well. It works pretty well when I run it locally, but I only have about 20 or so buffers open, maybe someone with a larger amount of buffers could experience performance issues.

andymandias: How about we have an optional setting inside `config::Server`, and the setting in `config::Config` is the default?  It's a bit more complicated, but will get us our config where we want and provide more customization to users.

englut: You mean to copy the config `order_channels_by` into `config.server.<server_name>` as well? Ah right, if we default it to the value of `config.sidebar`... That's a great idea! I will try that out :saluting_face:

andymandias: We do a similar thing with `proxy`, if that helps.  There's a default one in `config::Config` and a server specific one in `config::Server`.

englut: Okay! I managed to figure it out! This works even better now! Thanks @andymandias :saluting_face:

src/screen/dashboard/sidebar.rs
englut: Updated :pray: 

I looked into the pre-sorting of this as well.. and you're right, I think that maybe for now, centralizing it into `history.get_unique_queries()` is okay. Will need for sorting to happen on fetch I think.

englut: Ah yeah. Looking at this more... Would need to refactor the HashMap usage in `data/src/history/manager.rs`.. which is used by more than just queries. Would probably be too complicated right now.

data/src/config/server.rs
```diff
@@ -63,6 +63,11 @@ pub struct Server {
     pub channels: Vec<String>,
     /// A mapping of channel names to keys for join-on-connect.
     pub channel_keys: HashMap<String, String>,
+    /// Order server's channels
+    pub order_channels_by: Option<config::sidebar::OrderChannelsBy>,
+    /// Channel list used for sorting (preserved for bouncer networks)
+    #[serde(skip)]
+    pub sort_channels: Vec<String>,
```
andymandias: Generally we want to keep data structures that aren't settings themselves out of the config data structures unless we can help it.  I think in this case we can avoid doing so.

data/src/client.rs
```diff
@@ -379,6 +379,27 @@ impl Client {
...
+                    |c1, _, c2, _| {
+                        compare_channels(
+                            &config,
+                            &chantypes,
+                            c1.as_normalized_str(),
+                            c2.as_normalized_str(),
+                        )
+                    },
+                );
+            }
```
andymandias: Can we use `IndexMap::sort_by` here, instead of `drain`→`insert_sorted_by`?

data/src/client.rs
```diff
@@ -3543,6 +3550,57 @@ fn compare_channels(
...
+                &config.sort_channels
+            } else {
+                &config.channels
+            };
+            let a_pos = channel_list
+                .iter()
+                .position(|ch| ch.eq_ignore_ascii_case(a));
+            let b_pos = channel_list
+                .iter()
+                .position(|ch| ch.eq_ignore_ascii_case(b));
```
andymandias: We should compare normalized values here.  I think ideally that means that our reference ordering of channels is already normalized.

data/src/client.rs
andymandias, line 2337: Since we're messing with channel sorting, we should probably implement the channel re-sort here when casemapping changes (just as you did in `update_config`).

data/src/config.rs
```diff
@@ -431,7 +431,8 @@ impl Config {
         })
         .map_err(|e| Error::Parse(e.to_string()))?;
 
-        let servers = ServerMap::new(servers).await?;
+        let servers =
+            ServerMap::new(servers, sidebar.order_channels_by).await?;
```
andymandias, line 436: If we do the override here, when entering the `server::Config` into `ServerMap`, then we lose the knowledge of whether the server's `order_channels_by` is the result of being set explicitly or by adopting the default.  I think that's okay information to lose in this case, just something to consider.

data/src/config/server.rs
```diff
@@ -172,9 +177,11 @@ impl Server {
             nick_password_command: Option::default(),
             nick_identify_syntax: Option::default(),
 
-            // channels not relevant
+            // channels not relevant (don't auto-join on bouncer networks)
             channels: Vec::default(),
             channel_keys: HashMap::default(),
+            // but preserve parent's channel list for sorting
+            sort_channels: self.channels.clone(),
```
andymandias: If we don't overwrite channels here, and instead prevent auto-joining inside `data/src/client.rs`, then I think we can use `channels` for sorting in all cases.

data/src/client.rs
andymandias, line 2822: If we're dropping `sort_channels` for `channels`, then we need to make this conditional on `!self.server.is_bouncer_network()`.

data/src/config.rs
englut: for sure. I tried to copy the proxy config to have it fallback to the global if the server isn't available. but I couldn't quite figure out how to do that without doing what I've done here. I think in either case, we'd lose the knowledge of where that value came from explicitly.

data/src/client.rs
englut: ahhh for some reason I thought I had to rebuild the chanmap here in order to resort... but I guess all that matters is the sorting! It's a lot simpler now!

data/src/config/server.rs
englut: Removed! Made possible due to the `is_bouncer_network()` check :pray:

data/src/client.rs
englut: Right. can just do a direct compare since it's already normalized. :saluting_face:

data/src/config/server.rs
englut: Removed the `sort_channels` and also stopped clearing out `channels` as well :pray: made possible due to the `is_bouncer_network()` check :pray:

data/src/client.rs
englut: Nice! This was something I couldn't find myself before. So thank you for this! So much cleaner than `sort_channels`!

englut: This one was tricky for me. I can sort the `self.chanmap`, but logically... I'd just be sorting the the old casemapping right? So I added some logic to rebuild the `self.chanmap`. I think this is right? But maybe you could look at what I've done and see if it's what you were expecting. I also did the same thing to the `self.resolved_queries` as well.

englut: will replace `self.resolved_queries` with this when `feat/typing` is merged:

```rust
    self.querymap = take(&mut self.querymap)
        .into_iter()
        .map(|(mut query, state)| {
            query.renormalize(casemapping);
            (query, state)
        })
        .collect();
```

data/src/client.rs
```diff
@@ -2222,6 +2237,39 @@ impl Client {
...
+                                                self.chanmap = take(&mut self.chanmap)
+                                                    .into_iter()
+                                                    .map(|(channel, data)| {
+                                                        let updated_channel = target::Channel::from_str(
+                                                            channel.as_str(),
+                                                            &chantypes,
+                                                            casemapping,
+                                                        );
+                                                        (updated_channel, data)
+                                                    })
```
andymandias: This, and the `target::Query::parse` that follows, *should* be okay.  That is, they should never mangle the existing queries.  But I think this is an opportunity to introduce a bit of functionality to `Target`/`target::Channel`/`target::Query`: a `renormalize` function which would work exactly as `Nick::renormalize` does in `data/src/user.rs`.

data/src/config.rs
andymandias: In case you're curious the proxy configuration handling I'm was attempting to refer to is in `data/src/stream.rs`, where the global proxy is a separate argument provided in addition to the `server::Entry`/`server::Config`.  There is a minor utility to that since we avoid spinning up a `reqwest::Client` if the server is using a global proxy (since it can use the global proxy's `reqwest::Client` in that case).  All that is not necessary in this case.  I think the *only* thing it'd gain us here is to be able to do something like
```rust
let order_channels_by = self.config.order_channels_by.unwrap_or(default_order_channels_by);
```
instead of
```rust
if let Some(order_channels_by) = self.config.order_channels_by
```
with the implicit assumption that `self.config.order_channels_by` will never be `None`.  Not worth the hassle in my opinion.

data/src/client.rs
```diff
@@ -3543,6 +3579,46 @@ fn compare_channels(
...
+         * will be in different sections of the list.
+         */
+        Some(config::sidebar::OrderChannelsBy::NameAndPrefix) => a.cmp(b),
+        /*
+         * If config.sidebar.order_channels_by is `config`, this will sort channels based on your
+         * config.server.<server_name>.channels. Anything not in this list is sorted by `name` (default).
+         */
+        Some(config::sidebar::OrderChannelsBy::Config) => {
+            let a_pos = &config.channels.iter().position(|ch| ch == a);
+            let b_pos = &config.channels.iter().position(|ch| ch == b);
```
andymandias: I forgot that the `config` field is an `Arc`, so I don't think we want to try to normalize `config.channels` in place (plus, that would mean we lose the values as the user typed them). Let's pass the server's `isupport::CaseMap` in here and normalize as part of this comparison.  Channel comparison happen infrequently enough I think we don't need to store a separate, normalized copy.

data/src/history/manager.rs
```diff
@@ -507,7 +507,8 @@ impl Manager {
                 .then_some(query),
                 _ => None,
             })
-            .collect::<Vec<_>>()
+            .sorted_by(|a, b| a.as_normalized_str().cmp(b.as_normalized_str()))
```
andymandias: `Ord` is defined for `target::Query`, so let's use that.

data/src/client.rs
englut: Gotcha! I think this makes sense to me now. I have added renormalize to both targets and updated the logic here to use the new functions.

data/src/client.rs
englut: ahhhhh yes that's a good point! Updated :saluting_face:

data/src/history/manager.rs
englut: Duh. That's what that is for. I was staring at it wondering if it was... I should ask more questions :sweat_smile:

data/src/config.rs
englut: Ahhh I think you're right. But I did attempt to clean this up a little more. :pray:

data/src/client.rs
```diff
@@ -2311,9 +2332,34 @@ impl Client {
...
+                                                    })
+                                                    .collect();
+
+                                                self.querymap = take(&mut self.querymap)
+                                                    .into_iter()
+                                                    .map(|(mut query, state)| {
+                                                        query.renormalize(casemapping);
+                                                        (query, state)
+                                                    })
+                                                    .collect();
```
andymandias, line 2363: Do we want to sort these?  I don't really expect the ordering to change, but I think that might be the more thorough approach.

data/src/client.rs
```diff
@@ -289,22 +290,24 @@ impl Client {
...
-                            self.chantypes(),
-                            self.statusmsg(),
-                            self.casemapping(),
-                        )
-                        .ok()
-                    })
-                    .filter(|channel| !self.chanmap.contains_key(channel))
-                    .collect::<Vec<_>>(),
-            );
+            if !self.server.is_bouncer_network() {
```
andymandias, line 294: We may want to allow this for modals, when modal matching to bouncer networks is added.  Not now :)

data/src/client.rs
englut: I tried that initially actually! But because the `querymap` isn't actually being used to display the queries in the sidebar or anywhere else that I'm aware of, I decided it was unnecessary here.

data/src/client.rs
englut: Added a TODO comment here for that future scenario!

Issue #1624, opened 2026-03-08
chat history doesn't load


e00E:
commit fc82be0eb3a1b06a8f0e80593df8f08bfba2a763, Arch Linux, Sway, Soju

config:

```
[buffer.server_messages.topic]
enabled = false

[buffer.chathistory]
max_messages = 0
```

Reproduction:
- open halloy
- halloy opens with no channel buffer open
- click on a channel buffer
- it opens and shows at least a page of message history
- close channel buffer
- open channel buffer again
- there is no history, have to click "Request Older Chat History" button

The history should show by default at the last step. I shouldn't have to click the button. Just like how it shows by default when I open the channel buffer for the first time after halloy launches.

casperstorm:
@melocene, @andymandias perhaps you two can take a look at this.

melocene:
This seems to be working as intended to me and not a bug per se.

When you first connect to the bouncer you get a history from the bouncer on the first load of Halloy. This is a server provided history not local disk. The history from the server is saved in memory until the buffer is closed, at which point no messages are persisted to disk due to `max_messages` being zero. From then opening the buffer will attempt to load from disk rather than the server since you are technically already connected. Since no history was persisted to disk, `max_messages = 0`, no history is going to show up automatically. Clicking the button seems like that may just refetch the history from the server and stores it back in memory which is why it repopulates manually when you click the button.

`max_messages` only prevent history from being written to disk locally, it won't control if anything the server/bouncer sends to the client, which the client may then store in memory.

e00E:
Could we have more settings around that then? How long did halloy have the on disk history? Previously halloy had the behavior I want. But maybe that was already using the disk cache? I only turned off the disk cache because it makes halloy close very slowly. I see the following approaches:

- Setting to auto request history on buffer open.
- Setting to have an in memory chat history buffer instead of on disk.

These approaches are mutually inclusive. We can have both. I don't want halloy writing history to disk because my bouncer already does that. But I wouldn't mind having the most recent 10 megabytes (configurable) of history in RAM.

e00E:
> When you first connect to the bouncer you get a history from the bouncer on the first load of Halloy. This is a server provided history not local disk. The history from the server is saved in memory until the buffer is closed

[...]

Issue #1623, opened 2026-03-08
halloy is slow to close


e00E:
Arch Linux, Halloy package, halloy 2026.4 (5cc0677d), Sway

Closing halloy by pressing ctrl+q or using the desktop environment close hotkey, makes halloy freeze for around five seconds consuming 100% CPU before it closes.

Still happens with current commit and `cargo run --release`, but it closes in 2 seconds instead 5.

englut:
@e00E this should be fixed by #1614 and will be in the upcoming release 🙏

casperstorm:
Correct! Will leave it open until we release so others can see.

e00E:
I see. I still find halloy slow to close even with the current commit, which includes the commit you link. It is much faster than at the current release, but it is still noticeably slow.

Looking at the commit, this is related to compressing history files in `~/.local/share/halloy`. Why does halloy keep these history files when I use a bouncer (soju) with chathistory support?

casperstorm:
> I see. I still find halloy slow to close even with the current commit, which includes the commit you link.
> 
> Looking at the commit, this is related to compressing history files in `~/.local/share/halloy`. Why does halloy keep these history files when I use a bouncer with chathistory support?

You can disable it

```toml
[buffer.chathistory]
max_messages = 0
```

e00E:
Undocumented? https://halloy.chat/configuration/buffer/chat-history/index.html

e00E:
Ah, I see the setting is new and not released yet.

e00E:
I can confirm that with current commit and `max_messages=0`, halloy closes quickly.

casperstorm:
After discussing with @andymandias, we’ve decided to revert https://github.com/squidowl/halloy/commit/7bd0b4ee84d3acdfe89f9363ca51cceeaf810efe (the PR that added max_messages). We’ll revisit it once we have better tools to handle 0 and very high values. Eg, when we can implement a proper “thin soju client”. However, that’s not something we’re ready for right now :)

melocene:
@casperstorm @e00E I opened https://github.com/squidowl/halloy/pull/1632 which should help here without the side effects of `max_messages` being set either low or to high values.

PR #1632, opened 2026-03-10
Add chathistory.persist option for writing history to disk

### LINKED ISSUE #1491: How to disable history in %appdata%/halloy/history?

melocene:
As this feels like an important privacy feature, and after reading some other issues https://github.com/squidowl/halloy/pull/1520 may have been overengineered and caused more issues. This is a different approach, only add a toggle for writing to disk, don't touch `MAX_MESSAGES` or other buffer related code. It should still resolve #1491 and https://github.com/squidowl/halloy/issues/1624 and possibly partially resolve https://github.com/squidowl/halloy/issues/1623 if the user is already using a bouncer for history. I tested this on Windows and the new toggle does prevent the history writing without other side effects (that I can tell from my minimal use case).

If a user already has history files before `persist = false` is set it will not currently remove that history. Those files will still be read and loaded as normal and already handled by existing code. At worst the user will have an empty `history` directory created normally by `halloy` but this is not a new change with this PR.

@e00E Are you also able to test this branch to see if you still get a similar issue to #1624 please?

### DIFF
--- a/data/src/history.rs
+++ b/data/src/history.rs
@@ -633,9 +633,11 @@ impl History {
                     *last_updated_at = None;
 
                     if !persist {
+                        let messages = messages.clone();
                         return Some(
                             async move {
-                                metadata::save(&kind, &[], read_marker).await
+                                metadata::save(&kind, &messages, read_marker)
+                                    .await
                             }
                             .boxed(),
                         );


e00E:
Correctly doesn't store history on disk and makes halloy close quickly. Doesn't resolve #1624 . I still have to click Load More Messages manually the second time I open a channel as described in the issue.

melocene:
@e00E That makes sense since I didn't change the buffer behavior as that would be a separate issue/PR in my opinion since halloy is just a client. The buffer messages are tightly coupled to the history on disk and does not rely on the server as a summary. Glad to hear it solved the main issue though of close speed and not duplicating the cache.

I wanted to specifically focus this on the privacy aspect of not storing local history, which also helps avoid issues when writing the history before closing halloy. As noted in #1624 this only controls the on disk history not what the server is sending nor what the client requests. A couple proposed solutions are in #1624 for the issue of not refreshing from the server when opening a buffer.

> When you first connect you have no buffers and no history loaded from disk. When the connection finishes to your bouncer you get the latest messages from the bouncer by way of [load_metadata()](https://github.com/squidowl/halloy/blob/main/src/screen/dashboard.rs#L3063-L3086) which is called from [Event::JoinedChannel](https://github.com/squidowl/halloy/blob/main/src/main.rs#L1442-L1453). Joining a channel is separate from closing or re-opening a buffer.

melocene:
@e00E I think the newest commit I added should now do what you need, try to request the history from the server instead of the files if the local disk history is turned off.

cc @casperstorm

e00E:
Yes, it works. Thanks!

casperstorm (changes requested):
missclick

melocene:
@casperstorm Do you have a preference which way you want this to go? The second option of the docs update seems most straightforward to me and quickest to implement.

casperstorm:
> @casperstorm Do you have a preference which way you want this to go? The second option of the docs update seems most straightforward to me and quickest to implement.

Nevermind my previous review, i pulled them. This look good.

melocene:
No problem. Let me know if anything else is needed here to make sure this is complete and achieves the behavior you all expect in halloy.

melocene:
@casperstorm You actually had a good point in the removed comment about the documentation so I went ahead and fixed that. Thank you for calling out that I missed it.

As far as the other portion about the metadata info being written I might need a hand resolving that if still applicable and possible. Edit: I took at a shot as fixing both things, and ensuring metadata is written for better UX. Feel free to edit the commits if needed in case I missed something or it wasn't complete.

melocene:
@andymandias Thanks for the thorough response. Would it be possible for you to open an issue regarding the history manager so I can follow along please and help where able? As noted I am willing and able to help with this but without additional guidance or documented roadmap its a bit difficult to do so.

I also do not use a bouncer and have been running this patch for awhile without noticing any issues. For those using a bouncer I did consider the network traffic issue and this was the compromise I landed on after #1520 was reverted, and considering the notes by @casperstorm in this and the other PR.

I also made sure that by default this setting was set to false so there were no default behavior changes. The was fully opt-in for users who want it, and those who don't wouldn't have anything changed for them. Not too sure if I made that part clear in the original comment.

PR #1637, opened 2026-03-11
feat/typing

### LINKED ISSUE #1539: Support ircv3 `typing` client tag
With the latest updates to libera.chat [supporting typing indicators](https://libera.chat/news/new-and-upcoming-features-3), and since halloy's support channel lives on libera, it'd be good to support [ircv3's `typing` client tag](https://ircv3.net/specs/client-tags/typing) spec at some point.

~~Of course not all of libera's servers have been upgraded to the latest version that supports this, but I have found these server(s) do (will add to the list as I find more):~~

- ~~`calcium.libera.chat`~~
- ~~`iridium.libera.chat`~~
- ~~`osmium.libera.chat`~~
- ~~`uranium.libera.chat`~~
- ~~`molybdenum.libera.chat`~~

I imagine implementing this will require tracking keyboard events. I haven't touched the irc spec support parts of the codebase yet, but I THINK we could take advantage of iced's [keyboard event subscription](https://docs.rs/iced/latest/iced/keyboard/fn.listen.html) to do this.

I think it would be cool if a little box with "<nick> is typing..." (single person) or "A few people are typing... " (multiple people) pops up above the text input field when it's supported. But I'm no designer 😂

Edit: I can't confirm, but so far I'm seeing most of the servers on libera.chat have been updated to a version that supports `+typing` 🙏

casperstorm:
this adds typing support for networks which supports it. fixes #1539.

image

API is as follows

```toml
[buffer.channel.typing]
share = false # Control whether Halloy shares your typing status with other users.
show = true # Control whether Halloy shows typing status from other users.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ Thanks:
 
 - Contributions: @KaiKorla, @englut
 - Bug reports: chmod222, @whitequark, @englut, sebbu, @ascarion
-- Feature requests: @gAlleb, @jtbx, @cyrneko
+- Feature requests: @gAlleb, @jtbx, @cyrneko, @englut
 
 # 2026.4 (2026-03-03)
 


src/buffer/query.rs
```diff
@@ -127,7 +146,25 @@ pub fn view<'a>(
...
+    let scrollable: Element<'a, Message> = if let Some(typing) = typing {
+        let typing_overlay: Element<'a, Message> = container(typing)
+            .width(Length::Fill)
+            .height(Length::Fill)
+            .padding(padding::left(2))
+            .align_y(iced::alignment::Vertical::Bottom)
+            .into();
+
+        column![
+            stack![content, typing_overlay].height(Length::Fill),
```
casperstorm, line 160: using a `stack![]` so that the scrollable of messages slides behind the typing view, similar to Discord.

src/buffer/scroll_view.rs
```diff
@@ -669,6 +693,11 @@ pub fn view<'a>(
             bottom_spacer,
             space::vertical().height(line_spacing),
         ]
+        .padding(if reserve_bottom_line_for_typing {
+            padding::bottom(theme::resolve_line_height(&config.font) + 2.0 /* for padding */)
```
casperstorm: Need to reserve some spacing below the messages in order to show the typing, without collision - however i add a bit (`2.0`) to have enough for the padding ive added to the typing view.

data/src/client.rs
```diff
@@ -150,6 +151,7 @@ pub struct Client {
     resolved_host: Option<String>,
     configured_nick: Nick,
     chanmap: IndexMap<target::Channel, Channel>,
+    querymap: IndexMap<target::Query, QueryState>,
     resolved_queries: HashSet<target::Query>,
```
andymandias: Now that we have `querymap` (`quermap`? `query_map`?), would it make sense to have it take over for resolved_queries?  The latter was created to be able to provide a canonical non-normalized casemapping for nicknames; such that e.g. `FooBar` wouldn't show up as foobar.  I believe that remains its only purpose.

I *think* all we'd need is to add an entry to `querymap` for chathistory messages (if one does not already exist), and then utilize `querymap` in `resolve_query`.

data/src/client.rs
```diff
@@ -3470,6 +3653,41 @@ impl Client {
...
+                }
+                isupport::ClientOnlyTags::Denied(name) if name == "typing" => {
+                    (has_deny_all, typing_allowed, true)
+                }
+                _ => (has_deny_all, typing_allowed, typing_denied),
+            },
+        );
+
+        typing_denied || (has_deny_all && !typing_allowed)
+    }
```
andymandias: Very clean!  I think this would be a nice function to have as a generic in `isupport` (where you can pass in a `str` of the tag you're checking), similar to functions like `get_statusmsg_or_default` that are there now.

src/buffer/input_view.rs
```diff
@@ -2037,4 +2148,168 @@ impl State {
...
+                Some(buffer),
+                clients.nickname(buffer.server()),
+                clients.get_server_is_connected(buffer.server()),
+                &clients.get_isupport(buffer.server()),
+                config,
+            ),
+            Err(command::Error::MissingSlash)
+                | Ok(data::Command::Irc(command::Irc::Me(_, _)))
+        )
+    }
```
andymandias, line 2240: I think we want to cover `/format` and `/plain` as well, which get converted to a `command::Irc::Msg`.  If I'm right about that, then the criteria matches that of `multiline_bytes` returning `Some((_, MultilineBatchKind::PRIVMSG)`.  Ideally I think we should utilize the same logic for both;  either make a wrapper around `multiline_bytes` for this purpose, or a new function that `multiline_bytes` will then use.  That does mean we'll want to be sure to only call `maybe_send_typing_status` after `self.parse_lines` has been called (so it can used the cached `self.parsed`);  I think that means only the call inside `text_editor::Action::Edit(_)` would stay.  I believe that would simplify things overall, but would want to verify it covers all cases we want covered.

data/src/client.rs
casperstorm: I like this.

PR #1644, opened 2026-03-13
Add reactions

### LINKED ISSUE #1594: UI for selecting reactions
parent issue: #550 

Halloy can display reactions and send ones that have already been received, but there is currently no way to send the first reaction or select them.

This issue is for UI design ideas as well as tracking the feature.

casperstorm:
Add reactions.

Fixes #1594 and #550.

image

### DIFF
--- a/src/screen/dashboard/modal/reaction.rs
+++ b/src/screen/dashboard/modal/reaction.rs
@@ -7,9 +7,8 @@ use iced::widget::{
     Scrollable, button, column, container, scrollable, text_input,
 };
 
-use crate::emoji;
 use crate::widget::{Element, Row, text};
-use crate::{theme, widget};
+use crate::{emoji, theme, widget};
 
 const MODAL_WIDTH: f32 = 380.0;
 const MODAL_HEIGHT: f32 = 250.0;
--- a/src/buffer/input_view/completion.rs
+++ b/src/buffer/input_view/completion.rs
@@ -5,10 +5,6 @@ use std::ops::RangeInclusive;
 use std::sync::LazyLock;
 use std::{fmt, iter};
 
-use crate::emoji;
-use crate::font;
-use crate::theme::{self, Theme};
-use crate::widget::{Element, double_pass, text};
 use chrono::{DateTime, Utc};
 use const_format::concatcp;
 use data::buffer::SkinTone;
@@ -24,6 +20,10 @@ use iced::widget::text::Shaping;
 use iced::widget::{button, column, container, row, text_editor, tooltip};
 use irc::proto;
 use itertools::{Either, Itertools};
+
+use crate::theme::{self, Theme};
+use crate::widget::{Element, double_pass, text};
+use crate::{emoji, font};
 
 const MAX_SHOWN_COMMAND_ENTRIES: usize = 5;
 const MAX_SHOWN_EMOJI_ENTRIES: usize = 8;


src/screen/dashboard/modal/reaction.rs
```diff
@@ -0,0 +1,157 @@
...
+pub struct State {
+    msgid: message::Id,
+    selected_reactions: HashSet<String>,
+    search_query: String,
+}
+
+#[derive(Debug, Clone)]
+pub enum Message {
+    SearchChanged(String),
+    SelectEmoji(String),
```
4e554c4c: Should be a `Cow<'static, str>` since the emoji almost always has `'static` storage class

src/screen/dashboard/modal/reaction.rs
```diff
@@ -0,0 +1,157 @@
...
+}
+
+#[derive(Debug, Clone)]
+pub enum Event {
+    Toggle {
+        msgid: message::Id,
+        text: String,
```
4e554c4c: ditto

src/screen/dashboard/modal/reaction.rs
```diff
@@ -0,0 +1,157 @@
...
+            }
+        }
+    }
+}
+
+pub fn view<'a>(state: &'a State, config: &'a Config) -> Element<'a, Message> {
+    let query = normalized_query(&state.search_query);
+    let filtered = filtered_emojis(&query);
+    let has_results = !filtered.is_empty();
+    let grid = emoji_grid(&filtered, &state.selected_reactions);
```
4e554c4c, line 75: So when i was originally thinking that we could have multiple tabs in the modal

This would correspond to having tabs given by `emojis::Group::iter()` (the first tab being "any")

then each tab would have `group.emojis()` emojis

src/screen/dashboard/modal/reaction.rs
```diff
@@ -0,0 +1,157 @@
...
+    .width(Length::Fixed(EMOJI_BUTTON_WIDTH))
+    .height(Length::Fixed(EMOJI_BUTTON_HEIGHT))
+    .style(move |theme, status| {
+        theme::button::secondary(theme, status, selected)
+    })
+    .on_press(Message::SelectEmoji(emoji.as_str().to_owned()))
+}
+
+fn filtered_emojis(query: &str) -> Vec<&'static emojis::Emoji> {
+    emojis::iter()
```
4e554c4c: the `emoji` crate provides a nice emoji fuzzy search implementation.

i think for now we shouldn't bring in any extra dependencies though.

src/screen/dashboard/modal.rs
```diff
@@ -0,0 +1,52 @@
...
+}
+
+#[derive(Debug, Clone)]
+pub enum Message {
+    Reaction(reaction::Message),
+}
+
+#[derive(Debug, Clone)]
+pub enum Event {
+    Toggle {
```
4e554c4c: Maybe this should be `ToggleReaction`?
or `Reaction(Event::Reaction)`

data/src/client.rs
```diff
@@ -3667,6 +3667,22 @@ impl Client {
...
+                "draft/react",
+            )
+            && !isupport::is_client_only_tag_denied(
+                &self.isupport,
+                "draft/unreact",
+            )
+            && !isupport::is_client_only_tag_denied(
+                &self.isupport,
+                "draft/reply",
+            )
```
4e554c4c: reply is ratified, so we use both
```suggestion
            && (
                !isupport::is_client_only_tag_denied(
                    &self.isupport,
                    "draft/reply",
                )
                || !isupport::is_client_only_tag_denied(
                    &self.isupport,
                    "reply",
                )
            )
```

p.s. maybe it would make sense to have an `isupport::is_client_tag_allowed` instead

src/screen/dashboard/modal/reaction.rs
casperstorm: I thought about it, but i rather not introduce tabs just yet when we have a search-bar.
Tabs could then be introduced if/when we get custom reactions, etc.

src/screen/dashboard/modal/reaction.rs
andymandias: We do emoji search for the autocomplete in `Emojis::process` in `src/buffer/input_view/completion.rs`.  Would be good to consolidate I think, but that can be a follow-up.

PR #1616, opened 2026-03-06
Aliases

casperstorm:
This will be the first version of aliases. Fixes (for real this time) #372 and makes it possible to handle #1608.

Define custom aliases in config:

```toml
[buffer.commands.aliases]
op = "/mode #halloy +ooo $1 $2 $3"
halloy = "/me says halloy to $1!"
topic = "/topic #halloy $1-"
```

`$1` through `$9` insert individual arguments, and `$1-` through `$9-` insert everything from that argument to the end.

So for example:

`/halloy casper` expands to `/me says halloy to casper!`
`/topic hello world again` expands to `/topic #halloy hello world again`

PR #1653, opened 2026-03-14
Exec command

### LINKED ISSUE #1650: Feature request: `exec` command
Hi, I'm trying to switch from WeeChat to Halloy. One feature that I'm missing is the `exec` command, which lets me run shell commands and send their output to IRC. One use case for this is to send the currently playing song using `mpc current` or similar.

There are many ways to support something like this and I don't have strong opinions. The end goal for me would be to be able to define a short alias for a command that runs a given shell command, parses its output as an arbitrary Halloy command (e.g. `/me is now playing ...`) and runs it.

casperstorm:
Run a local shell command and send the first line of stdout to the current buffer.

```
/exec python -c "import random; print(f'/me rolled {random.randint(1, 6)}')"
=>
18:57 * casperstorm rolled 2
```

Fixes #1650

### DIFF
--- a/docs/.vitepress/config.mts
+++ b/docs/.vitepress/config.mts
@@ -73,6 +73,7 @@ export default defineConfig({
           { text: "Connect with soju", link: "/guides/connect-with-soju" },
           { text: "Connect with ZNC", link: "/guides/connect-with-znc" },
           { text: "Custom Themes", link: "/guides/custom-themes" },
+          { text: "Exec Command", link: "/guides/exec-command" },
           {
             text: "Example Server Configurations",
             link: "/guides/example-server-configurations",
--- a/docs/configuration/buffer.md
+++ b/docs/configuration/buffer.md
@@ -404,6 +404,8 @@ Configure `/exec`.
 `/exec` runs a local shell command on your machine. Enable it only if you trust the commands you plan to run.
 :::
 
+See the [Exec Command guide](../guides/exec-command.md) for a few simple examples.
+
 #### `enabled`
 
 Enable `/exec`.
--- a/docs/guides/exec-command.md
+++ b/docs/guides/exec-command.md
@@ -1,0 +1,23 @@
+# Exec Command
+
+The `/exec` command runs a local shell command on your machine and sends the first non-empty line of stdout to the current buffer.
+
+::: warning
+Enable `/exec` only if you trust the commands you plan to run. See [buffer command configuration](../configuration/buffer#exec).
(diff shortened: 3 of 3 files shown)


ncfavier:
This handles my use case perfectly, thanks.

WinnerWind:
this works great! but if i'mnot mistaken, one needs to run /exec <some command> all the time. Maybe a future PR will allow this to be defined in config, or does htis PR handle this usecase too?

andymandias:
> this works great! but if i'mnot mistaken, one needs to run /exec all the time. Maybe a future PR will allow this to be defined in config, or does htis PR handle this usecase too?

I may be misunderstanding your concern, but the recently merged #1616 will allow terse invocations (e.g. the example above could be aliased to `/roll`).

WinnerWind:
Ah yeah, that's what I was thinking of. Hopefully this can be merged soon!

src/buffer/input_view/exec.rs
```diff
@@ -0,0 +1,84 @@
...
+    let output = time::timeout(Duration::from_secs(timeout_secs), async move {
+        let mut process = if cfg!(target_os = "windows") {
+            let mut process = Command::new("cmd");
+            process.arg("/C").arg(command);
+            process
+        } else {
+            let mut process = Command::new("sh");
+            process.arg("-c").arg(command);
+            process
+        };
```
andymandias, line 21: Maybe the shell could be configurable?  In case users want to utilize `bash`/`fish`/`zsh`/etc features.  I guess users can always `bash -c` if they want.  Not necessary for this PR.

src/buffer/input_view.rs
```diff
@@ -1815,6 +1865,33 @@ impl State {
...
                         );
                     }
+                    command::Internal::Exec(command) => {
+                        if !config.buffer.commands.exec.enabled {
+                            self.error = Some(String::from(
+                                "exec is not enabled by the user",
+                            ));
+
+                            return (Task::none(), None);
+                        }
```
andymandias, line 1853: Should we catch this in `input::parse` validation, and prevent it from being "sent"?

src/buffer/input_view/exec.rs
casperstorm: Good idea. Lets do this in a follow up imo.

src/buffer/input_view.rs
casperstorm: God idea.

Issue #1654, opened 2026-03-15
Feature request: more informative window title


ncfavier:
Given that Halloy supports opening buffers in separate windows, I think it would make sense to set the window title to something more informative than "Halloy". Maybe this could be customisable, but a reasonable default would be "$focused_buffer_name – Halloy" (en dash).

PR #1657, opened 2026-03-15
Show focused buffer in window title

4e554c4c:
This commit displays the open buffer in the window title, or "Halloy" if none is selected. I was a little frustrated about the clones in `dashboard.rs` (see the triple clone in `open_buffer`) so i removed one clone and dropped a todo for the rest.

Resolves: #1654

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -402,8 +402,20 @@ impl Halloy {
         Task::none()
     }
 
-    fn title(&self, _window_id: window::Id) -> String {
-        String::from("Halloy")
+    fn title(&self, window_id: window::Id) -> String {
+        let default = "Halloy".to_owned();
+        let s = match &self.screen {
+            Screen::Dashboard(dashboard) => {
+                match dashboard.focused_buffer_name(window_id) {
+                    Some(s) => s,
+                    None => return default,
+                }
+            }
+            Screen::Help(_) => "Help".to_owned(),
+            Screen::Welcome(_) => "Welcome".to_owned(),
+            Screen::Exit { .. } => return default,
+        };
+        format!("{s} – Halloy")
     }
 
     fn update(&mut self, message: Message) -> Task<Message> {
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -402,8 +402,14 @@ impl Halloy {
...
-    fn title(&self, _window_id: window::Id) -> String {
-        String::from("Halloy")
+    fn title(&self, window_id: window::Id) -> String {
+        let s = match &self.screen {
+            Screen::Dashboard(dashboard) => dashboard.focused_buffer_name(window_id),
+            Screen::Help(_) => "Help".into(),
+            Screen::Welcome(_) => "Welcome".into(),
+            Screen::Exit { .. } => return "Exiting ...".to_owned(),
+        };
+        format!("{s} – Halloy")
```
ncfavier, line 418: Maybe display only "Halloy" if there is no focused buffer, or if we're exiting?

src/screen/dashboard.rs
```diff
@@ -4052,6 +4051,16 @@ impl Dashboard {
...
+
+    pub fn focused_buffer_name(&self, w: window::Id) -> Option<String> {
+        self.panes
+            .iter()
+            .filter(|(win, _, pane)| {
+                *win == w && !matches!(pane.buffer, Buffer::Empty)
+            })
+            .map(|(_, _, pane)| pane.buffer.to_string())
+            .next()
+    }
```
casperstorm, line 4115: Could we use `get_focused()`?

```rust
pub fn focused_buffer_name(&self, w: window::Id) -> Option<String> {
    self.get_focused().and_then(|(window, _, pane)| {
        (window == w && !matches!(pane.buffer, Buffer::Empty))
            .then(|| pane.buffer.to_string())
    })
}
```

4e554c4c: No. 
This is what i did originally, but only one window's PaneGrid is focused at a time. If you only look at the focused panes, you get a wrong answer for other windows.

casperstorm: Right that make sense! Just wanted to check in with you about it.

casperstorm:
Do you mind adding a entry to CHANGELOG, @4e554c4c.

PR #1660, opened 2026-03-16
Fix MonitoredOffline notifications using wrong config

### LINKED ISSUE #1659: MonitoredOffline notifications use monitored_online config instead of monitored_offline
I was reading through the notification handling code and noticed that the `Notification::MonitoredOffline` arm in `src/notification.rs` references `config.monitored_online` instead of `config.monitored_offline` in two places:

1. The `self.execute()` call passes `&config.monitored_online` — so the toast, sound, delay, and include/exclude filters all come from the *online* config
2. The `request_attention` return value also reads from `config.monitored_online.request_attention`

This means if a user configures `[notifications.monitored_offline]` differently from `[notifications.monitored_online]` (e.g. different sound, or `show_toast = true` only for offline), the offline-specific settings are silently ignored and the online settings are used instead.

Looks like a copy-paste oversight from when `MonitoredOffline` was added alongside `MonitoredOnline`. The fix is straightforward — just changing the two references from `config.monitored_online` to `config.monitored_offline` in the `MonitoredOffline` match arm.

mango766:
## Summary

Fixes #1659

The `Notification::MonitoredOffline` handler in `src/notification.rs` was using `config.monitored_online` instead of `config.monitored_offline` in two places:

- `self.execute(&config.monitored_online, ...)` — controls toast, sound, delay, and include/exclude behavior
- `config.monitored_online.request_attention` — controls whether the window requests user attention

This is a copy-paste issue from the `MonitoredOnline` arm directly above it. The result is that any user-configured `[notifications.monitored_offline]` settings (different sound, toast preference, etc.) are silently ignored in favor of the online config.

## Changes

- Changed both `config.monitored_online` references in the `MonitoredOffline` match arm to `config.monitored_offline`

## Test plan

- Verified the two-line diff only touches the `MonitoredOffline` arm and does not affect any other notification path
- Confirmed that `config::Notifications` already has a `monitored_offline` field (used nowhere else in the notify function, further confirming it was simply missed)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,8 +25,8 @@ Fixed:
 
 Thanks:
 
-- Contributions: @KaiKorla, @englut
-- Bug reports: chmod222, @whitequark, @englut, sebbu, @ascarion, @4e554c4c, @BKVad1m
+- Contributions: @KaiKorla, @englut, @mango766
+- Bug reports: chmod222, @whitequark, @englut, sebbu, @ascarion, @4e554c4c, @BKVad1m, @mango766
 - Feature requests: @gAlleb, @jtbx, @cyrneko, @englut
 
 # 2026.4 (2026-03-03)
--- a/src/notification.rs
+++ b/src/notification.rs
@@ -145,7 +145,7 @@ impl Notifications {
             }
             Notification::MonitoredOffline(targets) => {
                 self.execute(
-                    &config.monitored_online,
+                    &config.monitored_offline,
                     notification,
                     if targets.len() == 1 {
                         "Monitored user is offline"
@@ -157,7 +157,7 @@ impl Notifications {
                     None,
                 );
 
-                config.monitored_online.request_attention
+                config.monitored_offline.request_attention
             }
             Notification::FileTransferRequest {
                 nick,


andymandias (approved):
Looks good to me.  Thank you!

SwirlyStone5877:
Hey! I realise that this was three months ago, but I'm not sure whether any maintainers noticed this. Commit https://github.com/squidowl/halloy/commit/826dc9d5b8a513900f193bdb3ad57abc96f2adcc which appears to be from this PR contains `Co-Authored-By: Claude (claude-opus-4-6) <noreply@anthropic.com>` which goes against the AI policy as listed on the website (https://halloy.chat/contributing#ai).

andymandias:
> Hey! I realise that this was three months ago, but I'm not sure whether any maintainers noticed this. Commit https://github.com/squidowl/halloy/commit/826dc9d5b8a513900f193bdb3ad57abc96f2adcc which appears to be from this PR contains `Co-Authored-By: Claude (claude-opus-4-6) <noreply@anthropic.com>` which goes against the AI policy as listed on the website (https://halloy.chat/contributing#ai).

While it was not noticed by me at the time, we had not set an AI policy when it was submitted.  So there wasn't any guidance on the matter available for contributors then.  There has been some discussion in #1814 on the incongruity.

SwirlyStone5877:
Ah, noted. Apologies!

PR #1661, opened 2026-03-16
More Multiline Fixes

andymandias:
Fixes for:
- Re-sending IRCv3 multiline messages in the event they fail to send.
- Don't allow actions (`/me`) and regular PRIVMSGs to both be contained in a single IRCv3 batch, because the encoding for actions doesn't allow them both to exist in a single message.

Sent actions are also restricted to IRCv3 multline concatenation utilization (i.e. when IRCv3 draft/multiline is supported then one long `/me` can be sent, but an action with linebreaks cannot be).  This is more of a coherence/stylistic change rather than a fix.

### DIFF
--- a/data/src/capabilities.rs
+++ b/data/src/capabilities.rs
@@ -4,6 +4,7 @@ use std::string::ToString;
 
 use irc::proto::{self, Tags, command, format};
 
+use crate::message::formatting::{Formatting, Modifier, update_with_modifier};
 use crate::{Target, User, config, message};
 
 // This is not an exhaustive list of IRCv3 capabilities, just the ones that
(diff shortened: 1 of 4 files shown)


4e554c4c:
i really don't think ctcp should be allowed in a multiline batch, even if it uses multiline-concat

since this will be a malformed ctcp for any multiline-unsupporting client

andymandias:
By malformed, you mean subsequent lines won't show up as an action, correct?

4e554c4c:
even the initial line, no? since an action has to have x01 on both the beginning and end of the message.

andymandias:
Horse docs doesn't appear to require a trailing `\x01`: https://modern.ircdocs.horse/dcc.html#action

4e554c4c:
it requires it in sending, not in receiving https://modern.ircdocs.horse/ctcp#message-syntax

but i don't know if all software supports that

andymandias:
Are we not talking about other clients receiving?  Or is the worry a server won't support receiving it?

andymandias:
This appears to contain the relevant spec changes: https://github.com/ircv3/ircv3-specifications/pull/499/files

Which requires some changes to concatenation that I will incorporate into this PR.

4e554c4c:
formatting and ctcp are not really the same here, but yeah from this reading of the spec i think that starting with `0x1` could be ok? it's just a matter of what you do with the closing `0x1`

andymandias:
My reading of the prevailing sentiment in #ircv3 was that ACTION doesn't mix well with multiline, so I'm happy to forbid it on the expectation that we'll be in line with other implementations that way.  I will leave the concat-allowed version in the commit history, so it will be easy to switch back in the event sentiment changes.

PR #1667, opened 2026-03-17
Add `AWS_LC_SYS_CFLAGS="-O0"` to fix flatpak build

englut:
Seeing the following error when building the latest `iced-update` branch for flatpak:

```log
warning: aws-lc-sys@0.38.0:    47 |  #error "The CPU Jitter random number generator must not be compiled with optimizations. See documentation. Use the compiler switch -O0 for compiling jitterentropy.c."
warning: aws-lc-sys@0.38.0:       |   ^~~~~
[...]

PR #1669, opened 2026-03-17
Add `AWS_LC_SYS_NO_JITTER_ENTROPY="1"` in flatpak manifest to fix build error

englut:
Reopening #1667 🙏

### DIFF
--- a/assets/flatpak/org.squidowl.halloy.json
+++ b/assets/flatpak/org.squidowl.halloy.json
@@ -25,7 +25,8 @@
             "buildsystem": "simple",
             "build-options": {
                 "env": {
-                    "CARGO_HOME": "/run/build/halloy/cargo"
+                    "CARGO_HOME": "/run/build/halloy/cargo",
+                    "AWS_LC_SYS_NO_JITTER_ENTROPY": "1"
                 }
             },
             "build-commands": [


casperstorm:
THIS time im not gonna close it. But i swear... it was automatically closed before!

englut:
> THIS time im not gonna close it. But i swear... it was automatically closed before!

🤣 It's probably cause the branch it was a PR for got deleted. All good! 🫶

andymandias:
My recommendation was from looking at a few GH issues along with [the documentation for `AWS_LC_SYS_NO_JITTER_ENTROPY`](https://aws.github.io/aws-lc-rs/resources.html#entropy-configuration) and surmising that we're probably better off with jitter entropy than not.

As for testing, I was able to confirm that it triggered a recompile of `aws-lc-sys` (so presumably the flags are taking affect) and operation of Halloy appears unchanged, but since I didn't have flatpak tools installed I relied on @englut to confirm that the change prevented the flatpak build error.

englut:
The alternative is to put this in the flatpak manifest directly (since the error is only with flatpak builds). I'm open to either solution!

nekopsykose:
doesn't this build /all/ of aws-lc with -O0? that would probably not be what you want

englut:
> doesn't this build /all/ of aws-lc with -O0? that would probably not be what you want

Based on the error, I found that using this `AWS_LC_SYS_NO_JITTER_ENTROPY=1` also allows the build to pass. But I'm not certain which env var is better.

andymandias:
Not sure how to get `-O0` into the build environment for just `jitterentropy.c`, or why it shows up as an error for Flatpak's build and not regular `cargo build`.  I suspect we'll be hard pressed to tell the difference either way (less secure vs less optimized), so I don't have a strong preference between them.

englut:
Sounds like there's some uncertainty, shall I just put `AWS_LC_SYS_NO_JITTER_ENTROPY=1` in the flatpak manifest, since this only affects flatpak builds?

englut:
Alright, I'm seeing issues like this: https://github.com/aws/aws-lc-rs/issues/1008#issuecomment-3774105038, in the `aws-lc-rs` repo, where the recommended approach (for now) is to pass `AWS_LC_SYS_NO_JITTER_ENTROPY=1`.

I've updated the PR to reflect this, and it's only for the flatpak manifest, so it doesn't affect any of the other builds.

casperstorm (approved):

Issue #1655, opened 2026-03-15
FR: Direct Image Uploads to Third Party


WinnerWind:
Right now, to upload an image into a chatroom, one must use a third party site like https://x0.at. This means that one goes to that site, uploads their image/attachment and then returns to halloy with the new link.

It would be nice if Halloy could allow an optional third party integration that allows for pasting attachments into the textbox automatically, and Halloy uploads it to a configurable third party at message sending time and adds a link into the message.

furudean:
This sounds like a worse version of https://github.com/squidowl/halloy/issues/726 IMO. if people want to upload things to file hosts irrespective of an IRC protocol there is already other software to do this. This is not something that should be in Halloy.

WinnerWind:
Hi,
I don't use Soju, i'm on ZNC which, IIRC, does not provide this protocol. 

This option could be helpful for those who don't use a Bouncer, too.

classabbyamp:
FILEHOST is becoming a draft ircv3 spec, so it could get implemented for znc

4e554c4c:
> This option could be helpful for those who don't use a Bouncer, too.

The FILEHOST spec can be implemented by an ircd as well. There is nothing necessitating a bouncer in its spec.

4e554c4c:
That being said, the FILEHOST protocol is entirely controlled by the upstream (ircd/bouncer). Even though you could get around with with a local proxy/bouncer, it'd be more convenient to be able to override it in the config

I'd understand having a `override_filehost=<url>` thing in the config that works the same as `FILEHOST` but that halloy uses first

PR #1680, opened 2026-03-20
support `soju.im/FILEHOST` and `draft/FILEHOST`

### LINKED ISSUE #726: Support `soju.im/FILEHOST`
Soju's [`FILEHOST`](https://codeberg.org/emersion/soju/src/branch/master/doc/ext/filehost.md) allows an irc client to upload files to a designated http host to share as hyperlinks. This avoids proxy issues that can limit DCC, and allows for automating the common process of uploading images / pastes to a file host and sharing them.

Existing implementations:
* [goguma](https://codeberg.org/emersion/goguma/src/commit/f510b8b82b46c5c282f41a11de48458730226ec8/lib/widget/composer.dart#L465)
* [gamja](https://codeberg.org/emersion/gamja/src/commit/9fef11564db2c9d8dd436f9095c1ba294e18ccda/components/composer.js#L138)

Implementing this would involve:
1. Supporting the `soju.im/FILEHOST` ISUPPORT token
2. Adding a UI button for uploading files
3. Adding code to check MIME Types and preparing files to upload

furudean:
adds four ways to upload a file:
- `/upload` (with path autocomplete)
- paste
- drag and drop
- system file picker

the first time a file is uploaded to a host, a modal will appear asking you to confirm if this is okay.

many files can be sent at once. uploads can be canceled.

~~credentials are not sent by default, (unless connected through a bouncer). this breaks from the spec. this feature is opt in under the flag `server.<name>.filehost.send_credentials`. for non-bouncer servers~~ no longer true - see comments

fixes #726
relates to #1655

## notes and future improvements

- drag and drop does not work on wayland (is upstream https://github.com/iced-rs/iced/issues/2538)
- i'm not sure paste works on wayland either? (https://github.com/1Password/arboard/issues/207)
- you can only drag and drop into the currently active buffer, which is unintuitive. the reason is that buffers do not receive drop events, only the whole window does. so mapping it correctly is subject to hacks and i opted to not add any more hacks.
- when `soju.im/bouncer-networks` are used with soju, the bouncer's filehost will be used. draft/filehost does not pass through with soju due to [soju#374](https://codeberg.org/emersion/soju/issues/374), but may in the future.

## open questions
- ~~what happens if there is both `soju.im/FILEHOST` and `draft/FILEHOST`, what url should be used? and how does this interact with `soju.im/bouncer-networks`?~~ soju currently does not pass through draft/FILEHOST and is not planning to do so, so there is not much to do here other than using bouncer's filehost
- ~~should we send credentials? i cant think of many reasons to not do this implictly, especially given that the feature is enabled by default~~ not send credentials by default with a screen to confirm

## todo
- [x] expand mime type support. use a library?
- [ ] ~~`/upload` should be able to take many files~~ (does not work nicely with spaces, not doing this)
- [x] `/upload` should auto complete file paths like `cd`...?
- [x] drag and drop
- [x] clipboard paste
- [x] a setting to disable this feature if you don't want it
- [x] a setting to overwrite filehost url per server
- [x] documentation


casperstorm:
image


This right here gentlemen is what we all are chasing. You hit compile and you see it works ❤️ 
Let me know when you want me to review, @furudean

furudean:
i will open up the pull request when its ready for review. there still are some features i'd like to add!

casperstorm:
perfect, if you run into any design or technical questions then just ping me or @andymandias and we will help.

furudean:
I'm having second thoughts with sending the user's credentials and I'm strongly now leaning that this is a **bad** idea. A server being able to specify an ISUPPORT token for uploads (that may be mistakenly sent) is an abuse vector. 

There are no semantics in `draft/FILEHOST` that defines how authentication should be made with the server. The non-normative example contains an Authorization header, but does not define _how_ authentication should be done. `soju.im/filehost` on the other hand, does specify:

> Clients SHOULD authenticate their HTTP request with the same credentials used on the IRC connection (e.g. HTTP Basic for SASL PLAIN, HTTP Bearer for SASL OAUTHBEARER).

It follows by implementation that soju's upload server does require credentials, or the request will fail. 

The client should follow the semantics of the upload server, but this is not specified well for `draft/FILEHOST`. Guessing is a bad idea. We could dynamically activate `send_credentials` when `soju.im/filehost` and  send credentials in that case, though I don't like this very much either.

I would prefer if it instead explicit in the config and then documented how to use this feature with soju instead. It's more to configure, but we avoid potentially leaking people's IRC credentials just for connecting.

reesericci:
It might make sense here to have a standardized OAuth/OIDC-style handshake between the file hosting providers, the bouncer/server and the client. I'd be happy to write up an RFP to IRCv3 to add this as the standard if people would be interested. Here's a little mermaid diagram of how this could work:

```mermaid
sequenceDiagram
actor Client
participant Server
participant Filehost@{ "type": "database"}
	Filehost-->>Server: Requests OIDC configuration
	Server->>Filehost: Returns OIDC configuration with JWKS url
	Filehost-->>Server: Requests JSON Web Keyset (JWKS) to verify tokens issued by the server
	Server->>Filehost: Returns JWKS
	Client-->>Server: Establish connection
	Server->>Client: Advertises FILEHOST
	Client -->>Server: Requests access token
	Server -->>Filehost: Requests nonce and provides client IP address to expect upload from
	Filehost->>Server: Returns nonce (bound to client IP)
	Server->>Client: Returns signed JWT with expiration, nonce and hashed username as data
	Client->>Filehost: Sends file using multipart/form-data with key `file` and the JWT in the Authorization header
	Filehost->>Client: Verifies signature with cached JWKS, nonce, expiration, audience and if valid returns file URL
```

furudean:
@reesericci i think this is out of scope for this PR, but in general i positive to this change. there is a need for servers to validate that uploads are made by credentialed users. someone (who?) had also suggested extjwt in the past. feel free to hop on the #halloy or #ircv3 channels on libera if you want to chat about this!

docs/configuration/servers.md
```diff
@@ -743,3 +743,46 @@ relevant when combined with the `exclude` setting.
 [servers.<name>.confirm_message_delivery]
 include = { channels = ["#halloy"] }
 ```
+
+## `filehost`
+
+See the [Filehost](/guides/filehost) guide for usage information.
```
furudean:
```suggestion
See the [Filehost guide](/guides/filehost) for usage information.
```

furudean:
I walked back some of the stuff on credentials, if soju.im/bouncer-networks and a filehost is detected then credentials will be sent (though the user will be prompted with a modal if this is really okay). The docs have more details on how its implemented exactly.

I still would like to see the spec changed to work using tokens, but I believe this is enough safeguards to ship this.

docs/guides/filehost.md
```diff
@@ -0,0 +1,80 @@
...
+soju requires you to [authenticate](#Authentication) your upload requests
+:::
+
+### convoyeur
+
+[convoyeur](https://codeberg.org/classabbyamp/convoyeur) can act as a proxy to external file upload services. 
+
+### Others
+
+If your server does not allow advertising a filehost token, or you do not control your server, you can configure a URL manually via [`filehost.override_url`](/configuration/servers#override_url).
```
furudean:
```suggestion
If your server does not allow advertising a filehost token, or you do not control the IRC server, you can configure a URL manually via [`filehost.override_url`](/configuration/servers#override_url).
```

furudean:
I've been running https://github.com/slingamn/ergo-filehost on my personal IRC server and all of a sudden it makes a lot of sense to send credentials by default. Many of my users use Halloy, so that additional friction of enabling it isn't great. I wonder if the confirmation modal is good enough to send credentials...? If you trust your IRC server, you should also trust that its caps are valid and safe.

On the other hand if users provide their own URL via `override_url` the potential for leaks increases - they may use a public host that they don't have control over and send their IRC credentials there. We should assume these users know what they're doing, and if somehow they do not, they still get a modal asking them to confirm with big red letters.

furudean:
~~i'd also like to add support for client certificates via SASL EXTERNAL. currently only basic auth is supported. this PR is already quite big so that might be better as a follow up~~ this was done, not a big effort

casperstorm:
Is this ready for review? :)

furudean:
yes, sorry for the back and forth. i'm happy with this!

casperstorm:
Since this is a big PR, @furudean you are free to squash those 66 commits. It might be easier for you to keep rebased because it might take a bit to review.

data/src/config/filehost.rs
```diff
@@ -0,0 +1,39 @@
+use serde::Deserialize;
+
+#[derive(Debug, Clone, Deserialize)]
+#[serde(default)]
+pub struct FileHost {
+    /// Master switch to disable all file upload functionality. Defaults to `true`.
+    pub enabled: bool,
+    /// Show the upload button in the input bar. Defaults to `true`.
+    pub button: bool,
+    /// Handle uploadied files from the clipboard via paste. Defaults to `true`.
```
casperstorm:
```suggestion
    /// Handle uploaded files from the clipboard via paste. Defaults to `true`.
```

data/src/config.rs
```diff
@@ -358,6 +362,7 @@ impl Config {
             pub keyboard: Keyboard,
             pub notifications: Notifications,
             pub file_transfer: FileTransfer,
+            pub filehost: FileHost,
```
casperstorm: Nit: `file_host` if our struct is `FileHost`, otherwise it should be `filehost: Filehost`. Personally i would go with the latter.

src/screen/dashboard.rs
```diff
@@ -63,6 +75,9 @@ pub struct Dashboard {
     previews: preview::Collection,
     preview_client: Option<Arc<reqwest::Client>>,
     buffer_settings: dashboard::BufferSettings,
+    pub file_being_hovered: bool,
+    known_filehosts: data::KnownFilehosts,
+    pending_filehost_upload: Option<PendingFilehostUpload>,
 }
```
casperstorm: This is quite a lot to "slam into" Dashboard state. We should make a manger for this; `filehost::Manager` or similar.

data/src/client/on_connect.rs
```diff
@@ -106,7 +106,8 @@ pub fn on_connect(
                             | command::Internal::Exec(_)
                             | command::Internal::Hop(_, _)
                             | command::Internal::SysInfo
-                            | command::Internal::Reconnect => None,
+                            | command::Internal::Reconnect
+                            | command::Internal::Upload(_) => None,
```
casperstorm, line 110: When i do `/upload` i see no tooltip, but i haven't configured anything and i also don't see the upload button in text input so i feel like we perhaps should tell user that its not configured or something when you do `/upload`? 🤔

casperstorm (commented):
Thanks for attacking this. Its a big feature. I have done a quick first review.

data/src/client/on_connect.rs
furudean: The current user experience of using `/upload` without a filehost is that the command isn't really recognized. I don't know if there is a precedent here to handle commands that are invalid due to state but I think showing an error message is a good change

andymandias: I had a similar thought. Our existing method is to not show commands that aren't available on the server, but I think it could be better for users if we show them the command then display an immediate error on selecting the command when it is not supported by the server. Updating all the old completions (e.g. chathistory, monitor, etc) wouldn't need to be done in this PR.

src/screen/dashboard.rs
furudean: i created a filehost::Manager and used the Message pattern to pass state back and forth. let me know what you think

data/src/config.rs
furudean: done!

docs/guides/filehost.md
```diff
@@ -0,0 +1,83 @@
...
+### soju
+
+soju supports filehost via its own `soju.im/FILEHOST` token. Configure filehost [in your soju configuration](https://codeberg.org/emersion/soju/src/branch/master/doc/file-upload.md).
+
+soju will advertise the configured URL to clients via `soju.im/FILEHOST`. For networks provided by bouncer, Halloy will use the bouncer's filehost.
+
+::: info
+soju requires you to [authenticate](#Authentication) your upload requests
+:::
+
```
furudean:
```suggestion
```
this is no longer relevant

docs/guides/filehost.md
```diff
@@ -0,0 +1,83 @@
...
+[`filehost.send_credentials`](/configuration/servers#send_credentials) controls whether Halloy sends credentials with upload requests. It defaults to `true`.
+
+- **SASL PLAIN** — sends an `Authorization` header with server `username:password`
+- **SASL EXTERNAL** — presents the client certificate
+
+::: warning
+Only send credentials to filehosts you trust.
+:::
+
```
furudean:
```suggestion
```

since this is now a default, this warning is meaningless

src/buffer/input_view/completion.rs
```diff
@@ -206,9 +215,17 @@ impl Completion {
...
-            Some(column![emojis_view, command_view].spacing(4).into())
+        if command_view.is_some()
+            || emojis_view.is_some()
+            || paths_view.is_some()
+        {
+            Some(
+                column![emojis_view, paths_view, command_view]
+                    .spacing(4)
+                    .into(),
+            )
```
furudean, line 228: i guess the +nightly formatter changed its behavior? this was not an intentional change

src/buffer/input_view/completion.rs
```diff
@@ -1240,6 +1260,20 @@ fn connected_command_list<'a>(
...
     command_list.extend(isupport_commands);
 
+    if has_filehost {
+        command_list.push(Command {
+            title: "UPLOAD".into(),
+            args: vec![Argument {
+                text: "file".into(),
+                kind: ArgumentKind::Optional { skipped: false },
+                tooltip: Some(
+                    "Path to a file, or omit to open file picker".to_string(),
```
furudean:
```suggestion
                    "Path to a file".to_string(),
```
this is inaccurate - must provide a filename now

src/screen/dashboard.rs
```diff
@@ -1912,6 +1944,39 @@ impl Dashboard {
...
+                    .height(Length::Fill),
+                )
+                .width(Length::Fill)
+                .height(Length::Fill)
+                .style(theme::container::transparent_overlay),
+            ]
+            .into()
+        } else {
+            base
         }
```
andymandias: Drag & drop without being able to drag onto individual buffers is a little rough, but that can be addressed in another PR.  For the time being, there are a couple scenarios I think may be a little confusing for users: 
- If multiple windows are open the file gets uploaded to the most recently active buffer, not necessarily the window the file is dragged onto.
- If the focused buffer is not compatible with filehost then there's no feedback.

I think a quick clarification could be adding a bit more info to the modal maybe? Something like "Unable to upload file, no filehost available for {server}" if not available, and "Drop to upload file to {buffer} @ {server}" to make it clear which pane is targeted?

data/src/config/server.rs
```diff
@@ -232,6 +233,29 @@ impl Default for Server {
...
+    }
+}
+
+#[derive(PartialEq, Eq, Debug, Clone, Deserialize)]
+#[serde(default)]
+pub struct Filehost {
+    /// Whether to use the server's filehost. Defaults to `true`.
+    pub enabled: bool,
+    /// Override the filehost URL advertised by the server via ISUPPORT
+    pub override_url: Option<String>,
```
andymandias, line 247: I think it would be nice to be able to specify auth for the `override_url`.  That way if a user has their own filehost provider that requires auth they don't have to have the same credentials for the filehost as the server.

src/buffer/input_view.rs
```diff
@@ -1055,6 +1184,94 @@ impl State {
...
+                    ));
+                }
+                self.input_content.perform(text_editor::Action::Edit(
+                    text_editor::Edit::Paste(std::sync::Arc::new(url)),
+                ));
+                history.record_draft(RawInput {
+                    buffer: buffer.clone(),
+                    text: self.input_content.text(),
+                });
+                (self.focus(), None)
```
andymandias: I think we want `(Task::none(), None)` here.  Otherwise if a user switches what buffer they're typing in while a large file uploads then the buffer that the URL pastes into will steal focus back when it pastes.

src/buffer/input_view.rs
```diff
@@ -1055,6 +1184,94 @@ impl State {
...
+                self.spinner_hovered = false;
+                (Task::none(), None)
+            }
+            Message::SpinnerHovered(hovered) => {
+                self.spinner_hovered = hovered;
+                (Task::none(), None)
+            }
+            Message::FilehostUrlReady(url)
+                if !url.is_empty() && self.uploading > 0 =>
+            {
```
andymandias, line 1247: Maybe we could add an `INFO` level log when a file upload succeeds?  Just to have a way to reference successful uploads (without having to add a "File Uploads" pane to this PR), just in case the user deletes the URL by accident or whatever.

src/buffer/input_view/completion.rs
```diff
@@ -1733,6 +1775,70 @@ impl Text {
...
     }
 
+    fn process_paths(&mut self, input: &str, cursor_position: usize) {
+        use std::path::Path;
+
+        let Some(word) = get_word(input, cursor_position) else {
+            *self = Self::default();
+            return;
+        };
+
```
andymandias, line 1785: It's not documented (my fault), but the `word` returned by `get_word` is space delimited.  I think we want something like
```rust
        let Some((_, path)) = input.split_once(' ') else {
            *self = Self::default();
            return;
        };
```
so we can play nice with spaces in filenames.

andymandias (commented):
Has worked well in my testing, and has a lot of nice touches!  Haven't looked through all the lines yet, but it seems like it's close.

data/src/config/server.rs
furudean: that would mean pulling in all of `password_file` and `password_command` etc, and how do we handle external certificate-based authentication? it sounds like a lot of complexity to pull in.

what's the use-case here in practice?

src/buffer/input_view.rs
furudean: hmm, this is a very valid point. i don't personally use multiple buffers at a time so i overlooked this. i don't love that focus isn't returned to the input, because what you would want to do after completing an upload is hit enter to send it. makes it feel less fast than it can be.

src/buffer/input_view.rs
furudean: good idea. done! 👍

src/buffer/input_view/completion.rs
furudean: good suggestion, i had not tested spaces in filenames. i pushed a commit with this change

src/modal/confirm_file_upload.rs
```diff
@@ -0,0 +1,58 @@
...
+                    .style(|theme, status| theme::button::secondary(
+                        theme, status, false
+                    ))
+                    .on_press(Message::ConfirmFileUpload),
+                button(text("Do not"))
+                    .padding(5)
+                    .style(|theme, status| theme::button::secondary(
+                        theme, status, false
+                    ))
+                    .on_press(Message::Cancel),
```
casperstorm: Lets do `Upload` and `Close` here, similar to connect_to_server.rs

src/buffer/input_view.rs
andymandias: Over here with `(Task::none(), None)` I don't see any focus loss if I stay in the same text input, typing more or not.  You're seeing that happen when you make that change?

src/screen/dashboard.rs
casperstorm: Agree, i would even consider disabling drag to drop files by default 🤔

data/src/config/server.rs
andymandias: In my case I have filehost set up on my Soju instance.  I also directly connect to a few servers (for testing purposes in my case, but I can imagine some other rationales).  `override_url` doesn't work for me since the auth for the filehost is not set up the same as the auth for those servers (I just get a 403 forbidden back).  I *could* add additional auth to the filehost (or to the upload service, since I control that), but that may not be convenient for everyone.

I'm okay with the functionality being in another PR.

data/src/config/filehost.rs
```diff
@@ -0,0 +1,39 @@
...
+    pub file_drop: bool,
+}
+
+impl Default for Filehost {
+    fn default() -> Self {
+        Self {
+            enabled: true,
+            button: true,
+            paste: true,
+            file_drop: true,
```
casperstorm: As noted elsewhere; i think we should default to false on `file_drop` for the time being.

src/filehost.rs
```diff
@@ -0,0 +1,208 @@
...
+        upload_url: String,
+        has_credentials: bool,
+        window: window::Id,
+    },
+}
+
+pub struct Manager {
+    pub file_being_hovered: bool,
+    known: data::KnownFilehosts,
+    pending: Option<PendingUpload>,
```
casperstorm: What happens if you upload two files at the same time? 🤔

furudean: only the first one is tracked, oops... this could just be a Vec, let me fix that

data/src/fileupload.rs
```diff
@@ -0,0 +1,154 @@
...
+    if matches!(auth, Some(Auth::External { .. })) && base.scheme() != "https" {
+        return Err(Error::InsecureTransport);
+    }
+
+    let file_name = path.file_name().map_or_else(
+        || String::from("file"),
+        |n| n.to_string_lossy().into_owned(),
+    );
+
+    let bytes = tokio::fs::read(path).await?;
```
casperstorm: I am wondering if we should consider a streamed request body instead? I am just worried about loading the entire file into memory if it is a big one.

casperstorm:
You did an amazing job with this feature, i've added a few nits here and there but we should be able to land this soon.
Could you update CHANGELOG as well, when you have the time?

src/buffer/input_view.rs
furudean: when you press the upload button in the buffer you lose focus, is what i mean. it's never returned after that except manually. that's the use case for the focus. i'm ok to lose it either way, if it does more harm than good

data/src/config/server.rs
furudean: let's follow up on this in another PR

src/buffer/input_view.rs
andymandias: I'd rather focus neither be lost nor stolen, but I'm not sure whether I'm seeing what you're seeing.  Is the order of operations for you: click on the plus button on the right end of the text input, select a file from native file picker, at which point the Halloy window regains focus but the buffer you clicked the upload button in doesn't?

data/src/config/filehost.rs
furudean: done

src/filehost.rs
furudean: done!

src/modal/confirm_file_upload.rs
furudean: done

data/src/fileupload.rs
furudean: it was easy enough to implement so i went ahead and did this. at first i was worried that `infer` wouldn't work but then i realized we could just read the first few bytes and pass that to infer before continuing to read the rest :D

src/buffer/input_view.rs
furudean: what you describe is what i'm seeing. you press the button, focus is lost on the buffer. my idea was to "return" it, with `self.focus()` but as you point out that isn't straightforward to do

src/screen/dashboard.rs
furudean: I changed the drag and drop text to `Drop to upload file to {buffer} @ {server}` and also disabled drag and drop by default, along with a suite of other fixes to more consistently show the overlay. The states should be a bit more clear to users now, no matter how it's configured.

Not perfect, but we can't upload to a specific buffer until Iced gets an update.

src/buffer/input_view.rs
andymandias: Interesting, what OS/DE are you on again?  @casperstorm are you able to reproduce that behavior if you change `(self.focus(), None)` to `(Task::none(), None)`?

furudean: i'm on macOS

furudean:
thank you! i have updated the changelog as well

data/src/fileupload.rs
casperstorm: Awesome, i think that is the right thing to do!

src/buffer/input_view.rs
furudean: i ended up changing this to `Task::none()` as i can't reproduce it and now i'm unsure if i ever could

reesericci:
OK I started working on a spec that cleans up the security issues found in this and would appreciate people's feedback: https://github.com/ircv3/ircv3-specifications/pull/597

casperstorm (approved):
Well done 🚀

andymandias (approved):
Thank you for tackling this, it's a great addition!

ncfavier:
Dropping files seems to work even when `file_drop` is disabled. It says "File drop is disabled here" but then proceeds to upload the file anyway.

andymandias:
> Dropping files seems to work even when `file_drop` is disabled. It says "File drop is disabled here" but then proceeds to upload the file anyway.

Thank you for the report.  We will be testing & polishing a bit more before release, so please open new issues for anything you find.

PR #1685, opened 2026-03-21
Warn on insecure password file permissions

melocene:
This is a follow up to f180e41061db393acf65bc99f5c5e7397586d9cb which adds a warning if password files have insecure permissions. This is similar to how SSH will warn on insecure private keys, but unlike SSH this will not block using said keys/password files. It is only a small warning in the log. The code only runs on Unix platforms so Windows users won't see the warning.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ Security:
 
 Thanks:
 
-- Contributions: @KaiKorla, @englut, @mango766, @ncfavier, @classabbyamp
+- Contributions: @KaiKorla, @englut, @mango766, @ncfavier, @classabbyamp, @melocene
 - Bug reports: chmod222, @whitequark, @englut, sebbu, @ascarion, @4e554c4c, @BKVad1m, @mango766, esden
 - Feature requests: @gAlleb, @jtbx, @cyrneko, @englut, @ascarion
 


furudean:
If we do this, can we also warn on unsafe sasl-external perms? https://halloy.chat/configuration/servers#sasl-external

data/src/config.rs
```diff
@@ -612,6 +612,24 @@ fn create_owned_file(path: &Path, contents: &[u8]) -> std::io::Result<()> {
...
+#[cfg(unix)]
+pub fn check_password_file_permissions(path: &Path) {
+    use std::os::unix::fs::PermissionsExt;
+    if let Ok(meta) = std::fs::metadata(path) {
+        let mode = meta.permissions().mode();
+        if mode & 0o077 != 0 {
+            log::warn!(
+                "password file {:?} has permissions {:#o}; consider restricting to 0600",
+                path,
+                mode & 0o777,
```
andymandias: Any reason in particular to use the `Debug` print here instead of `path.display()`?  Would be nice for the two permissions listed to have the same format, preferably one that would with `chmod`.  So either
```rust
                "password file {:?} has permissions 0{:o}; consider restricting to 0600",
```
or
```rust
                "password file {:?} has permissions {:o}; consider restricting to 600",
```
I guess?

data/src/server.rs
andymandias: I agree with @furudean, it would be nice to check the `cert` and `key` fields of `Sasl::External` as well.

melocene:
Updated to include SASL certs and keys as well and make the log messages consistent with the existing style.

Example 1 (A standard password file with 644 permissions):**
`WARN: [libera] password file /home/user/.config/halloy/libera.pass has permissions 0644; consider restricting to 0600`

**Example 2 (A SASL external key with 777 permissions):**
`WARN: [oftc] SASL external key /home/user/.ssh/oftc_cert.key has permissions 0777; consider restricting to 0600`

PR #1705, opened 2026-03-24
bind `cmd+,`/`ctrl+,` to open config

furudean:
this has been a standard keybind in mac native programs since pretty much forever

~~there is no standard lingo for this on either linux or windows, so i didn't change anything for those platforms. i'm open to suggestions tho~~ ctrl+, on others!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,16 @@
 # Unreleased
 
+Added:
+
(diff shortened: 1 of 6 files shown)


casperstorm (commented):
Good addition, thanks. I think honestly we could make it default for Windows and Linux as well. On Linux, CTRL+, also opens configuration in many apps now days.

furudean:
done! @casperstorm

casperstorm (approved):

PR #1707, opened 2026-03-24
Query Server Messages

andymandias:
Fixes query history processing to apply server messages settings to server messages that appear in queries (e.g. `change_host` server messages).

Also introduces an `away` server message kind, to toggle whether automated away responses (`RPL_AWAY`) appear in direct messages.  The `smart` setting in this case is the minimum amount of time allowed between automated away responses; so, for example, if you set it to `900` then you will only see an automated away response every 15m at most.


data/src/capabilities.rs
```diff
@@ -127,13 +127,8 @@ pub fn multiline_concat_lines(concat_bytes: usize, text: &str) -> Vec<&str> {
             if last_space > line_start {
                 lines.push(&text[line_start..last_space + ' '.len_utf8()]);
 
-                println!(
-                    "line_bytes {line_bytes} last_space {last_space} line_start {line_start} space_len {}",
-                    ' '.len_utf8()
-                );
```
casperstorm: 😅

omentic:
Woah, thanks!

PR #1709, opened 2026-03-25
persist draft state

furudean:
adds a new setting to persist drafts in the event of a poorly-timed program exit. enjoy the confidence that your beautiful/thoughtful/weird/creepy message lives across sessions, or disable it if you don't like it

it is possible that a draft becomes inaccessible and stored forever if the server-side state changes in such a way that the user can no longer access the target. this pr contains no garbage collection mechanism for this as i couldn't figure out how to consistently detect these cases. suggestions are welcome

this probably works less well if you have multiple halloy sessions open once. but you wouldn't do that, would you...?


furudean:
done! drafts should be saved now in the exit hook

PR #1711, opened 2026-03-25
Fix shortcuts and command bar handling in popped out windows

### LINKED ISSUE #1710: Keyboard shortcuts don't work in non-main windows
### System information

Version: 2026.5
Commit: 9bf68d8bddcb5a85f759ab18eceb0e23f45568d7
GPU Backend: Vulkan
GPU Adapter: Apple M1 Max (G13C C0)

### Reproduction steps

1. Open a buffer in a new window
2. Switch to the new window
3. Press <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>M</kbd> to toggle the nick list, <kbd>Ctrl</kbd>+<kbd>↓</kbd> to scroll to the end of the buffer, <kbd>PgUp</kbd> or <kbd>PgDown</kbd>, or any other registered shortcut

### Expected behavior

The action that's mapped to the shortcut is executed.

### Actual behavior

Nothing happens.

KaiKorla:
Fixes #1710

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,12 @@ Fixed:
 - Server notices with wildcard targets (for example `NOTICE *`) are shown in the server buffer again
 - Fix URLs with non-ASCII characters getting percent-encoded when rendered in buffers
 - chathistory: support for TARGETS pagination and more conservative treatment of pagination on reconnect
+- Keyboard shortcuts now work in popped out buffer windows, including command bar shortcuts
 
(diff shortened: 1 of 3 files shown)


andymandias (approved):
Nice, thank you!

My only suggestion is that we could make it so that using the command bar shortcut, when the command bar is already open on another window that isn't currently in focus, could both close the command bar on the other window (as it does now) and open the command bar on the current window.

PR #1712, opened 2026-03-25
Show wildcard server notices in server buffer

KaiKorla:
Route server NOTICE messages without a user source, such as NOTICE *, to the server buffer instead of dropping them. Add a regression test for wildcard-target server notices and document the fix in the changelog.

I can't test it on a real server at the moment. Maybe someone else can give it a try.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Added:
 Fixed:
 
 - `sidebar.order_by` setting works when set to `"config"`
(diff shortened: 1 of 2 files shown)


andymandias:
I will add, I also don't have a set-up to test this in the wild.  I have a rinky-dink test server (and will hopefully soon have *two* such servers), but if there is someone who is willing and able to test as a server operator that would be appreciated.

PR #1713, opened 2026-03-25
Normalize hosts and decode channel targets in IRC URIs

### LINKED ISSUE #1691: irc uri scheme with ipv6 doesnt connect
`irc://[ipv6]/channel` does not actually connect, it works when you add it via the config file (without brackets)

```
01:51 connecting to server...
01:51 connection to server failed (io error: failed to lookup address information: Name or service not known)
```

KaiKorla:
Fix #1691.

Let me know if this new added crate is ok for you.

### DIFF
--- a/data/src/url.rs
+++ b/data/src/url.rs
@@ -1,10 +1,11 @@
 use std::str::FromStr;
 
 use fancy_regex::Regex;
+use percent_encoding::percent_decode_str;
 
 use crate::appearance::theme;
-use crate::config;
 use crate::server::ServerName;
+use crate::{config, isupport};
 
 #[derive(Debug, Clone)]
 pub enum Url {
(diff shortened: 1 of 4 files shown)


data/src/url.rs
```diff
@@ -109,21 +110,26 @@ fn generate_server_name(host: &str) -> &str {
 
 fn parse_server_config(url: &url::Url) -> Option<config::Server> {
     let nickname = config::random_nickname();
-    let server = url.host()?.to_string();
+    let server = match url.host()? {
+        url::Host::Ipv6(address) => address.to_string(),
```
andymandias, line 117: Let's make a note here that the purpose of not using `url::Host::to_string` is to strip the square brackets, for visitors in the future.

data/src/url.rs
```diff
@@ -109,21 +110,26 @@ fn generate_server_name(host: &str) -> &str {
...
                 return None;
             }
 
-            if channel.starts_with('#') {
-                Some(channel.to_string())
+            if matches!(channel.chars().next(), Some('#' | '&' | '+' | '!')) {
+                Some(channel.into_owned())
             } else {
                 Some(format!("#{channel}"))
             }
```
andymandias, line 144: If we want to expand allowed channel prefixes to be as inclusive as possible and avoid false positives, we should probably probably delay channel determination until we have an available CHANTYPES to reference.  That would also allow the first prefix in CHANTYPES to be used as the default (i.e. that gets prepended to un-prefixed targets), as suggested by [the URL draft horsedocs links](https://datatracker.ietf.org/doc/html/draft-butcher-irc-url-04).  That would be the most thorough handling, but it is a fair amount of work.  Alternatively, we could use `DEFAULT_CHANNEL_PREFIXES` which is the existing fallback channel prefix set constant.

andymandias (changes requested):
Looks good, and it looks like a good crate choice to me as well.  Only potential change I think we want is re: channel prefixes.

andymandias (approved):
Nice work, thank you!

Issue #1714, opened 2026-03-25
Decode URLs before displaying them


whitequark:
Here's how Chrome displays [this URL](https://en.wiktionary.org/wiki/%E7%99%BE%E8%81%9E%E3%81%AF%E4%B8%80%E8%A6%8B%E3%81%AB%E5%A6%82%E3%81%8B%E3%81%9A):

<img width="979" height="127" alt="Image" src="https://github.com/user-attachments/assets/cf5983e1-9e07-418b-80b1-4b9a525b6c4a" />

Here's Halloy:

<img width="2110" height="173" alt="Image" src="https://github.com/user-attachments/assets/495a0106-4fa4-44f5-9a0f-4889edd1fd23" />

The former is readable; the latter isn't. I'd like Halloy to do the former also.

whitequark:
It actually looks like Halloy *actively percent-encodes* links it receives?

SnoopJ:
As the person who sent those messages, I can confirm that the `PRIVMSG` sent from my client had a valid UTF-8 sequence for Japanese text (i.e. Halloy has here applied an interpretation of the message I sent that _I did not intend_), which distorts what I was saying in-context (following discussion was about [Unicode Technical Standard #58](https://www.unicode.org/reports/tr58/), etc.)

As a non-Halloy user (who has been considering trying it out) I am _very_ surprised by this behavior.

andymandias:
This issue coincidentally just arose in PR #1713.  If someone gets to this before me: `url::Url::parse` percent encodes as part of the parse process, and can't be relied on to produce slices of the original string.

furudean:
the standard canonical representation of non-ascii urls is to percent encode them. this is simply how IDN works. when you copy that url out of your browser, it will add the string `https://en.wiktionary.org/wiki/%E7%99%BE%E8%81%9E%E3%81%AF%E4%B8%80%E8%A6%8B%E3%81%AB%E5%A6%82%E3%81%8B%E3%81%9A` to the pasteboard.

the correct way to render urls for presentation is to use something like the [idna crate](https://crates.io/crates/idna) to decode

furudean:
i can corroborate that all links proactively get transformed into percent encoding when pasted into halloy

whitequark:
[...]

PR #1715, opened 2026-03-26
display URLs in buffers exactly as sent

furudean:
relates to #1714

URLs containing non-ASCII characters were being displayed in their percent-encoded form in message buffers, which does not match was what actually sent.

example: https://en.wiktionary.org/wiki/百聞は一見に如かず would appear as https://en.wiktionary.org/wiki/%E7%99%BE%E8%81%9E%E3%81%AF%E4%B8%80%E8%A6%8B%E3%81%AB%E5%A6%82%E3%81%8B%E3%81%9A when sent as a message

## how this was fixed

halloy was changing the url when passing it through `Fragment::Url(url::Url)` in order to create a canonical representation. this canonical representation was then rendered in the ui. the change is now pass the url through as `Fragment::Url(String)`, storing the original. we can still parse the url in places that need a `Fragment::url()`, so this results in no logical change to how they're stored

### DIFF
--- a/src/buffer/scroll_view.rs
+++ b/src/buffer/scroll_view.rs
@@ -182,18 +182,17 @@ fn has_visible_preview(
         return fragments.iter().filter_map(message::Fragment::url).any(
             |url| {
                 // Check if URL is hidden by user
-                if message.hidden_urls.contains(&url) {
+                if message.hidden_urls.contains(url) {
                     return false;
                 }
 
                 // Check if URL is in visible URLs list
-                if !visible_urls.contains(&url) {
+                if !visible_urls.contains(url) {
                     return false;
                 }
 
                 // Check if preview is loaded and visible for source
-                if let Some(preview::State::Loaded(preview)) =
-                    previews.get(&url)
+                if let Some(preview::State::Loaded(preview)) = previews.get(url)
                 {
                     let is_visible_for_source =
                         if let Some(visible_for_source) = visible_for_source {
@@ -212,14 +211,15 @@ fn has_visible_preview(
     false
 }
 
-fn eligible_preview_urls(
-    urls: impl IntoIterator<Item = url::Url>,
+fn eligible_preview_urls<'a>(
+    urls: impl IntoIterator<Item = &'a url::Url>,
     hidden_urls: &HashSet<url::Url>,
     max_per_message: usize,
 ) -> Vec<url::Url> {
     urls.into_iter()
-        .filter(|url| !hidden_urls.contains(url))
+        .filter(|url| !hidden_urls.contains(*url))
         .take(max_per_message)
+        .cloned()
         .collect()
 }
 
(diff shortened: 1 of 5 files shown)


src/buffer/scroll_view.rs
```diff
@@ -415,10 +416,7 @@ pub fn view<'a>(
...
                 {
                     let urls = eligible_preview_urls(
-                        fragments
-                            .iter()
-                            .filter_map(message::Fragment::url)
-                            .cloned(),
+                        fragments.iter().filter_map(message::Fragment::url),
                         &message.hidden_urls,
                         config.preview.max_per_message,
                     );
```
andymandias, line 422: Putting the parse inside `view` here means all URLs in the render buffer will get (re-)parsed ~every frame.  We can probably afford to do that, but it would be preferable not to.

We could change `Message.hidden_urls` to a `HashSet<String>` and have `eligible_preview_urls` output a `Vec<String>`, but it is nice to have a canonical representation of the URL when interacting with the preview image cache.  This is why I recommended storing a normalized version next to the canonical version, but that does mean we'd use ~2x storage for URLs.

Admittedly, none of the quick solutions that come to mind are *particularly* appealing.  Any ideas @casperstorm?  If not I think we can do the extra parsing or use the extra storage.

furudean: if we don't want the extra cpu cost, we could have a `raw` field like we do with messages and reference that. is that what you had in mind for a storage-based solution?

andymandias: Essentially, yes.  If you look at `message::Fragment::User`, for example, the second field is the raw representation and the first field has the normalized user (and its metadata).  The only tricky thing there is that the deserialization procedure for `Fragment` will need to be able to deserialize the old format as well as the new (the conversion `Url` → `String` avoids that because they have the same serialized representation).

If you do want to try that, then be sure to run the tests beforehand, and if you don't have access to chathistory be sure to back up your message store, because it's easy to wind up overwriting your history with a blank one if something goes wrong.

But we can wait for casper to weigh in; it's not the end of the world if we have that extra parsing in there, just not ideal.

andymandias: In case it's not clear by "tests" I meant running cargo test for the data crate in particular.

furudean: i pushed a commit which addresses the rendering perf by trading off for 2x storage. we can undo if we prefer code simplicity

andymandias (approved):
Just to restate my code comment more officially: I don't see a simple way to get this without paying a small performance or memory penalty.  Implementations of both variants mentioned have done proactively and both look good to me, I just want to confirm with @casperstorm that he doesn't see a way to avoid paying.  Thank you!

casperstorm:
Looks good to me, i just rebased to handle the conflict. Thank you.

PR #1722, opened 2026-03-27
Chathistory Improvements

andymandias:
Two improvements in automatic chathistory handling:
- Use more conservative calculations of batch size, making use of helpful answers/discussion in `#ircv3` (thanks to emersion, slingam, and jwheare), in order to provide better reliability when handling chathistory pagination on reconnect.
- Add pagination support to chathistory TARGETS

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -860,7 +860,10 @@ impl Client {
                                 // soju treats it as the number of batch
                                 // messages;  for our purposes the number of
                                 // batch messages is the more conservative
-                                // option
+                                // option.  This is expected to change in the
+                                // future, for all chathistory implementation to
+                                // treat a multiline message as a single
+                                // message.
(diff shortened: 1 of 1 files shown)


data/src/client.rs
```diff
@@ -854,14 +854,23 @@ impl Client {
                                 .and_then(|batch| self.batches.get_mut(batch))
                             {
                                 parent.events.extend(finished.events);
+
+                                // Ergo treats a multiline message a single
+                                // message toward the chathistory limit while
+                                // soju treats it as the number of batch
+                                // messages;  for our purposes the number of
+                                // batch messages is the more conservative
+                                // option
```
casperstorm: 👍🏻

casperstorm (approved):
Seems like a reasonable change to me!

PR #1723, opened 2026-03-27
Disable automatically start on passive or quiet installation runs

### LINKED ISSUE #1720: Windows Installer: Don't run halloy after /quiet install (winget)
Currently halloy is launched each time when silently updated by winget. It would be good If this can be suppressed

KaiKorla:
Fix/Add #1720


andymandias (approved):
Looks good to me.  I'm using https://learn.microsoft.com/en-us/windows/win32/msi/installer-uilevel as reference since don't have a Windows box to test at the moment.

PR #1724, opened 2026-03-28
Ergo/Solanum Mass Message

andymandias:
Reroutes out-going Ergo/Solanum mass messages to the server buffer.


esden:
I have tested this branch once again. It all seems to be working well enough.

When receiving notices from ergo they end up being sorted into the senders query pane. But as far as I understand redirecting it to the server buffer is a bigger issue. I think having it in the query window is *good enough*:tm:

It is definitely already a huge improvement over having confusing errors. :)

Thank you for putting it together @andymandias

PR #1731, opened 2026-03-29
+typing=done should not be sent when the message is sent

### LINKED ISSUE #1729: +typing=done should not be sent when the message is sent
from the spec:

> The typing=done notification MAY be sent by clients once when the user clears the text-input field without sending a message.

I read this as `typing=done` should only be sent if no message was sent.

casperstorm:
Fixes #1729 by not sending done after sending a message.


andymandias (approved):
Looks good!

PR #1733, opened 2026-03-29
Popped typying style

casperstorm:
Add `style` to `typing` with two variants: `popped` and `padded`.
* Padded: Reserves a space for typing indicators so that content wont shift when it appears. This is the style currently in main.
* Popped: When typing indicator appears it is popped inbetween messages and text input making the messages shift upwards.


andymandias (approved):
Looks good!  Made some minor changes to get channel and query buffers to have a more consistent look.

I don't really have a sense which style would be more popular, but I think popped is a bit more common than padded so I think it's the right default?

PR #1734, opened 2026-03-29
Ignore User `/MOTD` re: Registration 

andymandias:
Ignores `RPL_ENDOFMOTD`/`ERR_NOMOTD` past the first received (i.e. does not assume they denote the end of registration), assuming those are due to user requests for the MOTD.


casperstorm (approved):
Good addition 👍🏻

PR #1735, opened 2026-03-29
add "copy message" to right-click menu

### LINKED ISSUE #1504: Copying selection via context menu
I think the selection can usually (always?) be copied via Ctrl+C, but it would be nicer for people who mainly use mouse for this to have a context menu option to do this. This already exists for links, just not plaintext.

achille:
adds a "copy message" right click menu. 
text is computed lazily from &message::Content only when the menu renders
closes https://github.com/squidowl/halloy/issues/1504"


andymandias (approved):
Nice and clean.  Works well.  Thank you!

PR #1736, opened 2026-03-29
fix multiline messages missing from chathistory replay

achille:
halloy fails to load multiline messages from chathistory because they're nested inside the chathistory batch. 
this assembles nested draft/multiline batches before extending into parent, routes through handle_chathistory to preserve history (no notifications, ctcp filtering), and strips the batch tag to avoid double-counting parent size.


andymandias:
Thank you for catching this!  I appreciate the tests as well.

andymandias (approved):

PR #1743, opened 2026-03-31
Typing indicator additions

casperstorm:
* Change text to dot icons
* Vertically centered the icons 
* Config to adjust hide animation
* Config to adjust animation size


andymandias (approved):
Looks good to me!  It's a bit fiddly, but I think the only other option is to convert it into an SVG (which has limited alignment options at the moment).

When the animation is off there are no dots;  I'm okay with that, or with restoring dots as `•••` or `...` in the static case.

PR #1746, opened 2026-04-01
Unrequest Attention

andymandias:
Clears any existing attention request for the window when focusing that window.


casperstorm (approved):
Not tested but make sense to me.

PR #1753, opened 2026-04-04
Input Improvements

### LINKED ISSUE #1726: `/mode #channel =b` is valid but not accepted
some ircds (like solanum) support `MODE #channel =b` (or another list mode) for listing the contents of that mode. the `/mode` command validation doesn't accept that.

Image

andymandias:
Various minor improvements to the input:
- Validate /monitor command
- Always show tooltips for Halloy supported commands, but produce an error if the server does not support the command
- Show tooltip for and validate `/invite` command (along with improvements to INVITE message display)
- Show tooltip for and validate `/massmessage` command
- Support =-prefix for /mode command (fixes #1726)
- Restore draft when opening a server buffer (partially addressing #1754)
- Allow `/upload` in server buffers (as appropriate)
- Split default server messages' `dimmed` setting into two, one for "actions" (e.g. invite, monitor, kick, standard reply, etc) and one for "passive" server messages (e.g. join, part, quit, change-host, etc).  "Actions" will trigger unread and not be dimmed by default, while "passive" server messages will not trigger unread and are dimmed by default.

### DIFF
--- a/data/src/command.rs
+++ b/data/src/command.rs
@@ -24,7 +24,7 @@ pub use self::alias::Alias;
 #[derive(Debug, Clone)]
 pub enum Command {
     Internal(Internal),
-    Irc(Irc),
+    Irc(Irc, Option<Warning>),
 }
 
 #[derive(Debug, Clone, PartialEq)]
@@ -367,10 +367,13 @@ fn parse_command(
     config: &Config,
 ) -> Result<Command, Error> {
     let unknown = || {
-        Command::Irc(Irc::Unknown(
-            cmd.to_string(),
-            args.iter().map(ToString::to_string).collect(),
-        ))
+        Command::Irc(
+            Irc::Unknown(
+                cmd.to_string(),
+                args.iter().map(ToString::to_string).collect(),
+            ),
+            None,
+        )
     };
 
     match cmd.parse::<Kind>() {
@@ -487,11 +490,11 @@ fn parse_command(
                         }
                     }
 
-                    Ok(Command::Irc(Irc::Join(chanlist, chankeys)))
+                    Ok(Command::Irc(Irc::Join(chanlist, chankeys), None))
                 })
             }
             Kind::Motd => validated::<0, 1, false>(args, |_, [target]| {
-                Ok(Command::Irc(Irc::Motd(target)))
+                Ok(Command::Irc(Irc::Motd(target), None))
             }),
             Kind::Nick => validated::<1, 0, false>(args, |[nick], _| {
                 if let Some(isupport::Parameter::NICKLEN(max_len)) =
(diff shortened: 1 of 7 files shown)


data/src/features.rs
```diff
@@ -0,0 +1,38 @@
...
+}
+
+impl Features {
+    pub fn enable_supported(&mut self, server_version: &str) {
+        if server_version.starts_with("soju") {
+            self.detach = true;
+            self.version_request = VersionRequest::Need(true);
+        } else if server_version.starts_with("ergo") {
+            self.mass_message = true;
+        } else if server_version.starts_with("solanum") {
```
classabbyamp, line 33: i would avoid hard-coding solanum in this. `=mode` support came from its predecessor charybdis, and solanum (like all in its ancestral line) have a long history of being forked with a new name

data/src/command.rs
```diff
@@ -953,23 +992,32 @@ fn parse_command(
 
                 Ok(Command::Irc(Irc::Away(comment)))
             }),
-            Kind::SetName => validated::<1, 0, true>(args, |[realname], _| {
+            Kind::SetName => {
```
casperstorm: > Servers advertising the setname capability MUST support usage of the command even while the capability is not negotiated. Clients MUST NOT prevent users from manually using the command while the capability is not negotiated.

Hmm, I might be wrong here but i read the docs as we should always parse `/setname`? 🤔
And then if `NAMELEN` exists we can enforce it, and if it doesnt we send the command without validation.

data/src/features.rs
classabbyamp: I had started working on this before I saw this PR, and I just made the validation a bit more relaxed: https://github.com/squidowl/halloy/compare/main...classabbyamp:halloy:modal-equality

data/src/command.rs
```diff
@@ -1512,9 +1695,24 @@ fn parse_command(
...
+                        Err(Error::CommandNotEnabled { command: "upload" })
+                    }
+                } else {
+                    Err(Error::CommandNotAvailable {
+                        command: "upload",
+                        context: buffer.map_or(String::new(), |buffer| {
+                            format!(" on {}", buffer.server())
+                        }),
+                    })
+                }
```
casperstorm: Image

If i add a `override_url` i see `+`-button and can choose a file, but when i type `/upload` i get an error.
Do we want to gate it all behind FILEHOST? also the button?

data/src/command.rs
andymandias: Technically users aren't prevented from manually sending via `/raw` (as they never are from sending any command that way), but we can make the error not prevent send or not show the error.

data/src/command.rs
andymandias: Ah, forgot about that.  Thanks!

data/src/features.rs
andymandias: Error has been downgraded to a warning when not confirmed working.

PR #1776, opened 2026-04-07
actually block file_drop if disabled

### LINKED ISSUE #1755: filehost: File dropping always works
Dropping files seems to work even when `file_drop` is disabled. It says "File drop is disabled here" but then proceeds to upload the file anyway.

cc @furudean https://github.com/squidowl/halloy/pull/1680

furudean:
fixes #1755

missed a spot!


andymandias (approved):
Thanks!

PR #1781, opened 2026-04-07
Add support for "draft/ICON"

luca020400:
- **Support draft/ICON**
- **Cache the server icons**


data/src/server_icon/cache.rs
```diff
@@ -0,0 +1,250 @@
...
+    environment::cache_dir()
+        .join("server_icons")
+        .join("images")
+        .join(&digest.as_ref()[..2])
+        .join(&digest.as_ref()[2..4])
+        .join(&digest.as_ref()[4..6])
+        .join(format!(
+            "{}.{}",
+            digest.as_ref(),
+            image::ImageFormat::Ico.extensions_str()[0]
```
luca020400: that's wrong, should use extension from format.

data/src/server_icon.rs
```diff
@@ -0,0 +1,187 @@
...
+    url: Url,
+    http_client: Arc<reqwest::Client>,
+) -> Result<Icon, LoadError> {
+    let response = http_client
+        .get(url.clone())
+        .send()
+        .await?
+        .error_for_status()?;
+
+    let bytes = response.bytes().await?;
```
casperstorm: lets not buffer the whole server provided icon response without a size limit 🤔

casperstorm:
Nice feature, and over all shape. One quick thing i noticed is that we should have the option to disable the server icons.

luca020400:
Something I didn't think of with the config is how to handle Bouncers, since you can disable the bouncer icon, but this config also gets replicated to each sub-network

data/src/server_icon.rs
```diff
@@ -0,0 +1,232 @@
...
+            return Task::none();
+        };
+
+        let Some(http_client) = http_client else {
+            log::warn!(
+                "[{server}] File upload disabled: Unable to build HTTP client"
+            );
+            self.drop_request(&server);
+            return Task::none();
+        };
```
casperstorm, line 60: Shoudn't this be something like: "server icon fetching disabled"

data/src/server_icon.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    let format = image::guess_format(&bytes).map_err(LoadError::ParseImage)?;
+
+    let mut hasher = Sha256::default();
+    hasher.update(&bytes);
+
+    let digest = cache::HexDigest::new(&hasher.finalize());
+    let image_path = cache.blob_path(&digest, format.extensions_str()[0]);
+
+    if !image_path.exists() {
```
casperstorm: We could do this instead to avoid synchronous filesystem checks:

```rust
if !fs::try_exists(&image_path).await? { }
```

casperstorm (commented):
Minor stuff, but overall gret. Let's get the last few nits fixed and then we can merge.

data/src/server_icon.rs
luca020400: I think I just copied that from filehost.rs, fixed

casperstorm (approved):

PR #1785, opened 2026-04-08
filehost: use proxy configuration

furudean:
filehost did not inherit from the proxy reqwest interface, meaning no filehost uploads were going through any configured proxy

also renames `preview_proxy_client` to the more generic `http_client`: indicating it serves http traffic in general. it was already being used for filehost (oops?) but maybe this is fine? also open to having two http clients for-purpose, though they should be mostly equivalent in their configuration. curious what maintainers think about this.

as a part of this refactor, we also reduced duplicated code paths that created a proxy client or regular client (yay!)

suggested by @andymandias in irc


casperstorm (approved):
Nice one.

andymandias:
Thank you!

PR #1794, opened 2026-04-09
Improve tab ergonomics

luca020400:
This is a quick/first attempt towards #1790

Tha main issue atm is alias handling, meaning I don't handle them at all they collapse into the same underlying command.

### DIFF
--- a/src/buffer/input_view.rs
+++ b/src/buffer/input_view.rs
@@ -1021,7 +1021,7 @@ impl State {
             Message::Tab(reverse) => {
                 let cursor_position = self.input_content.cursor().position;
 
-                if let Some(entry) = self.completion.tab(reverse)
+                if let Some(entry) = self.completion.tab(reverse, config)
                     && let Some(line) = self
                         .input_content
                         .line(cursor_position.line)
@@ -1036,12 +1036,7 @@ impl State {
                         config,
                     );
 
-                    let result =
-                        self.on_completion(buffer, history, actions, true);
-                    self.process_completion_and_notice(
-                        buffer, clients, history, config,
-                    );
-                    result
+                    self.on_completion(buffer, history, actions, true)
                 } else {
                     (Task::none(), None)
                 }
--- a/src/buffer/input_view.rs
+++ b/src/buffer/input_view.rs
@@ -1040,11 +1040,7 @@ impl State {
                         self.on_completion(buffer, history, actions, true);
 
                     // If there is only one tab candidate process the completion immediately.
-                    if self
-                        .completion
-                        .tab_candidate_count()
-                        .is_some_and(|count| count == 1)
-                    {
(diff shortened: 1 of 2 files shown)


casperstorm (approved):
This is really solid. Thank you.

luca020400:
> This is really solid. Thank you.

This depends on #1792 (that I manually picked in my branch)
I don't know how gh pr handle this scenario, but I'll probably have to rebase when that one gets merge.

furudean:
should a space also be inserted after the command? if you do `/up` + <kbd>tab</kbd> you get `/upload`, then you have to hit a space to get any actual completion for the commands. we could save a keystroke here potentially by making it `/upload ` (note the space) right away

luca020400:
> should a space also be inserted after the command? if you do `/up` + <kbd>tab</kbd> you get `/upload`, then you have to hit a space to get any actual completion for the commands. we could save a keystroke here potentially by making it `/upload ` (note the space) right away

That should be possible, I'll have to specifically case that for commands as I think with emojis you may not want it.

That also matches the behaviour I've had on weechat so it would ease my muscle memory

luca020400:
This is good to go, I'd appreciate if it gets merged before the next release :)

andymandias (commented):
This works really well, thank you!

There are a couple small tweaks I think would be nice:
- Tab-complete for commands is a bit awkward in that the command is typed out but the command description is not available until an additional character is typed.  I think it would be nice if the command description was also visible (similar to how the emoji picker is visible above the command description when using the emoji picker for a command).
- The command picker shows aliases, but cycles through the root command instead.  E.g. `/q` will auto-complete to `/quit` and `/msg` (instead of `/query`).  I realize this behavior predates this PR, but the improved auto-complete highlights the behavior.  It would be nice if we could fix that behavior.

luca020400:
> This works really well, thank you!
> 
> There are a couple small tweaks I think would be nice:
> 
> * Tab-complete for commands is a bit awkward in that the command is typed out but the command description is not available until an additional character is typed.  I think it would be nice if the command description was also visible (similar to how the emoji picker is visible above the command description when using the emoji picker for a command).

Maybe?

[Screencast_20260421_162204.webm](https://github.com/user-attachments/assets/639ff993-a528-4710-b571-c1ad61f0600e)

> * The command picker shows aliases, but cycles through the root command instead.  E.g. `/q` will auto-complete to `/quit` and `/msg` (instead of `/query`).  I realize this behavior predates this PR, but the improved auto-complete highlights the behavior.  It would be nice if we could fix that behavior.

That should be done

andymandias (approved):
Nicely done, thank you!

luca020400:
Fyi https://github.com/squidowl/halloy/pull/1866
I don't think it actually matters, but I changed behavior by mistake.

PR #1798, opened 2026-04-09
filehost: upload progress in input

furudean:
possible fix to #1764

shows a string like (uploading...) where the upload is going to be, and then replaces it with the url once done. preserves cursor position like the user may expect.

uploads are now inserted where the cursor is, not at the end of the input

### DIFF
--- a/src/buffer/input_view.rs
+++ b/src/buffer/input_view.rs
@@ -1329,6 +1329,7 @@ impl State {
                                     text_editor::Motion::DocumentEnd,
                                 ),
                             );
+                            // insert a space if there is none immediately before the URL
                             if !self.input_content.text().trim_end().is_empty()
                             {
                                 self.input_content.perform(
--- a/src/buffer/input_view.rs
+++ b/src/buffer/input_view.rs
@@ -1309,11 +1309,12 @@ impl State {
                 match url {
                     Some(url) => {
                         if let Some(ghost_byte_pos) = content.find(&ghost) {
-                            let ghost_char_pos = UnicodeSegmentation::graphemes(
-                                &content[..ghost_byte_pos],
-                                true,
-                            )
-                            .count();
+                            let ghost_char_pos =
+                                UnicodeSegmentation::graphemes(
+                                    &content[..ghost_byte_pos],
+                                    true,
+                                )
+                                .count();
                             let replaced = content.replacen(&ghost, &url, 1);
                             let delta = url.chars().count() as i64
                                 - ghost.chars().count() as i64;
@@ -1367,11 +1368,12 @@ impl State {
                         .find_map(|s| content.find(&s).map(|pos| (s, pos)));
 
                         if let Some((search, ghost_byte_pos)) = found {
-                            let ghost_char_pos = UnicodeSegmentation::graphemes(
-                                &content[..ghost_byte_pos],
-                                true,
-                            )
-                            .count();
+                            let ghost_char_pos =
+                                UnicodeSegmentation::graphemes(
+                                    &content[..ghost_byte_pos],
+                                    true,
+                                )
+                                .count();
                             let replaced = content.replacen(&search, "", 1);
                             let delta = -(search.chars().count() as i64);
                             let cursor = adjust_cursor(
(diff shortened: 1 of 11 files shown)


src/buffer/input_view.rs
```diff
@@ -2683,3 +2834,186 @@ fn macos_clipboard_file() -> Option<std::path::PathBuf> {
...
+    line: usize,
+    col: usize,
+) -> usize {
+    let mut chars = 0;
+    for i in 0..line {
+        chars += content.line(i).map_or(0, |l| l.text.chars().count()) + 1;
+    }
+    chars += col;
+    chars
+}
```
andymandias, line 2817: `text_editor::Position` is actually somewhat cursed in that `line` is as you expect it to be, but `column` is bytes within the line (if I'm remembering correctly).  And then motion is by grapheme, and delete is I believe by character (so you can, for example, decompose 👩🏾‍🚒 into 👩🏾 then 👩 via backspace).  I've modified a couple tests to hopefully help demonstrate/test the quirks.

src/buffer/input_view.rs
```diff
@@ -2683,3 +2834,186 @@ fn macos_clipboard_file() -> Option<std::path::PathBuf> {
...
     path.is_file().then_some(path)
 }
+
+fn upload_ghost(id: u32) -> String {
+    if id <= 1 {
+        String::from("[Uploading...]")
+    } else {
+        format!("[Uploading...{id}]")
+    }
+}
```
andymandias, line 2789: Not necessary, but it might be nice to have a short, UUID-like string representation of the ID.  Slightly less likely to match, and produces a predictable length.

andymandias (commented):
Short on time so this is just a preliminary review with only first impressions.  Apologies if I've overlooked something with the comments.

I think this feature has real utility, but I do worry it will be tricky to ensure we disturb user input as little as possible.

Feel free to ignore this, but could we get similar utility by stacking a thin progress bar on the top of the input view?  Something like the thin red progress bar that appears on the top of the page on YouTube when navigating the site.  I have not worked out the details, but I think we could use `reqwest::Body::wrap_stream` for tracking progress.  (Or if that doesn't work for some reason, an indeterminate progress indicator could be used.)

src/buffer/input_view.rs
furudean: i pushed some code to address this. basically just using UnicodeSegmentation::graphemes to determine where the offsets are instead of a naive char length

src/buffer/input_view.rs
furudean: i wanted to avoid making the placeholder too long, but this isn't a bad idea. we could ditch the sequential tracking in that case and have less LoC. currently, there might be situations where someone does write the string `[Uploading...]` into their buffer and end up with a surprise.

another idea i had was to utilize some kind of character that users won't type, like `b"\0"` to delimit, though we may have to filter such characters out before sending a message. i have no idea how those are handled in inputs either.

furudean:
the code is definitely complex on its own, and there may be edge cases we have not considered yet. i think i've fixed the issue you flagged with emojis (and other graphemes) as the tests pass now.

i don't _really_ want to have two loading indicators, so it would be a question on whether or not to replace the one we have with something that can accommodate more situations. what we don't get by having a more overall visible upload indicator is the ability to edit around uploads with the ghosts, so that's the trade-off. i'm preferential to my own solution of course ;-)

src/buffer/input_view.rs
andymandias: Definitely don't want it too long, was thinking of something like [`small_uid`](https://docs.rs/small_uid/latest/small_uid/) or [`cuid`](https://docs.rs/cuid/latest/cuid/) that have fairly short options.  The only thing that gives me pause with using characters outside ASCII alphanumeric is that we have to worry about how servers, bouncers, and other clients might handle it if the user sends the ghost.

Either way, could be a follow-up doesn't necessarily need to be in this PR.

andymandias (approved):
That's a good point re: having a placeholder to edit around.  Would be good to have a progress indicator at some point, and I agree it should be one indicator or the other (but I don't have a strong preference between the two).  Let's make that a separate PR from this one.

I've pushed a few more tests that aim to be a bit more comprehensive, all passing :)

furudean:
I think these changes are good! I tested the behavior a bit locally and I find it natural to use.

The tests are especially appreciated, I'm glad we're testing actual UI behaviors!

PR #1801, opened 2026-04-09
Reaction notifications

englut:
Adds notifications for reactions!

- unreacts don't fire off notifications
- notifications are sent if:
  - window is unfocused
  - reaction's buffer (channel or query) is not the current open buffer.
  - react is by non-self
  - reacted to message is originally sent by you

~There's still one more scenario that needs solving, but may require a bit more consideration:~

- ~notifying only when reacts are on your own messages~

~Currently, this isn't obvious how I can achieve this. I think I would need to attach the original message to the reaction context.~

~Leaving this in draft until that is sorted.~

### DIFF
--- a/data/src/history.rs
+++ b/data/src/history.rs
@@ -278,21 +278,22 @@ pub async fn append(
         if let Some(message) =
             find_reaction_target(&mut all_messages, &id, &pending.server_time)
         {
-            if message.is_echo && message.direction == Direction::Received {
-                if let Ok(target) = Target::try_from(message.target.clone()) {
-                    let message_text = message.text();
-                    for reaction in pending.clone().reactions.into_iter() {
-                        let reaction_to_echo = ReactionToEcho {
-                            reaction: reaction::Context {
-                                inner: reaction,
-                                target: target.clone(),
-                                in_reply_to: id.clone(),
-                                server_time: pending.server_time,
-                            },
-                            message_text: message_text.clone(),
-                        };
-                        pending_reactions_flushed.push(reaction_to_echo);
-                    }
+            if message.is_echo
+                && message.direction == Direction::Received
+                && let Ok(target) = Target::try_from(message.target.clone())
+            {
+                let message_text = message.text();
+                for reaction in pending.clone().reactions.into_iter() {
+                    let reaction_to_echo = ReactionToEcho {
+                        reaction: reaction::Context {
+                            inner: reaction,
+                            target: target.clone(),
+                            in_reply_to: id.clone(),
+                            server_time: pending.server_time,
+                        },
+                        message_text: message_text.clone(),
+                    };
+                    pending_reactions_flushed.push(reaction_to_echo);
                 }
             }
 
@@ -309,9 +310,9 @@ pub async fn append(
         },
     );
 
-    let result = overwrite(kind, &all_messages, read_marker, chathistory_references)
-        .await;
-    result.map(|_| pending_reactions_flushed)
+    overwrite(kind, &all_messages, read_marker, chathistory_references)
+        .await
+        .map(|()| pending_reactions_flushed)
 }
 
 pub async fn delete(kind: &Kind) -> Result<(), Error> {
@@ -754,14 +755,14 @@ impl History {
 
                     return Some(
                         async move {
-                            let result = overwrite(
+                            overwrite(
                                 &kind,
                                 &messages,
                                 read_marker,
                                 chathistory_references,
                             )
-                            .await;
-                            result.map(|_| vec![])
+                            .await
+                            .map(|()| vec![])
                         }
                         .boxed(),
                     );
@@ -837,18 +838,16 @@ impl History {
                 chathistory_references,
                 pending_reactions,
                 ..
-            } => {
-                let result = append(
-                    &kind,
-                    seed,
-                    pending_messages,
-                    read_marker,
-                    chathistory_references,
-                    pending_reactions,
-                )
-                .await;
-                result.map(|_| ())
-            }
+            } => append(
+                &kind,
+                seed,
+                pending_messages,
+                read_marker,
+                chathistory_references,
+                pending_reactions,
+            )
+            .await
+            .map(|_| ()),
             History::Full {
                 kind,
                 messages,
--- a/data/src/history.rs
+++ b/data/src/history.rs
@@ -10,7 +10,7 @@ use futures::{Future, FutureExt};
 use tokio::fs;
 use tokio::time::Instant;
 
-pub use self::manager::{Manager, Resource};
+pub use self::manager::{Manager, PendingReaction, Resource};
 pub use self::metadata::{Metadata, ReadMarker};
 use crate::capabilities::LabeledResponseContext;
 use crate::message::{self, Direction, MessageReferences, Source};
@@ -267,8 +267,10 @@ pub async fn append(
     read_marker: Option<ReadMarker>,
     chathistory_references: Option<MessageReferences>,
     pending_reactions: HashMap<message::Id, reaction::Pending>,
-) -> Result<(), Error> {
+) -> Result<Vec<PendingReaction>, Error> {
     let loaded = load(kind.clone(), seed).await?;
+
+    let mut pending_reactions_flushed: Vec<PendingReaction> = vec![];
 
     let mut all_messages = loaded.messages;
     // pending reactions should only exist for unloaded history entries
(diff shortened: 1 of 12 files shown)


englut:
Okay! I actually managed to figure out how to link the message body and if the original message was self or not. But new problem: it's only recognized when the buffer is open. Otherwise no body or self check is performed (and no notification as a result).

I'm unsure how to access messages from a buffer that's not open. This is where my already limited halloy knowledge is even more sorely lacking :pray:

englut:
I managed to get a future working to fetch & filter messages on pending reactions (thx @andymandias for putting me onto that!), but I'm reading the full history from disk in that scenario (when there's occupied partial history). Not sure if that's inefficient or not, but the notifications work in all the scenarios I want now!

englut:
Refactored based on feedback from @andymandias (thanks again :pray: :heart_eyes:):
- Swapped out `ReactionTarget` for the message text, if it's a message that was originally sent by self.
- Removed the full history load & filter for original reaction message. Replaced with parsing `pending_reactions` from flush. There's a tradeoff with this approach, less disk read, but slight delay for notifications (in inactive buffers) until the buffers are flushed.

src/notification.rs
```diff
@@ -340,6 +342,61 @@ impl Notifications {
...
+                        Option<String>,
+                        String,
+                    ) = if config.reaction.show_content {
+                        (
+                            reaction.inner.sender.to_string(),
+                            Some(format!(
+                                "Reacted {} to your message in {react_sent_in}",
+                                reaction.inner.text
+                            )),
+                            message_text.to_string(),
```
andymandias: I feel like the reaction itself is the main content, while the message_text is context, so I'd suggest we include it in the body.  Something like:

```rust
                            Some(react_sent_in),
                            format!(
                                "Reacted {} to your message {message_text}",
                                reaction.inner.text
                            ),
```

where it's being assumed the reaction itself will be short the vast majority of the time.

src/notification.rs
```diff
@@ -340,6 +342,61 @@ impl Notifications {
...
+                if config.reaction.should_notify(
+                    &User::from(reaction.inner.sender.clone()),
+                    channel,
+                    server,
+                    *casemapping,
+                ) {
+                    let react_sent_in = match channel {
+                        Some(channel) => channel.as_normalized_str(),
+                        _ => "your direct message",
+                    };
```
andymandias, line 377: For consistency with other notifications, how about:

```rust
                    let react_sent_in = match channel {
                        Some(channel) => {
                            if cfg!(target_os = "macos")
                                || !notification_config.show_content
                            {
                                format!("{channel} ({server})")
                            } else {
                                format!("{channel}, {server}")
                            }
                        }
                        None => {
                            if cfg!(target_os = "macos")
                                || !notification_config.show_content
                            {
                                format!("query ({server})")
                            } else {
                                format!("query, {server}")
                            }
                        }
                    };
```

data/src/history/manager.rs
```diff
@@ -98,7 +107,7 @@ impl Manager {
             log::debug!("cleared messages for {kind}");
 
             return task.map(move |task| {
-                task.map(move |result| Message::Flushed(kind, result))
+                task.map(move |_| Message::Flushed(kind, Ok(vec![])))
```
andymandias: We should use `Message::Flushed(kind, result.map(|_| vec![]))` so we don't lose any error that might be produced.

data/src/history/manager.rs
```diff
@@ -1828,35 +1847,35 @@ impl Data {
...
             hash_map::Entry::Occupied(mut entry) => {
-                entry.get_mut().add_reaction(
-                    in_reply_to,
-                    reaction,
-                    server_time,
-                );
-
-                None
+                let message_text =
+                    entry.get_mut().add_reaction(reaction.clone());
```
andymandias: We can drop this clone.

data/src/history.rs
```diff
@@ -267,15 +267,44 @@ pub async fn append(
...
+            if message.is_echo && message.direction == Direction::Received {
+                let target = match message.target.clone() {
+                    message::Target::Channel { channel, source: _ } => {
+                        Some(target::Target::Channel(channel))
+                    }
+                    message::Target::Query { query, source: _ } => {
+                        Some(target::Target::Query(query))
+                    }
+                    _ => None,
+                };
```
andymandias: Let's make this an `impl TryFrom<message::Target> for target::Target`, that way it won't be replicated if it's needed again.

data/src/history.rs
```diff
@@ -732,13 +764,14 @@ impl History {
...
-                            overwrite(
+                            let _ = overwrite(
                                 &kind,
                                 &messages,
                                 read_marker,
                                 chathistory_references,
                             )
-                            .await
+                            .await;
+                            Ok(vec![])
```
andymandias: As in `clear_messages` we should use `map` here to maintain any error we receive.

data/src/history.rs
```diff
@@ -815,15 +848,16 @@ impl History {
...
                     &kind,
                     seed,
                     pending_messages,
                     read_marker,
                     chathistory_references,
                     pending_reactions,
                 )
-                .await
+                .await;
+                Ok(())
```
andymandias: I agree we can discard the ok return value from `append`, but we should use `map` here to maintain any error we receive.

data/src/history/manager.rs
```diff
@@ -40,6 +41,12 @@ impl Resource {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct PendingReaction {
+    pub reaction: reaction::Context,
+    pub message_text: String,
+}
```
andymandias, line 48: Technically the reactions will no longer be pending where these are returned, they'll have been applied to the user's message(s).  Perhaps something like `ReactionToSent` or `ReactionToEcho`?  (Technically the reactions will always be applied to echoes, since echoes have ids while sent messages don't, but I'm not sure if that's clarifying or complicating.)

data/src/history/manager.rs
```diff
@@ -1828,35 +1847,35 @@ impl Data {
...
-        in_reply_to: message::Id,
-        reaction: Reaction,
-        server_time: DateTime<Utc>,
-    ) -> Option<impl Future<Output = Message> + use<>> {
+        server: Server,
+        reaction: reaction::Context,
+    ) -> (
+        Option<String>,
+        Option<impl Future<Output = Message> + use<>>,
+    ) {
```
andymandias: I think we don't expect a future for `Entry::Occupied` and we don't expect message text for `Entry::Vacant` (the created partial history will be empty, and so it won't contain a message for the reaction to be applied to).  I wonder if we could return something like `future::ready(Message::PendingReactions(server, vec![pending_reaction])` (where I'm assuming `add_reaction` has been modified to return a `PendingReaction`), such that we'd have a single path for sending reaction notifications instead of two.

englut: Took a little to wrap my head around this one. But this makes much more sense! Cleaner too. I've essentially split up the reaction event (`Event::Reaction`) and the notification for the reactions (`Event::ReactionsToEcho`). Not as intuitive as I would have hoped, but it keeps the notifications for reactions on a single path.

Thanks for the tip! I ended up using the existing future pattern in that function, because `future::ready(..)` doesn't have the same signature as the `async move {..}.boxed()` pattern if you know of a way to make it match, I can flip over to using that instead, but it currently works the way it is :)

src/notification.rs
englut: Ah. That makes sense. I've updated the notification contents!

src/notification.rs
englut: Done. Now the title shows the server info as well!

data/src/history.rs
englut: Good call! Done :pray:

data/src/history/manager.rs
englut: Renamed to `ReactionToEcho`. Thanks, I wasn't sure what to name these as everything felt weird, but seeing the `Echo` makes a lot of sense now, since these react events are tied to the echoes.

data/src/history/manager.rs
englut: Woops. That's true. Fixed this and the other similar as well :pray:

andymandias (approved):
Looks good to me!

englut:
Thanks @andymandias :pray: I appreciate you spending the time to review :heart:  I snuck in a last commit, it's just to clean up the changes I made to the `Event::Reaction` handler I realized how it was before is cleaner to read.

furudean:
great work!

PR #1804, opened 2026-04-10
fix URL parsing after parenthes path segments

### LINKED ISSUE #1800: Certain URLs are not recognized correctly in buffers
Image

`https://billwurtz.com/(What)%20Love%20Is.mp3`

casperstorm:
Fixes #1800.
This was a regression from #1715.

@furudean you mind testing?

### DIFF
--- a/data/src/message.rs
+++ b/data/src/message.rs
@@ -3618,6 +3618,25 @@ pub mod tests {
     }
 
     #[test]
+    fn fragment_url_with_extra_closing_paren_in_path() {
+        let Content::Fragments(fragments) = parse_fragments(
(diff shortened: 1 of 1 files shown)


englut:
Ahhh looks like we both worked on this! Sorry I should have noted in the issue or something.

I took a sliiiightly different path from you. I'm not sure which is more efficient, but they both work :saluting_face:  
https://github.com/squidowl/halloy/commit/00eed98bce7ac55d56fe9dfcd042d6b0864e18c9

furudean (approved):
i tested this changeset and the url parses correctly now!

andymandias (approved):

PR #1807, opened 2026-04-10
Add a common cache interface

luca020400:
Currently preview is the only user, but with draft/ICON support this helps dedup code


luca020400:
I tried to preserve the same behavior as the old cache, but since the iface is generic I may have lost something.

A quick rundown of the main difference.

Before cache::load, since it was "tied" to the preview.rc, it knew it could call super::fetch directly in case the on disk file was missing compared to the state json.

This is now replaced by a cache miss if any of the files in the state json are missing by returning None and falling back to load_uncached.

I think this is equivalent, but instead of fetching _only_ the missing image (for example) it now re-creates the whole State.

src/screen/dashboard.rs
```diff
@@ -4823,3 +4829,13 @@ fn http_client_from_config(config: &Config) -> Option<reqwest::Client> {
...
+
+fn preview_cache(config: &config::Preview) -> cache::FileCache {
+    let root = environment::cache_dir().join("previews");
+
+    cache::FileCache::new(
+        root,
+        config.request.image_cache.max_size_bytes(),
+        config.request.image_cache.trim_interval,
+    )
+}
```
casperstorm: We should refresh FileCache when we reload config file.

casperstorm (changes requested):
Just a minior thing to address, otherwise it looks good.

luca020400:
> Just a minior thing to address, otherwise it looks good.

Ah I did not think about that, that'll get a bit in the way but It'll probably have to rebuild the TrimOptions everytime (unless I make a generic Fn that gets the required configs...)

What do you think?

src/screen/dashboard.rs
luca020400: I went ahead and implemented the easiest method

casperstorm (approved):
This looks good, i think it will be nice to use in the future.
Can you rebase your ICON pr on this?

PR #1808, opened 2026-04-10
Feature/system tray

joehonkey:
Added system tray support so that closing the main window hides Halloy to the system tray instead of quitting. The app keeps running in the background and stays connected to IRC servers.

## What was changed

**Cargo.toml** — added two new dependencies: ksni (a pure Rust StatusNotifierItem implementation) and gtk (required by the menu crate on Linux).

**src/main.rs** — a HalloyTrayIcon struct was added implementing ksni::Tray with the Halloy icon, tooltip, and a right-click menu with Show and Quit options. A global tokio::sync::broadcast channel via OnceLock allows tray click callbacks to send events back into the iced update loop. A window_open bool field was added to the Halloy struct to track whether the main window is currently visible. The CloseRequested event now hides the window instead of quitting when a tray icon exists. Two new messages, TrayIconClicked and TrayMenuQuit, handle toggling the window and clean exit. A tray subscription polls the broadcast channel and feeds events into the iced update loop. The tray runs in its own thread with its own tokio runtime to avoid blocking iced's event loop.

## Platform notes

Tested and working on Arch Linux with KDE Plasma on both X11 and Wayland. The ksni crate speaks the StatusNotifierItem DBus protocol directly so it is not X11-specific and works on any freedesktop-compliant desktop that supports SNI.


joehonkey:
Closing in favour of a cleaner cross-platform rewrite.

classabbyamp:
i recommend reading https://halloy.chat/contributing#ai

PR #1810, opened 2026-04-11
handle sigint+sigterm (for drafts)

furudean:
this PR started as me trying to figure out why the drafts functionality seemed inconsistent. i came up with two changes that help:

- sigterm and sigint are now intercepted and trigger the same exit path as a normal window close, meaning history and drafts are saved. previously some things would be lost.
- draft saves on an interval instead of after inactivity, and now saves every 10 seconds


casperstorm (approved):
LGTM

andymandias (approved):

PR #1812, opened 2026-04-11
Display URLs Percent Decoded

### LINKED ISSUE #1714: Decode URLs before displaying them

andymandias:
Adds a setting to automatically decode percent-encoded urls when displaying them.

Closes #1714.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,7 @@ Removed:
 
 Thanks:
 
-- Contributions: @furudean, @omentic, @KaiKorla, @achille, @classabbyamp, @ncfavier, @englut
+- Contributions: @furudean, @omentic, @KaiKorla, @achille, @classabbyamp, @ncfavier, @englut, @WinnerWind
 - Bug reports: sebbu, @whitequark, @SnoopJ, esden, @miyukoc, @ld-cd, @achille, @classabbyamp, vignoux, @esden, @ncfavier, @englut
 - Feature requests: @omentic, @classabbyamp, @ncfavier
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ Changed:
 - Colons, semicolons and double quotes are no longer considered part of channel names
 - Backslash escapes are now only interpreted when escaping markdown formatting characters
 - Tooltips will be shown for all commands, with an error tooltip shown if the server does not support the command
+- Parsed URLs will be displayed with IDN encoded domains (to avoid domain spoofing) and percent-decoded path & later components (for legibility) by default (`display.decode_urls`)
 
 Fixed:
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2935,6 +2935,7 @@ dependencies = [
  "objc2-foundation 0.3.2",
  "open",
  "palette",
+ "percent-encoding",
  "rand 0.10.0",
  "rand_chacha 0.10.0",
  "reqwest",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,7 @@ tokio-util = { version = "0.7", features = ["io"] }
 const_format = "0.2.32"
 iced = { version = "0.15.0-dev", default-features = false }
 nom = "8.0"
+percent-encoding = "2.3.2"
 
 [dependencies]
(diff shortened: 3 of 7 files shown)


furudean (changes requested):
URLs that are naively decoded from a percent-encoding may produce a final result that can be misleading to end users. WHATWG recommends that the decoding described in the Unicode Technical Standard #46 should be used when decoding URLs. Conveniently the [`idna`](https://crates.io/crates/idna) crate would solve this, along other issues for us, which should be a simple swap out from `percent-encode`. This function will only render out an URL if it is not potentially misleading, as defined by TN #46, otherwise the URL will stay as it is.

In addition, this supports domains that have unicode representations are processed. This implementation would follow the exact way that browsers display URLs, which matches user expectations and provides a good bar for safety. With this change, we could also default the added option to true as the risk dramatically decreases for misleading scenarios to occur. We might not even need an option at all. (I can't actually think of many situations where this being set to false is desirable...)

Sources:
- [WHATWG's report on how to process URLs](https://www.unicode.org/reports/tr46/)
- [Unicode Technical Standard #46 - Unicode IDNA Compatibility Processing](https://www.unicode.org/reports/tr46/)

casperstorm (approved):
CHANGELOG? :)

furudean:
inputting `https://bücher.de` into a message seems to result in `https://xn--bcher-kva.de/`. was there a regression? @andymandias

andymandias:
> inputting `https://bücher.de` into a message seems to result in `https://xn--bcher-kva.de/`. was there a regression? @andymandias

That's the punycode-encoded domain per IDNA according to [`Url::host_str`](https://docs.rs/url/latest/url/struct.Url.html#method.host_str), which matches how my browser handles it (converting to puny-code before resolving to https://www.buecher.de/).  I thought that was your suggestion for the domain handling, apologies if I misunderstood.

furudean:
I disagree that this matches browser behavior. Browsers use UTS #46 + UTS #39 as a foundation, and then add some additional heuristics on top (Of which the exact specifics are hard to replicate and are non-standard across browsers). In the current PR we do parse things with `Url::parse` which uses IDNA, but the output of `host_str` is always the punycode form, never the unicode display form. So we are arriving at the opposite of what I originally intended - it always renders punycode. We want something human-readable if it's safe, that should be the end goal of this feature and was also the original ask.

So how do we define safe? We could use [Uts46::to_user_interface](https://docs.rs/idna/latest/idna/uts46/struct.Uts46.html#method.to_user_interface) which lets you provide a closure that decides per-label whether to render as unicode or fall back to punycode. The closure is where UTS #39 comes in  -- a separate but related spec that defines confusable detection and restriction levels.

There are Rust-specific implementation of UTS #39 we could look into. On a quick glance there's the `confusables` crate and `unicode-security`, either of which would work together with the `idna` crate to provide that database.

I could try a PR if there's interest.

andymandias:
By matches browser behavior I mean explicitly this is what happens when I enter a URL into the location bar on my browser, so there's not really something to disagree about there?

The `idna` crate only deals with the domain, so it needs to be applied to the appropriate part of the URL only.  I didn't take a sidetrek for that in a PR that was intended to implement the requested feature of decoding of percent-encoded path components, but if you're interested in doing so please do!

furudean:
Browsers don't always turn the link into punycode, that's the disagreement, maybe? Because this is what the current code does in Halloy through my testing.

andymandias:
I'm sorry, when I wrote "how my browser handles it" then apologized for misunderstanding that wasn't a claim that all browsers everywhere handle it that way.  And I'm not really sure why it would be taken that way, or why it's a sticking point here.  I am fully in support of default conversion to puny-code in cases where potential spoofing is a problem, and if we can get that with less conversion to puny-code I'm in favor of that too.  I was only trying to explain where my confusion came from, not trying to dissuade you from having URLs rendered to your liking.

furudean:
I'm with you. I've put up a PR that handles the host part.

PR #1813, opened 2026-04-11
feat: system tray support

joehonkey:
Adds system tray support so closing the main window hides Halloy instead of quitting. The app keeps running in the background and stays connected.

Left-clicking the tray icon toggles the window. Right-clicking shows a small menu with Show and Quit.

On Linux I used ksni since it speaks the StatusNotifierItem DBus protocol directly, which means left and right clicks are handled separately. Tested on Arch Linux with KDE Plasma on both X11 and Wayland. ksni runs in its own thread with a single-threaded tokio runtime.

On Windows and macOS the tray-icon crate handles things natively. I haven't been able to test those platforms personally so that part would need verifying.

All the tray code is in a new src/tray.rs to keep it out of main.rs. Added ksni as a Linux-only dep and tray-icon for Windows/macOS.


joehonkey:
Sorry, just noticed the AI policy in the contributing guide. Closing this out.

PR #1815, opened 2026-04-11
Filtering linked items for common trailing punctuation and delimiters

englut:
Addresses #1795

This simplifies the regex for matching linked items (url, channel names, nicks).

I've introduced a couple of functions that parse for punctuation at the end of a matched item, then also parses for delimiters (`()`, `[]`, `{}`, `"`, `'`).

All the tests pass. There is one quirk:

`#channel!!!` does parse to `#channel`. Which may be desired, but also makes it harder for channels with an ending punctuation from being linked easily (e.g. `#help?`). I am assuming this may be a rare situation (could be wrong), so I'm leaving this out as I didn't want to even further complicate the delimiter function.

Also added a few more tests to cover more scenarios.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ Fixed:
 - `/MOTD` will not trigger end-of-registration actions
 - Repeat lines in some MOTDs would be deduplicated
 - Received reactions in queries
+- Filtering for common trailing punctuation and delimiters when linking channel names (and nicks & urls)
 
 Removed:
 
--- a/data/src/message.rs
+++ b/data/src/message.rs
@@ -52,7 +52,9 @@ static URL_REGEX: LazyLock<Regex> = LazyLock::new(|| {
 
 // Restrict characters per spec: https://modern.ircdocs.horse/#channels
 static CHANNEL_REGEX: LazyLock<Regex> = LazyLock::new(|| {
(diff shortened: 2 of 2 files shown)


classabbyamp:
`#!/bin/sh` might be a useful channel name for a test case

englut:
> `#!/bin/sh` might be a useful channel name for a test case

That would be funny channel name lol this case is already covered, pushed up the test :)

classabbyamp:
it's the shell scripting channel on libera :)

casperstorm:
Do we want channel matching to include cases like `foo#bar`, where it appears in the middle of a word? 🤔

classabbyamp:
IMO no, that seems like something a user is unlikely to do (and for local chans , it would mean someone mentioning american telecom company `AT&T` could be linkified)

casperstorm:
Agree, but as i understand this change we match on it? Could we add a test for it perhaps and make sure we don't?

englut:
> Agree, but as i understand this change we match on it? Could we add a test for it perhaps and make sure we don't?

It totally does! I just updated the regex to include the look behind and added the foo#bar test :)

casperstorm (approved):
LGTM

andymandias (approved):
Looks good, thank you!

PR #1824, opened 2026-04-13
Focus Reaction Modal Search Query

andymandias:
Focus the search query of the add-reaction modal when it is opened.


casperstorm (approved):
LGTM!

PR #1833, opened 2026-04-14
display urls (almost) like browsers do

furudean:
follow up to #1812. specifically, [this comment](https://github.com/squidowl/halloy/pull/1812#issuecomment-4240624647)

this should match the 2026 status-quo for how browsers parse URLs: rendering the host part as UTS #46 when it passes UTS #39 "Highly Restrictive" level with mixed-script confusable detection. this avoids homograph attacks, like the famous [`www.apple.com` spoof](https://9to5mac.com/2017/04/20/how-to-spot-a-phishing-attempt-fake-apple-site/) and provides a good balance between flexibility and security.

based on research done in https://www.unicode.org/reports/tr39/ and https://unicode.org/reports/tr46/, with cherry-picked points from https://chromium.googlesource.com/chromium/src/+/main/docs/idn.md as browsers have very different security requirements than an IRC client

### DIFF
--- a/data/src/url.rs
+++ b/data/src/url.rs
@@ -194,7 +194,7 @@ pub enum Error {
 /// If hosts don't pass a specific validation criteria, they are displayed as punycode.
 /// The path segment of a URL is always `percent-encoding` decoded.
 ///
-/// We apply a best effort to handle international domain names in a way that matches browser 
+/// We apply a best effort to handle international domain names in a way that matches browser
 /// status-quo; The host is rendered according to UTS #46 when it passes UTS #39 "Highly Restrictive"
 /// level, and its skeleton is not pure ASCII (which guards against homograph attacks).
 ///
--- a/data/src/url.rs
+++ b/data/src/url.rs
@@ -1,7 +1,10 @@
 use std::str::FromStr;
 
 use fancy_regex::Regex;
+use idna::uts46::{AsciiDenyList, Hyphens, Uts46};
 use percent_encoding::percent_decode_str;
+use unicode_security::confusable_detection::skeleton;
+use unicode_security::{RestrictionLevel, RestrictionLevelDetection};
 
 use crate::appearance::theme;
 use crate::server::ServerName;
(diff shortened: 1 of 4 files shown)


data/src/url.rs
```diff
@@ -186,6 +189,47 @@ pub enum Error {
...
+/// We apply a best effort to handle international domain names in a way that matches browser
+/// status-quo; The host is rendered according to UTS #46 when it passes UTS #39 "Highly Restrictive"
+/// level, and its skeleton is not pure ASCII (which guards against homograph attacks).
+///
+/// See https://unicode.org/reports/tr46/ and https://www.unicode.org/reports/tr39/ for
+/// motivations. https://chromium.googlesource.com/chromium/src/+/main/docs/idn.md
+/// is not a bad read either - though note that IRC clients have very different threat
+/// models than browsers.
+///
+pub fn display(u: &url::Url) -> String {
```
andymandias: We could return a `Cow` here to avoid the `u.as_str().to_owned()`.

andymandias (approved):
Thank you!  Looks really good, just one nit.

I assume this means we can remove `percent-encoding` from the base `Cargo.toml` as well.

data/src/url.rs
furudean: do you think that added complexity is worth it? we have to update the call sites to handle the cow. it will always return a new string unless host_str is None. that case will almost never be true for URLs

furudean:
fileupload.rs still uses `percent-encoding`, so i guess we need to keep it

data/src/url.rs
andymandias: `span` already handles it, and I *think* it will be seamlessly supported in the tests.  Should be ~4 line change I believe, but if that's not the case then we don't need it.

furudean: nice. i thought you had to resolve it in fragment but i was wrong upon testing!

andymandias:
Thanks again, appreciate you looking after this!

PR #1834, opened 2026-04-14
IRCv3 `no-implicit-names` support

englut:
Implements [`no-implicit-names`](https://ircv3.net/specs/extensions/no-implicit-names). Added temporary support for `draft/no-implicit-names` since this was ratified recently and I want to make sure we're compatible with soju and ergo both before they've had a chance to upgrade! 

Updated the `/WHO` commands to include filling in the channel's user list. Also, the `/WHO` commands sent on join scenarios have been updated to fire faster when appropriate as well.

~Any keen people want to go make those PRs to soju and ergo??? :eyes:~

[ergo PR](https://github.com/ergochat/ergo/pull/2390) - merged; will be in ergo v2.19
[soju PR](https://codeberg.org/emersion/soju/pulls/388) - merged; will be in soju v0.11.0

### DIFF
--- a/data/src/capabilities.rs
+++ b/data/src/capabilities.rs
@@ -428,14 +428,12 @@ impl Capabilities {
             requested.push("draft/multiline");
         }
 
+        // TODO(quaff): remove `draft/no-implicit-names` support when ergo & soju have both been upgraded
         if self.pending.contains_key("no-implicit-names")
             && !self.acknowledged(Capability::NoImplicitNames)
         {
             requested.push("no-implicit-names");
-        }
-
-        // TODO(quaff): remove `draft/no-implicit-names` support when ergo & soju have both been upgraded
-        if self.pending.contains_key("draft/no-implicit-names")
+        } else if self.pending.contains_key("draft/no-implicit-names")
             && !self.acknowledged(Capability::NoImplicitNames)
         {
             requested.push("draft/no-implicit-names");
--- a/data/src/capabilities.rs
+++ b/data/src/capabilities.rs
@@ -35,6 +35,7 @@ pub enum Capability {
     Multiline,
     MultiPrefix,
     Metadata,
+    NoImplicitNames,
     ReadMarker,
     Sasl,
     ServerTime,
@@ -64,6 +65,9 @@ impl FromStr for Capability {
             "draft/message-redaction" => Ok(Self::MessageRedaction),
             "multi-prefix" => Ok(Self::MultiPrefix),
             "draft/metadata-2" => Ok(Self::Metadata),
+            "no-implicit-names" => Ok(Self::NoImplicitNames),
+            // TODO(quaff): remove `draft/no-implicit-names` support when ergo & soju have both been upgraded
+            "draft/no-implicit-names" => Ok(Self::NoImplicitNames),
             "server-time" => Ok(Self::ServerTime),
             "setname" => Ok(Self::Setname),
             "soju.im/bouncer-networks" => Ok(Self::BouncerNetworks),
@@ -422,6 +426,19 @@ impl Capabilities {
             && MultilineLimits::from_str(multiline).is_ok()
         {
             requested.push("draft/multiline");
+        }
+
+        if self.pending.contains_key("no-implicit-names")
+            && !self.acknowledged(Capability::NoImplicitNames)
+        {
+            requested.push("no-implicit-names");
+        }
+
+        // TODO(quaff): remove `draft/no-implicit-names` support when ergo & soju have both been upgraded
+        if self.pending.contains_key("draft/no-implicit-names")
+            && !self.acknowledged(Capability::NoImplicitNames)
+        {
+            requested.push("draft/no-implicit-names");
         }
 
         if self.pending.contains_key("draft/metadata-2")
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -1944,10 +1944,16 @@ impl Client {
                         self.statusmsg(),
                         casemapping,
                     ));
-                    if let Some(pos) = self
-                        .who_polls
-                        .iter()
-                        .position(|who_poll| who_poll.channel == target_channel)
+                    if let Some(pos) =
+                        self.who_polls.iter().position(|who_poll| {
+                            who_poll.channel == target_channel
+                                && (matches!(
+                                    who_poll.status,
+                                    WhoStatus::Joined
+                                        | WhoStatus::Received
+                                        | WhoStatus::Waiting(_)
+                                ))
+                        })
                     {
                         self.who_polls.remove(pos);
                     }
@@ -2082,10 +2088,6 @@ impl Client {
             }
             Command::Numeric(RPL_WHOREPLY, args) => {
                 let channel = ok!(args.get(1));
-                let user = ok!(args.get(2));
-                let host = ok!(args.get(3));
-                let nick = ok!(args.get(5));
-                let flags = ok!(args.get(6));
 
                 let casemapping = self.casemapping();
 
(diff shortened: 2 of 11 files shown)


data/src/capabilities.rs
```diff
@@ -424,6 +428,19 @@ impl Capabilities {
...
+            requested.push("no-implicit-names");
+        }
+
+        // TODO(quaff): remove `draft/no-implicit-names` support when ergo & soju have both been upgraded
+        if self.pending.contains_key("draft/no-implicit-names")
+            && !self.acknowledged(Capability::NoImplicitNames)
+        {
+            requested.push("draft/no-implicit-names");
+        }
+
```
andymandias, line 441: We should probably only request one or the other.  Let's put `draft/no-implicit-names` on an `else` branch off of the `no-implicit-names` conditional.

data/src/client.rs
```diff
@@ -1750,6 +1751,21 @@ impl Client {
...
+                        self.statusmsg(),
+                        casemapping,
+                    ));
+                    if let Some(pos) = self
+                        .who_polls
+                        .iter()
+                        .position(|who_poll| who_poll.channel == target_channel)
+                    {
+                        self.who_polls.remove(pos);
+                    }
```
andymandias: Since the `who_polls` state controls how WHO replies are processed, we should remove the poll for the departed channel only if it's not in progress.

data/src/isupport.rs
```diff
@@ -1137,8 +1137,8 @@ pub enum WhoXPollParameters {
 impl WhoXPollParameters {
     pub fn fields(&self) -> &'static str {
         match self {
-            WhoXPollParameters::Default => "tcnf",
-            WhoXPollParameters::WithAccountName => "tcnfa",
+            WhoXPollParameters::Default => "tcnfuh",
+            WhoXPollParameters::WithAccountName => "tcnfuha",
```
andymandias: I believe we only need `tcnf`/`tcnfa` for regular polls, while `tcnfuh`/`tcnfuha` are necessary for the initial polls.  It's a small optimization, but it would be nice to have as only as much data transferred as necessary.

data/src/client.rs
```diff
@@ -1933,7 +1978,13 @@ impl Client {
                 }
             }
             Command::Numeric(RPL_WHOSPCRPL, args) => {
+                let token = ok!(args.get(1));
                 let channel = ok!(args.get(2));
+                let user = ok!(args.get(3));
+                let host = ok!(args.get(4));
+                let nick = ok!(args.get(5));
+                let flags = ok!(args.get(6));
+                let account = args.get(7);
```
andymandias: If the user sent the WHOX request then it may or may not have this mapping between `args` and fields.  Since `ok!` will trigger an early return that means user sent WHOX polls may not be handled correctly.  We should probably map after we confirm (as best we can) that its a reply to an automatic WHO poll.

andymandias (commented):
This has a lot of good ideas in it, and the optimization of WHO polls needed for initial state will be really nice to have.

I think the WHO poll logic is now sufficiently complicated, and no longer operates as a straightforward `VecDeque`, that we should codify it with a `struct` and `impl` (and move it out of the `client` module).

data/src/client.rs
englut: Hm. I can add more strict status checking for the prioritize_joined_who_polls() function to supplement. ~~but I think I need a check in the handlers for `WhoStatus::Received` and `WhoStatus::Waiting` to remove itself from the queue if the user isn't in the target channel, does that sound right?~~ Nevermind! That's already happening! Should be good to go.

data/src/isupport.rs
englut: Oh interesting. only get this state on initial poll... that makes sense! no unused data. Will adapt to this as well :pray:

data/src/capabilities.rs
englut: Good call! Moved to `else` instead of two `if`s! :pray:

data/src/client.rs
englut: Reworked this a bit, and included a `InitialJoin` WHOX param that is only passed and used on initial joins. Now reads the args at use again!

data/src/isupport.rs
englut: Updated this with new `InitialJoin` WHOX param. Non init who polls will now use the `tcnf`/`tcnfa`. As far as I can tell.. we only need `tcnfuha` for the initial join's who poll, right? Shouldn't need both?

englut:
Reorganized the who poll related logic out of `data/src/client.rs` and into `data/src/client/who_queue.rs`. The PR jumped in size :joy: but things feel cleaner for me too. I've also addressed all the other comments @andymandias :pray:  Thanks again as usual for your great feedback!

data/src/client/who_queue.rs
```diff
@@ -413,6 +402,7 @@ impl WhoQueue {
             if let Some(who_poll) = self.polls.remove(pos)
                 && chanmap.contains_key(channel)
                 && config.who_poll_enabled
+                && !capabilities.acknowledged(Capability::AwayNotify)
```
englut, line 566: From what I can tell. Preventing the who_poll from being added back to the queue if the server is cap acked `AwayNotify`, will save us from having to check for who_init (if cap acked `AwayNotify`) in the `WhoStatus::Waiting` arm of the tick match. 

This was causing a weird state where if you cycle through a bunch of channel buffers on initial load, all the channels that you prioritized but didn't allow to fully complete their who polls would get stuck in a state (`Waiting` + `who_init = true`, and no Nick list populated or `WHO done` trace 🤔), which meant the nicklist wouldn't populate via tick until you opened the buffer again.

data/src/client/who_queue.rs
```diff
@@ -17,336 +17,294 @@ use crate::{User, client, config, message};
...
+        isupport: &HashMap<isupport::Kind, isupport::Parameter>,
+        now: &Instant,
+    ) -> Vec<(message::Encoded, TokenPriority)> {
+        let in_flight_is_empty = self.in_flight.is_empty();
+        let interval_duration = self.interval.duration();
+        let reference_time = self.reference_time;
+
+        // Find any timed out polls and return them in the queue.
+
+        let timed_out_polls: Vec<_> = self
```
englut: I think this should be named `overdue_polls` or something to that effect. With `timed_out_polls`, it's not immediately clear that these are polls that should be moved to the front to be triggered immediately.

data/src/client/who_queue.rs
```diff
@@ -17,336 +17,294 @@ use crate::{User, client, config, message};
...
+                channel: channel.clone(),
+                source: WhoSource::Join { priority: false },
+                status: WhoStatus::Waiting {
+                    immediate: capabilities
+                        .acknowledged(Capability::NoImplicitNames),
+                },
+            }
+        };
+
+        self.insert_towards_back(who_poll);
```
englut, line 311: I know when the user opens the buffer, this will get prioritized, but maybe we should use `insert_towards_front` so that it's taken care of faster, since a join can be due to a user's interaction? 🤔

englut (commented):
Hot damn. This is so much easier to follow now! Thank you for taking the time to go through all this! 

I have minor minor nits. Other than that, I think this makes sense, and it runs well too. I'm happy with it!

data/src/client/who_queue.rs
```diff
@@ -0,0 +1,721 @@
...
+                            nick,
+                            accountname,
+                            casemapping,
+                        );
+                    } else if token == WhoXPollParameters::InitialJoin.token() {
+                        let flags = ok!(args.get(6));
+                        let nick = ok!(args.get(5));
+                        let username = ok!(args.get(3));
+                        let hostname = ok!(args.get(4));
+                        let accountname = ok!(args.get(7));
```
luca020400, line 511: I think InitialJoin handling is broken here.

InitialJoin => "tcnfuha"

```suggestion
                        let nick = ok!(args.get(3));
                        let flags = ok!(args.get(4));
                        let username = ok!(args.get(5));
                        let hostname = ok!(args.get(6));
                        let accountname = ok!(args.get(7));
```

data/src/client/who_queue.rs
```diff
@@ -0,0 +1,721 @@
...
+
+                        let user = User::from_whoreply(
+                            nick,
+                            flags,
+                            username,
+                            hostname,
+                            Some(accountname),
+                            casemapping,
+                            isupport::get_bot_mode_char(isupport),
```
luca020400:
```suggestion
                            bot_mode_char,
```

data/src/client/who_queue.rs
luca020400: Oh nvm, the order is spec defined not request defined.
https://ircv3.net/specs/extensions/whox#rpl_whospcrpl-354-numeric-reply

data/src/client/who_queue.rs
englut: ah right, it's inserted by source. hmmmm yeah actually I think this is fine, cause other joins by the User source should happen first 🙏

andymandias (approved):
Excellent work. Thank you for sticking with it 🙏

englut:
> Excellent work. Thank you for sticking with it 🙏

@andymandias Thank YOU for holding me hand through this and helping make sure the code is future looking and efficient :pray:  And of course all your direct changes!

PR #1836, opened 2026-04-14
Moved `typing` headings one level up to align with config.

englut:
This PR moves the `typing` section was at the correct level in the `servers` config.

| Before | After |
|--------|--------|
| Screenshot_20260414_110506 | Screenshot_20260414_110435 | 

*hint: it was under the `filehost` heading (or the `confirm_message_delivery` in the current release)


casperstorm (approved):
Thanks @englut!

PR #1838, opened 2026-04-14
filehost: show body on http error

furudean:
show error reason in buffer, if any

like 

```
HTTP POST https://files.milkmedicine.net/upload returned 500 Internal Server Error: mkdirat user/meri: permission denied.
```


andymandias (approved):

andymandias:
Thanks!

PR #1839, opened 2026-04-14
Expand Tooltip Setting

### LINKED ISSUE #1837: tooltips = false appears to be ignored
I have "tooltips = false" as the first item in my config file and I've reloaded my config, however I still see tooltips whenever I begin a message with "/".

Image

andymandias:
Expands tooltip setting into an enum to allow more detailed control of which tooltips are displayed and allow autocomplete tooltips to be hidden.

Fixes #1837.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Added:
 - Unread indicators for queries can be hidden via `sidebar.unread_indicator.exclude`
 - Settings to reroute direct PRIVMSG and/or NOTICE messages to another buffer (`servers.<name>.reroute.query` and `servers.<name>.reroute.notice`)
 - `channel-context` support
+- Expanded `tooltips` setting to allow hiding auto-complete tooltips
 
 Changed:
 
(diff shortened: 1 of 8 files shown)


casperstorm (approved):

englut:
Should this setting apply to the truncated nicknames & react emoji tooltips as well?

andymandias:
> Should this setting apply to the truncated nicknames & react emoji tooltips as well?

This formulation will expand readily into an array of strings (`Vec<Enum>`) if there is desire for that (which I don't currently expect, but I could be off base).  If there is such desire then reactions in particular will need additional work, since the tooltip provides data that is not currently visible via any other mechanism (as such I'd prefer it be handled in a separate PR regardless).

andymandias:
> Should this setting apply to the truncated nicknames & react emoji tooltips as well?

If this is your way of hinting that you'd like those options, we can definitely add truncated nickname tooltip suppression to this PR :)

englut:
> > Should this setting apply to the truncated nicknames & react emoji tooltips as well?
> 
> If this is your way of hinting that you'd like those options, we can definitely add truncated nickname tooltip suppression to this PR :)

LOL I admit I can have passive aggressive tendencies. but no, it was just a tooltip I noticed. I wasn't sure what was intended. Thinking on it more, not having tooltips on truncated nicknames and react emojis is losing access to data. Which isn't great UX. So for me personally, I like what you have in this PR. 

I think us discussing here is good enough of a paper trail to point back to if it ever gets asked.

PR #1841, opened 2026-04-15
filehost: only support `soju.im/filehost`

furudean:
based on conversation in libera #halloy, we agreed that `draft/filehost` is too unstable to advertise support for. the spec has changed since support was added and we don't support it in its current form. it's problematic to advertise that we support a spec that no longer really exists, and especially since is not even fully merged as an official draft spec. `soju.im/filehost` on the other hand has been stable for a long time, and is unlikely to change.

most ircd setups i've seen advertise both soju.im/filehost and draft/filehost as they used to overlap, so it should not be much be much of a functional change.


casperstorm (approved):
I think this is the right approach.

PR #1846, opened 2026-04-17
IRCv3 draft/message-redaction support

luca020400:
The standard doesn't expect the server to send any information
about the "redaction" policy, but at least on Ergo you can redact
messages if you sent a message or if you're an OP in a channel.

Thus for the sake of simplicity the client implements the same
behavior.

On redact a modal will open where the user can enter the redact reason (or none)

On recv of REDACT we search history for the matching entry, and
mark it as such. UI side it'll be dimmed.


luca020400:
image

My UI capabilities end here, the spec https://ircv3.net/specs/extensions/message-redaction#client-implementation-considerations recommends various way to handle the scenario, I went with the easiest solution for now.

furudean:
how about a tooltip when hovering a redacted message that shows the reason?

luca020400:
> how about a tooltip when hovering a redacted message that shows the reason?

image

Looks nicer ty!

casperstorm:
> > how about a tooltip when hovering a redacted message that shows the reason?
> 
> image
> 
> Looks nicer ty!

This is amazing, good work.

furudean:
Does it make sense to only show redact to ops? I don't think there is a straightforward way to know whether or not a user can redact a given message. It would depend on the IRCd, and maybe even chanops are able to redact in some cases. 

Should we always show the menu and just error if it's not allowed?

luca020400:
> Does it make sense to only show redact to ops? I don't think there is a straightforward way to know whether or not a user can redact a given message. It would depend on the IRCd, and maybe even chanops are able to redact in some cases.
> 
> Should we always show the menu and just error if it's not allowed?

That's one of the main pain points of this functionality, the spec leaves the redact policy up to implementers.

`clients MAY assume that their own messages are redactable. <snip>. Pending a mechanism for discovering redaction permissions, clients SHOULD allow users to attempt to delete any message via some mechanism.`

We can always allow redaction, errors are sent back via the FAIL cmd so they are user-discoverable.

I followed whatever logic Ergo had on their server impl, on Goguma for example they only allow own messages to be redacted.

furudean:
I think we should not assume whether or not a message is redactable without a specific indication from the server. The neutral middle ground might look like to always assume its allowed, but reject invalid attempts as returned by the server.

furudean:
I think oper is not a poor assumption for most IRCDs, but I think there can also be a lot of false positives and false negatives with the current implementation.

andymandias (approved):

casperstorm (approved):
Great work

PR #1848, opened 2026-04-17
buffer: add readline-like ^u and ^w to emacs bindings

okdana:
would you consider adding ^u for backward kill line and ^w for backward kill word to the emacs bindings, like this?

i know these are technically not bindings provided by emacs itself, but they're used in the default 'emacs-style' configuration of readline, editline, and zle (and thus most shells, repls, sql clients, etc). i'm very used to them because of that

(if it's too problematic to call these 'emacs', maybe a new set of bindings called 'readline' that's just a super-set of it?)


casperstorm (approved):
I think this is reasonable! Do you mind adding a small entry to CHANGELOG?

okdana:
sry, done that now

PR #1849, opened 2026-04-17
buffer: add kill_to_clipboard setting to control clipboard behaviour

okdana:
related to but independent of #1848

certain bindings like ^k simulate emacs by moving killed text to the clipboard. if you've mostly used these bindings in the terminal (which doesn't usually integrate with the clipboard by default, at least on macos), this can be very weird and disruptive

is it reasonable to have a setting to disable it, like this?


casperstorm (approved):

casperstorm:
I like this as well, thank you for both contributions, @okdana.

PR #1787, opened 2026-04-08
Implement ircv3 metadata

4e554c4c:
This commit implements the bare minimum for metadata to work in halloy. As an example, metadata info is routed to the view so that a user's display name is viewable. Currently only display-name is requested so we don't accidently run into limits when trying to request other keys.

Resolves #1606

PR #1853, opened 2026-04-19
render IRCv3 bot mode + allow truncating nicklist

### LINKED ISSUE #1831: Allow truncating nicks in channel nicklist
I'd like to request truncating the nicklist in channel buffers. Similar to [buffer.nickname.truncate](https://halloy.chat/configuration/buffer#truncate), or maybe we can re-use this config to apply the truncating to the channel buffer's nicklist as well? 🤔 

Another idea is adding `buffer.channel.nicklist.truncate` and apply truncating styles if [buffer.channel.nicklist.width](https://halloy.chat/configuration/buffer#width) is set. 

I'm in a channel where the nicklist has really long names and it's kind of ugly. Setting `buffer.channel.nicklist.width` breaks the long nicks onto two lines, I'd much prefer the truncate style that we have with `buffer.nickname.truncate` 🙏

furudean:
1. users with +b mode has a 🤖 icon next to their name. the icon can be hovered to show a tooltip with the meaning
2.  render numeric in /whois
3. also fixes #1831, scored a quick win here since i was already messing with the width calculation

1-2 are as per https://ircv3.net/specs/extensions/bot-mode

Screenshot 2026-04-20 at 12 58 00 PM


furudean:
found a hiccup with right-aligned nicklists. not 100% sure how to solve it yet. in html/css you can wrap things in columns to make them align together as blocks on rows, but i'm not sure this applies here because we are mixing text with other elements. so it's not a simple question of aligning text anymore.
 
Screenshot 2026-04-19 at 12 27 57 PM

furudean:
fixed. i don't love having to do all of this though. iced just doesn't handle text next to other elements very well and you end up in situations where you have to set a fixed width and then calculate said width. i'm genuinely not sure if there's a workaround

a minimal implementation of bot mode would be to not show an icon and just add the `/whois` line but 🤷‍♀️

furudean:
did another pass on this that uses the icon font instead, which greatly shrinks down how much logic there needs to be. kind of realizing why we rely so heavily on icon fonts now :-D

src/buffer/channel.rs
```diff
@@ -591,25 +599,86 @@ mod nick_list {
...
+                    theme, config, user,
+                )
+                .color;
+                let bot_tooltip =
+                    format!("{} has marked itself as a bot", user.nickname());
+                let bot_icon = tooltip(
+                    icon::robot().style(move |_| iced::widget::text::Style {
+                        color: nick_color,
+                    }),
+                    Some(bot_tooltip),
```
andymandias: I believe @casperstorm is making changes in how user metadata is displayed in #1787, so this tooltip may be subsumed by those changes.  But if not, then we should allow for it to be configured via:
```rust
self.config.tooltips.show_for_buttons().then_some(bot_tooltip)
```

(And in that case that function should probably be renamed to something like `show_for_descriptions`.)

src/buffer/message_view.rs
```diff
@@ -412,13 +412,23 @@ impl<'a> ChannelQueryLayout<'a> {
         let nick_text =
             self.config.buffer.nickname.brackets.format(user_display);
```
andymandias: It will be a bit of a pain, but existing style is for the bot icon to be inside the brackets.  I expect that means we'll have to branch on displaying the bot icon here.

andymandias (commented):
Excellent work.  A couple display nits, but these are very welcome additions that I'm looking forward to having merged.  The bot icon is cute 🙂  Thank you!

src/buffer/message_view.rs
furudean: There was an existing style? I might be a little confused on what you mean exactly, could you elaborate on what the ask is here?

src/buffer/channel.rs
furudean: Should we hold of on merging this until #1787 is in? Or is this just a heads up

src/buffer/message_view.rs
andymandias: Sorry, that was definitely clearer inside my head than in the actual comment.  Primarily I'm thinking of channel membership prefixes (`@`, `+`, etc), which all currently get placed inside the brackets (e.g. `<+andymandias>` or `<@oats>` rather than `@<casperstorm>`).  Along those lines then, `<andymandias🤖>` rather than `<andymandias>🤖`.

I believe there are also folks who utilize the right bracket with the assumption that it is the right-most glyph before the message, with a `:` or `|` as the "right bracket".  `alis🤖: <message>` signals more clearly as "alis the bot says" to me than `alis:🤖 <message>`.

furudean: I had not tested this with different formatting setups, but that totally makes sense. I usually just run the bare nick, so I had overlooked that. I overall agree with you that they should be included as part of the nick string, immediately before any pre-message glyph.

src/buffer/channel.rs
andymandias: A heads up, and invitation for casper to weigh in if they have a need.  No need to hold off on merging that I know of.

src/buffer/message_view.rs
furudean: i have pushed some <sup>terrible</sup> code to account for brackets

furudean:
brackets before and after

Screenshot 2026-04-24 at 5 29 36 PM
Screenshot 2026-04-24 at 5 29 34 PM

right-aligned chat is not so lucky, now having extra space reserved for an eventual bot icon compared to before. i experimented with some alternatives that chop of additional characters from truncate to make the overall nick size shrink but it was a mess. not sure if there is a good solution here...

Screenshot 2026-04-24 at 5 31 11 PM
Screenshot 2026-04-24 at 5 31 18 PM

andymandias (approved):
Looks great, thank you!

casperstorm (approved):
Thank you!

englut:
Wooo two awesome features, great work @furudean 😍

PR #1861, opened 2026-04-21
Rich query pane title

### LINKED ISSUE #1840: Display user's status in query buffer
Would be good to indicate if a user is online or away and if they're authenticated/unauthenticated (probably helps with spam & privacy) in a query buffer.

I think from a UX perspective, it'd be a quick and easy way to see if the user is offline or if they're unauthed (and I should just ignore them).

An example of how it's displayed in gamja:

Image
Image

We've already got the status bar at the top showing the query nick & server:

Image

Could add the status text(s) (or icon(s) if we want to be fancy 💁‍♀️)  next to the nick some how.

casperstorm:
fixes #1840 

Authenticated and not away
Screenshot 2026-04-21 at 23 05 19

Aythenticated and away
Screenshot 2026-04-21 at 23 05 37

Not authenticated and not away
Screenshot 2026-04-21 at 23 13 26


andymandias (approved):
Looks really nice!

PR #1868, opened 2026-04-22
Add `font.stretch` config

bb010g:
I have an Iosevka Aile Semi-condensed font that I couldn't use without this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Unreleased
 
+Added:
+
+- Configuration options for font stretch (`font.stretch`)
+
+Thanks:
+
+- Contributions: @bb010g
+
 # 2026.6 (2026-04-21)
 
 Added:
--- a/data/src/config.rs
+++ b/data/src/config.rs
@@ -224,6 +224,8 @@ impl Default for Scrollbar {
 #[serde(default)]
 pub struct Font {
     pub family: Option<String>,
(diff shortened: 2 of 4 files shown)


andymandias (commented):
This looks good to me, but none of the fonts I have appear to be affected by the setting.  I may be missing something obvious, but I don't see a semi-condensed when I extract the IosevkaAile zip from [the latest release](https://github.com/be5invis/Iosevka/releases).  Is there a publicly available font to test with?

bb010g:
> Is there a publicly available font to test with?

@andymandias Iosevka Aile ships with a condensed that should work.

casperstorm:
I installed the Iosevka Aile for MacOS, and it came with the following. 
However i am unsure what the stretch changes? I don't see any difference. Perhaps i am doing something wrong.  Screenshot 2026-04-23 at 09 06 53


```toml
[font]
family = "Iosevka Aile"
stretch = "semi-expanded"
```

bb010g:
Sorry; forgot that (semi-)condensed widths are entirely relegated to custom builds. I can confirm that stock Iosevka ships with an expanded width (“Iosevka Extended”) though. First screenie is `"normal"`, second is `"expanded"`:

Screenshot_20260423_063138

Screenshot_20260423_063227

casperstorm (approved):
Thanks for the explanation :)

PR #1869, opened 2026-04-22
Add default feature `bundled-default-font`

bb010g:
I would prefer to use my local Iosevka Term build.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+Added:
+
+- The default font is not bundled if default feature `iosevka-font` is disabled.
+
 Fixed:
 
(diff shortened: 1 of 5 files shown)


casperstorm (approved):
Thanks! I did a few changes to make sure we don't load iosevka if the feature is off. I also renamed it to `iosevka-font` so we can add more fonts in the future.

andymandias (approved):

bb010g:
> Thanks! I did a few changes to make sure we don't load iosevka if the feature is off.

Thank you! They all look good.

> I also renamed it to `iosevka-font` so we can add more fonts in the future.

I initially drafted that, but went for a more generic name to try and make the feature list simpler for those not enabling the `default` feature. I don't know & haven't researched Halloy's approach to crate feature stability, though.

PR #1871, opened 2026-04-23
support soju.im/bouncer-networks-notify CAP

furudean:
fixes #1797

### DIFF
--- a/data/src/bouncer.rs
+++ b/data/src/bouncer.rs
@@ -41,6 +41,10 @@ impl FromStr for NetworkState {
 pub struct BouncerNetwork {
     pub id: String,
     pub name: String,
+    #[serde(default)]
+    pub state: NetworkState,
+    #[serde(default)]
+    pub error: Option<String>,
 }
 
 // for ordering, we try to order lexiographically by name, and then check ID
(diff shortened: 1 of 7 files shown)


data/src/capabilities.rs
```diff
@@ -339,6 +343,12 @@ impl Capabilities {
             requested.push("soju.im/bouncer-networks");
         }
 
+        if self.pending.contains("soju.im/bouncer-networks-notify")
```
4e554c4c: shouldn’t this cap only be requested on the dedicated bouncer connection?

furudean: yes, the same is also true for soju.im/bouncer-networks above. should we add a `self.is_primary()` guard to both to be defensive?

4e554c4c: no, you need `soju.im/bouncer-networks` for the `BOUNCER` command, including `BOUNCER BIND` which is used on child connections. when i was originally designing this i kept around a primary (unbound) connection that's not used for anything, but could be used for bouncer-networks-notify, and individual upstream connections which (correctly) use `soju.im/bouncer-networks`

furudean: that makes sense. i think it's implemented correctly now

data/src/capabilities.rs
```diff
@@ -339,6 +344,13 @@ impl Capabilities {
             requested.push("soju.im/bouncer-networks");
         }
 
+        if is_primary
```
4e554c4c, line 347: `is_primary` is not fully resolved at this point (since we probably haven't received an ISUPPORT with the `BOUNCER_NETID=blah` token).

That means that if i sets up two networks with
`pounce/libera` and `pounce/oftc`, then both will request `soju.im/bouncer-networks-notify`. Will this cause problems?

data/src/client.rs
```diff
@@ -185,6 +186,7 @@ pub struct Client {
     who_polls: VecDeque<WhoPoll>,
     who_poll_interval: BackoffInterval,
     resolved_netid: Option<String>,
+    bouncer_networks: HashMap<String, BouncerNetwork>,
```
4e554c4c, line 189: maybe this can be an `Arc<BouncerNetwork>` like the other uses, and can be modified with `Arc::make_mut` if necessary

data/src/bouncer.rs
```diff
@@ -41,6 +41,10 @@ impl FromStr for NetworkState {
 pub struct BouncerNetwork {
     pub id: String,
     pub name: String,
+    #[serde(default)]
```
4e554c4c, line 44: I don't know how I feel about adding two extra fields to `BouncerNetwork`

`BouncerNetwork` is stored in `Server`, which is an _identifier_ that can't be changed (because it's `cloned()` all over the codebase) so `BouncerNetwork` is supposed to be an ID.  The reason I included `name` is _kind_ of a hack, since `Server` also needs to know how to pretty-print itself

Could we maybe split this into `BouncerNetwork` and `BouncerNetworkState` or similar? so that we don't have a bunch of nonconfluent copies of the same `BouncerNetwork`/`Server` in memory?

furudean: To your first point, do you think changing the identifier from the name to a randomly assigned one would help mitigate the confusion?  The name doesn't make for a great identifier as it may be changed during a session.

data/src/capabilities.rs
furudean: oof. it probably will cause problems

furudean:
i'm not sure how to implement this, so i'm closing this pull request for now

PR #1873, opened 2026-04-23
Iteratively decode open graph entries

### LINKED ISSUE #1847: Preview doesn't support HTML encoded text
When an URL is pasted and the preview is enabled it attempts to fetch metadata to fill in the preview card.

The entries in `og:*` are HTML encoded, but currently we don't decode any of the sequences when parsing the content.

Image

luca020400:
We'll break as soon the original input is returned
or we reach N=4 attempts at stabilzing the input.

As seen in the wild:
`doesn&amp;#39;t` -> `doesn&#39;t` -> `doesn't`

Fixes #1847

### DIFF
--- a/data/src/preview.rs
+++ b/data/src/preview.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::collections::HashMap;
 use std::io;
 use std::sync::{Arc, LazyLock, OnceLock};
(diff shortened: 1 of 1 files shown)


casperstorm (commented):
Looks great, i feel its worth to have a test for this fn at this point?

furudean:
decoding the content is fine, but iteratively? that follows no known specification. the standard way to handle html tag content is to call [decodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent) on it a single time. if the content is still in an escaped form, the markup is incorrect.

decoding content multiple times is off-spec and has the potential to break things where people are genuinely trying to write out an escaped sequence.if someone legitimately wants to display the literal string `&amp;` they'd write `&amp;amp;` in html, and the loop would turn that into `&`, eating their intentional content.

PR #1877, opened 2026-04-25
Fix `tor` enabled builds

englut:
Builds with `--all-features` or the flag to enable just `tor` are failing in the current release.

image

Looks like there was a change that ignored the `tor` arm of the match. I've made some changes to get the build going again.

Distros (like [archlinux](https://gitlab.archlinux.org/archlinux/packaging/packages/halloy/-/commit/1008cb85650902e23c920242a7570546fc90043d)) that build with `--all-features` are likely failing due to this bug.


andymandias (approved):
Thank you!

PR #1878, opened 2026-04-25
Show Server Messages in Labeled Context Buffer

andymandias:
Show server messages in labeled context buffer by default, when available.  In particular, to show standard replies in the buffer that the command was sent from (e.g. if attempting and failing to redact a message in a buffer, then the standard reply `FAIL` will show in that buffer).

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -29,7 +29,6 @@ use crate::isupport::{
     ChatHistoryState, ChatHistorySubcommand, MessageReference, WhoToken,
     WhoXPollParameters, find_target_limit,
 };
-use crate::message::source;
 use crate::rate_limit::{BackoffInterval, TokenBucket, TokenPriority};
 use crate::target::{self, Target};
 use crate::time::Posix;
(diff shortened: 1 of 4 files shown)


data/src/client.rs
```diff
@@ -131,7 +130,7 @@ pub enum Event {
     WithTarget {
         message: message::Encoded,
         our_nick: Nick,
-        target: message::Target,
+        target: Option<Target>, // None → Server
```
casperstorm: We could have a small `Destination` enum:

```rust
pub enum Destination {
    Server,
    Target(target::Target),
}

// ...

impl From<&buffer::Upstream> for Destination {
    fn from(buffer: &buffer::Upstream) -> Self {
        match buffer.target() {
            Some(target) => Self::Target(target),
            None => Self::Server,
        }
    }
}

// ...

destination: Destination::from(&buffer),
```

andymandias: Good idea!

casperstorm (approved):

PR #1881, opened 2026-04-25
bot mode: update own indicator when `/mode`

furudean:
when calling `/mode nick +B` on self, the indicator did not update

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -2182,6 +2182,31 @@ impl Client {
                         }
                     }
                 } else {
+                    let modes = mode::parse::<mode::User>(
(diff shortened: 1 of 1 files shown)


furudean:
one thing i noticed is that if you have `/samode <nick> +B` called on you, your local bot status won't update. as far as i know, there is no way to subscribe to an update about your modes except if you set them yourself or poll with `WHO`. not great solutions. any advice is welcome, but does not need to be solved in this pull

andymandias (approved):
I am not familiar with the mechanics of `/samode`, but this makes sense to me for `/mode`.  Thank you!

PR #1882, opened 2026-04-25
Truncation Tweaks Re: Bot Icon

andymandias:
Treats bot icon as one character for the purposes of truncation (to match the treatment of access levels), and makes buffer and nick list handling of truncation more consistent.


furudean (approved):
looks good!

casperstorm (approved):
Clean code as always.

PR #1883, opened 2026-04-26
Distinguish between reacts in chathistory vs active history

englut:
Chathistory reacts should not send notifications.


andymandias (approved):
Looks good to me, works well in my testing.  Thank you!

PR #1884, opened 2026-04-26
Truncation Character Setting

andymandias:
Adds a setting to customize the character used to indicate a nickname was truncated.


WinnerWind:
image
Here i have it set to a single period `.`
works good!

casperstorm (approved):
Looks great.

PR #1889, opened 2026-04-26
Store Bot Mode

andymandias:
Serialize a user's bot mode so it can be stored with a message, providing historical context for bot messages independent of current bot status.

Fixes regression where displayed access levels would be the user's current access level(s), rather than their access level(s) at the time of sending the message.


furudean:
great change!

casperstorm (approved):

PR #1892, opened 2026-04-27
History improvements

luca020400:
- **Consume pending_{reactions,redactions} instead of cloning them**
- **Force history flush after 60 seconds**


andymandias (approved):
Thank you!

PR #1895, opened 2026-04-27
Pane Title Tweaks

andymandias:
Adds coloring to channel and server pane titles to help differentiate panes, and in the case of channels highlight the channel title over the channel properties.


casperstorm (approved):

casperstorm:
Looks great 👍🏻

PR #1896, opened 2026-04-27
Enable wayland paste + file paste

luca020400:
- **Enable wayland support in arboard**
- **Handle clipboard file uploads**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -234,6 +234,7 @@ dependencies = [
  "parking_lot",
  "percent-encoding",
  "windows-sys 0.60.2",
+ "wl-clipboard-rs",
  "x11rb",
 ]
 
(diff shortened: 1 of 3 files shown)


luca020400:
Needs to ensure windows works fine, especially given the \r thingy I see on here.

Maybe the macos specific handling isn't needed with that?

andymandias (approved):
This look good, works well for me!  I have filed a bug with arboard regarding the trailing `\r` since I believe it should not be doing that (https://github.com/1Password/arboard/issues/216).

I agree with @luca020400 that testing with Windows & macOS is needed.

luca020400:
Upstream bug has been fixed, all that's left is testing on other platforms.

casperstorm:
When i copy two files on macOS (cmd+c) and paste (cmd+v) into the buffer it only creates a upload for a single one of them. Is that intentional?

luca020400:
> When i copy two files on macOS (cmd+c) and paste (cmd+v) into the buffer it only creates a upload for a single one of them. Is that intentional?

I think yes, since there's specific macOS handling right above, one could try to remove it and see what happens :)

casperstorm:
Gotcha, for the time being i think this is good enough. Good work on this!

furudean:
Late to the party here but only one file for Mac file paste was ever supported. Improvements to that implementation is welcome in the future!

luca020400:
> Late to the party here but only one file for Mac file paste was ever supported. Improvements to that implementation is welcome in the future!

In arboard they have mac os specific handling for file_list and image, maybe it's worth a shot.
But I assume that was broken since you went with a custom impl to begin with?

furudean:
I couldn't get arboard to work on my end with a Mac, feel free to take a chop at it though. I'm in the IRC if you want to chat more about this :)

PR #1897, opened 2026-04-28
Display Redacted Messages as Redactions

andymandias:
Adds `"redacted"` variant to `buffer.redaction.display` which displays redacted messages as their redaction reason (and redactor), with the original, redacted message visible by clicking on the redaction reason.

### DIFF
--- a/data/src/config/buffer/redaction.rs
+++ b/data/src/config/buffer/redaction.rs
@@ -11,11 +11,16 @@ pub struct Redaction {
 pub enum Display {
     #[default]
     None,
+    Redacted,
     Dimmed,
 }
 
 impl Display {
-    pub fn is_dimmed(self) -> bool {
-        matches!(self, Self::Dimmed)
+    pub fn is_visible(self) -> bool {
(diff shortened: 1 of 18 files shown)


src/buffer/message_view.rs
```diff
@@ -128,7 +128,15 @@ impl<'a> ChannelQueryLayout<'a> {
         }
 
         // A message can only be redacted once.
-        message.redaction.is_none()
+        if message.redaction.is_none() {
+            return false;
```
luca020400, line 132: true

andymandias: :facepalm: Thanks for catching that!

luca020400:
This looks good, thanks!

data/src/config/buffer/redaction.rs
```diff
@@ -11,11 +11,16 @@ pub struct Redaction {
 pub enum Display {
     #[default]
     None,
+    Redacted,
```
casperstorm, line 14: I think `redacted` should be default. 🤔

casperstorm (approved):

PR #1900, opened 2026-04-28
Minor Display Read Marker Fixes

andymandias:
A couple minor fixes:
- Fix "mark as read" button and "mark as read" shortcuts on logs and highlights panes (by updating the display read marker when the history is open)
- Use display read marker to determine whether an open pane has unread messages


casperstorm (approved):

casperstorm:
CHANGELOG entry to fixed?

PR #1904, opened 2026-04-28
ircv3 +reply rendering and sending

### LINKED ISSUE #1646: Feature Request: sending/receiving IRCv3 replies
Halloy currently supports the `+reply` client tag for reactions, but it does not allow for sending messages as replies or displays when a message is a reply.

furudean:
fixes #1646

## features
 if +reply is negotiated...
- render +reply in message batches
- messages can be replied to by right-clicking and choosing "reply" from the context menu. this will attach the +reply to that message when the next message is sent
- uses `highlight` notification config to provide notifications, takes precedence over `highlight` with a custom string

## architecture
- when a message is processed, we pass any associated +reply msgid to history manager
- the history manager looks up the referenced message in local history and if found, stores the looked up content alongside the existing message in history
- the message renderer has the context needed to render the reply with context
- draft replies are implemented into the data model and persist in the same context as draft messages

## non-goals with this pr
in order to keep this PR relatively composable the following was not implemented, and may be implemented in follow-ups:
- key bindings to clear a pending reply, and to navigate up/down the history and reply to a specific message without mouse
- jumping to the message that was replied to with ui interactions
- showing a full preview of the replied message (perhaps as a tooltip?)
- ~~any configuration options to customize the reply line~~

Screenshot 2026-04-28 at 11 59 51 AM


casperstorm:
Wow, super nice addition. You are killing it.
I'll review it later; so this is just a first glance thought for the UI:
1. The icon takes quite a bit of attention, i am not sure its even needed now that we have the lines as well.
2. If we keep the icon we should make it optional.
3. I see you don't have modes on replies, should we add that as a option?
4. Would be nice to click a reply and move to it for context.

furudean:
Thank you!  I'll respond to the above in bullet point form:
1. I copied Discord's homework here in having an icon mark replies. My very limited usability test (showing my girlfriend the UI and asking if it looks clear what's being conveyed) preferred the arm with the icon to separate it from a regular message. I think we could make the icon less prominent by making it smaller or making it have the timestamp styling, but I don't think it's as simple as just removing the icon.
2. Agree. I deliberately avoided any configuration options for this PR to keep it focused. There will be follow-ups that address this.
3. We could do a nick lookup on the channel at render time, it would be a pretty easy addition.
4. It would be nice, but I think the actual feature to enable that is out of scope. There is no feature to jump to a specific message in history **by msgid** yet, so I would prefer to separate that work and keep this PR focused on the core stuff.

I'm flexible on any of these points.

furudean:
here's a quick style tweak with a smaller icon 

Screenshot 2026-04-28 at 2 25 07 PM

casperstorm:
> I copied Discord's homework here in having an icon mark replies.

Hmm i don't see icons in Discord. Again not against the icon but i don't think it should be on by default. The line should be enough. Discord replies in a busy channel:

image

furudean:
Mine is a bit different. I'm using the IRC-like layout, so that's probably why?
Screenshot 2026-04-29 at 3 27 57 PM

It looks like user icons get replaced by the reply symbol when you use the smaller IRC-like layout.

furudean:
Here is a comparison with and without the icon:

Screenshot 2026-04-29 at 3 37 20 PM
Screenshot 2026-04-29 at 3 37 23 PM

I find the version with the icon a lot easier to read compared to the second one. The arm by itself is easy to miss, so having an additional element gives it strong affordance. I could concede to put this under a setting if that's the only way forward, but my view is it would make the best default.

casperstorm:
This just proves that we 100% need a knob for the icon. That's also the Halloy way, anyway 😄

casperstorm:
So before we can i think the only thing we need is to make the icon optional, and off by default. Would also make sense to make it possible to adjust the size of it.

furudean:
I have added the requested settings and dug the icon's grave 🪦

casperstorm:
For multiline replies, should we preserve the reply tags when adding the batch tag, instead of overwriting all tags? I am not entirely sure how it should work, but it looks like `send_input_line_batch` adds `+reply`/`+draft/reply` first, and then `send_multiline_batch` replaces the whole tag map with just the batch tag. Wouldn't that mean local messages would show as reply but not the message sent to the server?

4e554c4c:
> For multiline replies, should we preserve the reply tags when adding the batch tag, instead of overwriting all tags? I am not entirely sure how it should work, but it looks like `send_input_line_batch` adds `+reply`/`+draft/reply` first, and then `send_multiline_batch` replaces the whole tag map with just the batch tag. Wouldn't that mean local messages would show as reply but not the message sent to the server?

They should never be added to the individual messages to begin with. The `+reply` should be added to the `BATCH` command. The multiline spec forbids all tags except `batch` on lines inside a multiline batch, and treats all tags on the first `BATCH` as tags for the batch "as a whole"

So this function should probably have a `reply` argument to add to the `opening_batch` message

furudean:
@4e554c4c, @casperstorm thanks for your input! i've made the changes now.

furudean:
It looks like the behavior is slightly different when viewed with multiline (fig 1) vs no multiline support (fig 2, proxied through soju). in ergo, the first msgid is the batch id. this works with a fresh chathistory batch as well.

not sure that there is much that can be done here... but we should be following the spec at least :-)

Screenshot 2026-05-01 at 10 10 24 AM

Screenshot 2026-05-01 at 10 10 11 AM

4e554c4c:
yes i think per spec this is the best that can be done

data/src/history/manager.rs
```diff
@@ -639,6 +639,14 @@ impl Manager {
         self.data.can_mark_as_read(kind)
     }
 
+    pub fn is_our_message(&self, id: &str) -> bool {
+        self.data.map.values().any(|history| {
+            history.find_by_id(id).is_some_and(|msg| {
+                msg.direction == crate::message::Direction::Sent || msg.is_echo
+            })
+        })
+    }
```
andymandias: Can we use `history::find_message_target`?  I believe the same expectation will still hold: the reply should have a `server_time` after the message it replies to, so we can start the search at `server_time` and work backwards (resorting to a full search only if necessary).  It's minor, but we have it available if we can use it.

data/src/history.rs
andymandias: If we want notifications for messages that reply to the user's message in a closed buffer, then we'll want to check for it in `append` and return the found replies in the result `Vec`.  A generic `enum` that will cover `ReactionToEcho` and `ReplyToEcho` will be needed.

data/src/user.rs
```diff
@@ -226,6 +227,91 @@ impl User {
...
+                .take(len.saturating_sub(1) as usize)
+                .collect();
+            truncated.push(truncation_character);
+            return std::borrow::Cow::Owned(truncated);
+        }
+    }
+    std::borrow::Cow::Borrowed(nick)
+}
+
+impl User {
```
andymandias: Do we need to restore these for some purpose in particular?  They were removed when creating `user_display` to prevent their use, since everywhere they were being used was handled more properly/consistently by `UserDisplay`.

src/buffer/message_view.rs
```diff
@@ -1146,11 +1170,164 @@ impl<'a> LayoutMessage<'a> for ChannelQueryLayout<'a> {
...
+                let highlight = self.theme.styles().buffer.highlight;
+                container(base_nick_element)
+                    .style(move |_| container::Style {
+                        background: Some(Background::Color(highlight)),
+                        ..Default::default()
+                    })
+                    .into()
+            } else {
+                base_nick_element
+            };
```
andymandias: Since this isn't too far afield from what we output from `UserDisplay`, I think it would be nice if we can make this behavior by an argument to `UserDisplay::new`.  I.e. `with_tooltip: bool` and the `tooltip` field of `UserDisplay` is left `None`, and `into_element` with a `highlight: bool` argument that adds the visual highlight.  Mainly I'm hoping we can funnel as many user display elements through the one interface so we don't have to deal with things getting out of sync.

data/src/history/manager.rs
furudean: we can't use `history::find_message_target` directly since it returns `&mut Message`, and `is_our_message` is `&self`, but we could employ a similar binary search approach to get the same

data/src/user.rs
furudean: i thought i needed this in order to get some flexibility back, though after a lot of iteration i arrived at using a shared `UserDisplay` again, i think this might have been a stray. i will push some changes that addresses this and should align it more with main

src/buffer/message_view.rs
furudean: this is a great suggestion. same as above - i stuffed more things into `UserDisplay` so we deduplicate this logic better. 

as as sidenote, this all being selectable text isn't ideal but can hopefully be fixed with future improvements.

data/src/history.rs
furudean: done - we have `EchoEvent` now, and unloaded replies are threaded through with the same mechanism as reactions

src/buffer/message_view.rs
andymandias: I agree, we want to be able to request a selectable or non-selectable variant from `into_element`.  In addition to where it can be used here, non-selectable will also be needed if/when we utilize `UserDisplay` for query pane titles.  If you want to add that functionality in this PR feel free, but it could certainly be left for a follow up.

data/src/history/manager.rs
andymandias: Ahh, I see!  Sounds like we should cut off the last line of `history::find_message_target` to make `history::position_message_target`, then that can be the base of `history::find_message_target` and `history_find_message_target_mut`.

furudean: i didn't update here but i did implement something similar in [`6f9d61d`](https://github.com/squidowl/halloy/pull/1904/commits/6f9d61d3109bcc00a0660db18d0d2761518ace75), curious what you think

andymandias: Looks good to me.  I do think it would be nice to combine the functionality so everything uses one base function for searching and there's only one place to modify if/when the message store changes and search mechanics need to be updated.  But I'm not particular about which way things get combined.

src/buffer/message_view.rs
furudean: done in [`67147b3`](https://github.com/squidowl/halloy/pull/1904/commits/67147b3045a33c662a9ff4b749b32179a6fac96f)

data/src/history/manager.rs
furudean: That's a bit of a bigger refactor probably, but I would be positive to that kind of change. Not 100% sure how to do it myself^^

andymandias: I can help with this part tomorrow.  I'll take another look through the whole PR then, but I think that'll be last bit as far as I'm concerned.

furudean: Thank you! 🙏

andymandias (approved):
Great addition, nicely done!

casperstorm (approved):
Good work here.

4e554c4c:
Thank you!

PR #1905, opened 2026-04-28
Channel matching on single line only

englut:
Thanks for spotting this @andymandias!

Multiline channel matching should not be possible now :saluting_face:


andymandias (approved):
Looks good to me.  Thank you!

PR #1906, opened 2026-04-28
localhost http is treated as a secure transport when using `servers.<name>.filehost.override_url`

### LINKED ISSUE #1875: filehost and override_url
After updating halloy to the last version, I wanted to test the upload feature.
Sadly, none of my currently connected irc networks seems to implement it, so i tried to use `override_url` and [convoyeur](https://codeberg.org/classabbyamp/convoyeur) to bypass it.

`/upload` complains: `upload failed: refusing plaintext HTTP upload over an encrypted IRC connection` although the `override_url` is on localhost and convoyeur uses an external file upload service that DO use encryption, so the link that is send is https (on the single irc network i'm on that doesn't support encryption, so it allowed the http override_url).

Could there be an option to allow plaintext (with credentials=none in my case since everything is on localhost) when using override_url AND/OR credentials=none (since apparently it's the origin of the issue).

furudean:
fixes #1875

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ Fixed:
 Changed:
 
 - On-join topic messages disabled by default (`buffer.server_messages.topic`), instead the topic banner is enabled by default (`buffer.channel.topic_banner`)
+- For filehosts, localhost is now treated as a secure transport when using `servers.<name>.filehost.override_url`
 
 Thanks:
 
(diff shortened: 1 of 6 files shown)


andymandias (approved):
I haven't set up a localhost filehost to test, but it looks good to me.

Would it be worth putting a note in the `override_url` documentation to let users know about the exception?

furudean:
> Would it be worth putting a note in the `override_url` documentation to let users know about the exception?

Maybe? This kind of thing is invisible to end users

andymandias:
For users who try localhost directly, just works.  For users who try another http address, then it'll seem like no http address will work.

PR #1908, opened 2026-04-29
Close side effects for buffers

casperstorm:
Add knob to add side effect to close query when closing query buffer.


andymandias (approved):
Looks good, works well.

PR #1910, opened 2026-04-29
Always Be Updating Bot Status

andymandias:
Update bot status on PRIVMSG/NOTICE whether bot or not; necessary to track bot mode changes.

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -1523,8 +1523,9 @@ impl Client {
                         }
                     }
 
-                    if let Some(channel) =
-                        self.message_channel_target(&message.command)
+                    if self.isupport.contains_key(&isupport::Kind::BOT)
+                        && let Some(channel) =
+                            self.message_channel_target(&message.command)
                         && let Some(ch) = self.chanmap.get_mut(&channel)
                     {
                         ch.users.update_user(
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -1523,14 +1523,15 @@ impl Client {
                         }
                     }
 
-                    if message.tags.contains_key("bot")
-                        && let Some(channel) =
-                            self.message_channel_target(&message.command)
+                    if let Some(channel) =
+                        self.message_channel_target(&message.command)
                         && let Some(ch) = self.chanmap.get_mut(&channel)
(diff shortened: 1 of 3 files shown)


data/src/client.rs
```diff
@@ -1523,14 +1523,15 @@ impl Client {
...
+                        self.message_channel_target(&message.command)
                         && let Some(ch) = self.chanmap.get_mut(&channel)
-                        && let Some(mut chan_user) = ch.users.take(&user)
                     {
-                        chan_user.update_bot(true);
-                        ch.users.insert(chan_user);
+                        ch.users.update_user(
+                            &user,
+                            None,
+                            Some(message.tags.contains_key("bot")),
```
englut, line 1534: Maybe we can check if the value is changed before setting it :pray: 

```rust
let is_bot = message.tags.contains_key("bot");
let user_bot_needs_update = if user.is_bot() == is_bot {
    Some(is_bot)
} else {
    None
};
ch.users.update_user(
    &user,
    None,
    user_bot_needs_update,
);
```

englut:
This is great! I was wondering where this code was! Does your change here make the `update_bot` in `data/src/messages.rs` redundant?
https://github.com/squidowl/halloy/blob/8a0d69529cd8a323632dcf8550d42ad43962f0f6/data/src/message.rs#L2213

data/src/client.rs
andymandias: `user` here is parsed from the message, not the list of channel users.  We'd want to check if the user in the list of channel users already has bot status, and since that requires retrieving the user from the list we might as well just update.

englut: Ahhh that makes sense, not really optimizing at all then in that case. Fair enough! :pray:

andymandias:
> This is great! I was wondering where this code was! Does your change here make the `update_bot` in `data/src/messages.rs` redundant?
> 
> https://github.com/squidowl/halloy/blob/8a0d69529cd8a323632dcf8550d42ad43962f0f6/data/src/message.rs#L2213

We'll need it for queries still, but we should be able to remove it from `resolve_attributes` post-processing.  I'll test that.

casperstorm (approved):

PR #1915, opened 2026-05-01
Fix link matching for both end of line delimiters & punctuations

englut:
Fixes bug with link matching logic where doubling up on end of line delimiter + punctuation would only exclude the punctuation from the match.

e.g. `"a #test",` matched to `#test"`.

With fix in place, this matches to `#test`.


casperstorm (approved):

casperstorm:
LGTM, thank you!

PR #1919, opened 2026-05-02
Handle URLs ending with paired delimiter chars

englut:
Added more handling of valid test cases for URLs containing and ending with paired delimiters (includes ending with open chars).


andymandias (approved):
Looks good to me!

PR #1923, opened 2026-05-03
User Display Tweaks

andymandias:
Minor fixes/tweaks:
- Bot icon vertical alignment
- Show tooltip in the nicklist if the user is a bot even if the user display was not truncated
- Use nicklist_nickname setting when constructing nickname elements for the nicklist


casperstorm (approved):
LGTM

PR #1925, opened 2026-05-03
Parse Homepage Metadata

andymandias:
Parse homepage metadata and make it clickable in context menu.


casperstorm (approved):
All the metadata attributes!!

PR #1928, opened 2026-05-04
`VERSION` and `JOIN` commands on connect are skipped if server is primary soju server

englut:
Right now we send `VERSION` to the soju primary server, but soju doesn't support `VERSION` so this leads to seeing the following in the soju server buffer:

```
VERSION Unknown command
```

I'm also seeing `JOIN`s being attempted on the primary soju server, so I expanded the conditional for connect joins to exclude the primary server as well.

```
JOIN Cannot interact with channels and users on the bouncer connection. Did you mean to use a specific network?
```

Both scenarios are cleaned up via this PR.


andymandias (approved):

andymandias:
Thank you!

PR #1930, opened 2026-05-04
Misc. Minor Improvements/Fixes

andymandias:
A few miscellaneous imrpovements/fixes:
- Use SETNAME for setting resolved user/host (prepares the way for `draft/whoami` support)
- Hide metadata subscription confirmation from the server buffer
- Fix not clearing metadata when removing all metadata settings for a server
- Fix edge case where bot status could fail to be updated when it should be updated
- Use same color seed for query pane titles as is used for buffer/nicklist
- Reduce the number of excess chathistory messages requested from queries returned by a chathistory TARGETS request

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -2768,6 +2768,24 @@ impl Client {
                     self.resolved_host = Some(hostname.to_string());
                 }
             }
+            Command::SETNAME(_realname) => {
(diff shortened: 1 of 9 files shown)


casperstorm (approved):
Thanks for these QOL.

PR #1934, opened 2026-05-06
Handle leading delimiters when parsing link fragments to match nicks after quotes

englut:
~~Seems that there was an erroneous `"` in the user regex. I don't believe `"` is a valid char for IRC nicks.~~

~~Removing it~~ Fixed to allow~~ed~~ for nicks to be matched at the beginning of quoted text:

`This test "quaff, says hi".` matches `quaff` as the nick.

### DIFF
--- a/data/src/message.rs
+++ b/data/src/message.rs
@@ -1998,15 +1998,19 @@ fn parse_regex_fragments<'a>(
             &text,
         );
         let (matching, leading_delimiter) = filter_leading_delimiter(matching);
+
         if let Some(fragment) = (f)(matching) {
+            let mut leading_text = String::new();
             if i < re_match.start() {
-                fragments.push(Fragment::Text(
-                    text[i..re_match.start()].to_string(),
-                ));
+                leading_text.push_str(&text[i..re_match.start()]);
             }
             if let Some(delimiter) = leading_delimiter {
-                fragments.push(Fragment::Text(delimiter.to_string()));
+                leading_text.push_str(delimiter);
             }
+            if !leading_text.is_empty() {
+                merge_text_fragment(&mut fragments, leading_text);
+            }
+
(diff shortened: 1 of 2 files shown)


furudean:
to be pedantic, irc specs do not restrict `"` even though some server software may do so: https://modern.ircdocs.horse/#clients

englut:
> to be pedantic, irc specs do not restrict `"` even though some server software may do so: https://modern.ircdocs.horse/#clients

Yeah I saw that, but I am not seeing it being valid on libera, oftc, or ergo. Possible that there's a server out there that's allowing `"` in a nick. But this falls into the territory of which edge case do we care more about? I feel there's probably more scenarios where people put quotes around a `nick` than having a nick in there, but that's not scientific at all :joy:

englut:
Restored allowing " as a valid nick character (can't be first character though).

Added leading delimiter handling that removes " if it's at the beginning of a match with no trailing delimiter matched.

:pray:

data/src/message.rs
```diff
@@ -4328,33 +4343,58 @@ pub mod tests {
...
+                            User::from(Nick::from_str(nick, casemapping))
+                        })
+                        .collect::<ChannelUsers>(),
                 ),
-                Fragment::Text("?".into()),
-            ],
-        )];
+                vec![
+                    Fragment::Text("And they said: ".into()),
+                    Fragment::Text("\"".into()),
```
andymandias: Is it possible to get the new matching behavior and produce a single `Fragment::Text` here?  I believe it's generally preferable to produce as few fragments as possible.

englut: Gooood call. I added logic to look at the previous fragment (if there is one), and if it's Text, we merge into them. if not, we push a new one!

andymandias (approved):
LGTM!

PR #1937, opened 2026-05-06
chathistory Loop Fix

andymandias:
Fixes a scenario where chathistory requests could get in a loop, using unnecessary bandwidth and memory.

PR #1939, opened 2026-05-07
Minor History Changes

andymandias:
Two minor changes to history handling:
- Make the loop prevention merged in #1937 more defensive (updates should prevent looping even if the server/bouncer is malfunctioning)
- Make the update method for deduplicated messages and echo messages consistent; both types of message updates can be handled the same, and the criteria used to distinguish the two could fail


casperstorm (approved):

casperstorm (approved):
LGTM

PR #1942, opened 2026-05-07
support `draft/chathistory-end` message tag

furudean:
implements https://github.com/ircv3/ircv3-specifications/pull/598

this fixes an edge case when chathistory ends on a page boundary, where it is unclear whether or not there are any more messages to query. the server can indiciate draft/chathistory-end on a batch, and then we set `exhausted = true` eagerly and preventing a spurious request

has upstream support in ergo to be released in next version. no support in soju yet

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -5955,6 +5955,100 @@ mod tests {
     }
 
     #[test]
+    fn chathistory_end_tag_marks_history_exhausted() {
+        let mut client = test_client("tester");
+        let config = config::Config::default();
+        let server = proto::Source::Server("irc.test".to_string());
+        let alice = proto::Source::User(proto::User {
+            nickname: "alice".to_string(),
+            username: Some("alice".to_string()),
+            hostname: Some("example.test".to_string()),
+        });
+
+        let target = Target::parse(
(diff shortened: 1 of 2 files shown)


data/src/client.rs
```diff
@@ -3804,7 +3810,11 @@ impl Client {
                 ) = subcommand
             {
                 self.chathistory_exhausted
-                    .insert(target.clone(), size < *limit);
+                    .insert(target.clone(), chathistory_end || size < *limit);
+            }
+
+            if chathistory_end {
+                return None;
             }
```
andymandias: This makes perfect sense in terms of the return value, but we cut off all the log messages if we return here.  I think we need to either move the log messages above it, or use the bool inside the following match when determining whether to call a `continue_*` function.

andymandias (changes requested):
Looks good!  The test is really nice.  Just one nit, but otherwise good to go.

data/src/client.rs
furudean: yeahhh that makes a lot more sense. fixed now!

andymandias (approved):
Thank you!

PR #1945, opened 2026-05-08
adds `buffer.server_messages.condense.max`

furudean:
> The config `buffer.server_messages.condense.max` may be used to limit the number of user entries shown in condensed messages

useful in large channels, especially during restarts and netsplits

Screenshot 2026-05-08 at 10 18 25 AM


casperstorm (approved):
LGTM!

andymandias (approved):

PR #1947, opened 2026-05-08
Mark-as-Read On-Message Tweaks

andymandias:
Changes re: `buffer.mark_as_read.on_message`:
- Display marked as read for messages that don't trigger unread (i.e. prevent non-`PRIVMSG`/`NOTICE` messages from appearing after the backlog divider, or showing the backlog divider if it is set to hide).
- Expands the setting as an enum with `"focused"`, `"open"`, and `"none"` values;  corresponding to only for the focused pane, for all open panes, and for no panes
- Fixes bug/regression where setting would apply even when Halloy was not focused

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1939,6 +1939,8 @@ fn handle_single_event(
         return;
     };
 
+    handle_on_message_display_mark_as_read(server, &message, dashboard, config);
+
     commands.push(
         dashboard
             .block_and_record_message(
@@ -1974,6 +1976,8 @@ fn handle_with_target_event(
     ) else {
         return;
     };
+
+    handle_on_message_display_mark_as_read(server, &message, dashboard, config);
 
     commands.push(
         dashboard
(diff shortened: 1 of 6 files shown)


src/main.rs
```diff
@@ -2036,13 +2036,14 @@ fn handle_priv_or_notice(
         .as_ref()
         .and_then(|kind| dashboard.find_window_with_history(kind));
 
-    let should_mark_as_read = config.buffer.mark_as_read.on_message
-        && !msg.blocked
```
englut, line 2040: If we remove this, does that mean messages from blocked people will start showing the "mark as read" button?

andymandias: No, it means we'll send a MARKREAD to the server even for blocked messages.

englut: Ah right, which is probably desirable for blocked messages. gotcha!

andymandias: I can go either way on whether that's appropriate, my intuition is that `on_message = "open"` or `on_message = "focused"` users will prefer it that way but if I hear otherwise I'm happy to switch it back.

casperstorm (approved):
LGTM

PR #1948, opened 2026-05-08
tooltip adjustments

furudean:
- render bot mode tooltip with a color consistent to other tooltips (with `theme::text::secondary` instead of `theme::text::primary`)
- simplify bot mode vertical alignment independent of font
- all tooltips now have a text align of 1.0. this is more consistent with how tooltips are handled across different UIs

<details>
<summary>
before/afters
</summary>

Screenshot 2026-05-08 at 4 48 26 PM
Screenshot 2026-05-08 at 4 48 29 PM
Screenshot 2026-05-08 at 4 48 42 PM
Screenshot 2026-05-08 at 4 48 33 PM

</details>

### DIFF
--- a/src/widget/user_display.rs
+++ b/src/widget/user_display.rs
@@ -5,10 +5,11 @@ use data::target::{Query, TargetRef};
 use data::user::AccessLevel;
 use data::{Config, User, metadata};
 use iced::Color;
+use iced::alignment::Vertical;
 use iced::widget::{container, row};
 use unicode_segmentation::UnicodeSegmentation;
 
-use super::{Element, selectable_text};
+use super::{Element, selectable_text, text};
 use crate::widget::TextExt as _;
 use crate::{Theme, font, theme, widget};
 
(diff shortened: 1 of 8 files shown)


casperstorm (approved):

src/widget/user_display.rs
```diff
@@ -329,6 +360,7 @@ impl UserDisplayData {
                 self.right.map(|right| text_piece(right, font)),
             ]
             .spacing(theme::ICON_SPACE)
+            .align_y(Vertical::Center)
             .into()
```
andymandias, line 364: Using this instead of `line_height(iced::widget::text::LineHeight::Relative(1.45))` causes the nickname to copy-paste as
```
<limnoria
🤖
>
```
instead of as it appears in the buffer, `<limnoria🤖>`.  We need a workaround or to fix that behavior upstream if we want to use `align_y` for a row of selectable text.

andymandias (changes requested):
Overall this looks good, but we have a minor regression we should handle before we commit to it.

src/widget/user_display.rs
furudean: aw that's a bummer. i forgot that that was an issue. i can go back on this, but that does mean that the bot line will have a line height inconsistent with others just to offset it and make it look centered. variable line heights have an effect on the overall dimensions of its containing element

part of me wonders if its worth keeping the 🤖 icon selectable or just dropping it for simplicity

andymandias (approved):

furudean:
Sorry to nag, any blockers to merging this?

PR #1950, opened 2026-05-09
Add client side ping for websocket keepalive

luca020400:
This differs from the IRC ping as we don't send
anything on the wire, just a Message::Ping.

This is "needed" on servers behind a cloudflare proxy.


casperstorm (commented):
Looks good, @luca020400 - could you perhaps update documentation as well?

casperstorm (approved):

PR #1953, opened 2026-05-09
reply: show a preview on hover

furudean:
after hovering a reply for `buffer.reply.tooltip.delay` milliseconds, a tooltip will pop up near the mouse that shows the replied-to message in full with the same formatting as regular messages

Screenshot 2026-05-09 at 10 47 33 AM

also works with cards and image lists 
Screenshot 2026-05-09 at 10 50 53 AM

since rich content is rendered, the preview is no longer stored as plain text and instead as a `message::Content`.

a minor thing is that this tooltip uses the buffer's background instead of the regular tooltip background color. this is because the content in a theme is designed to work against that background, so this makes sure the actual content is readable across themes

oh, also refactors some of the reply line ui building logic to be more clear


casperstorm:
Thanks for this, @furudean. 
A small nit i found is if i post a url, and hide the preview - i can see the preview when hovering in a reply. I think it should respect if its not shown, similar to if a user has excluded it.

Screenshot 2026-05-10 at 20 25 30

furudean:
@casperstorm fixed in [`82df804`](https://github.com/squidowl/halloy/pull/1953/commits/82df804bb20ac903dd66e964b5277dae049971fe)

docs/configuration/buffer.md
```diff
@@ -1669,3 +1669,44 @@ This is useful for clients that don't support replies, as they may not otherwise
...
+enabled = false
+```
+
+#### `delay`
+
+Delay in milliseconds before tooltip appears.
+
+```toml
+# Type: integer
+# Values: positive integer
```
andymandias: Nit: non-negative integer

src/buffer/message_view.rs
```diff
@@ -1583,104 +1536,383 @@ impl<'a> ChannelQueryLayout<'a> {
...
+        row = row.push(preview);
+
+        let delay = iced::time::Duration::from_millis(
+            self.config.buffer.reply.tooltip.delay,
+        );
+
+        let element = element_tooltip(
+            row.align_y(alignment::Vertical::Center),
+            tooltip,
+            tooltip::Position::FollowCursor,
```
andymandias: With the current text layout/shaping we have available the tooltip ends up being a bit wobbly as the cursor moves.  I wonder if it might be better for the time being to anchor the tooltip with respect to the reply row?

data/src/config/buffer.rs
```diff
@@ -92,6 +93,25 @@ impl Default for Reply {
...
+    pub delay: u64,
+    pub max_width: f32,
+}
+
+impl Default for ReplyTooltip {
+    fn default() -> Self {
+        Self {
+            enabled: true,
+            delay: 500,
+            max_width: 500.0,
```
andymandias: I think these are the right settings and defaults.  Well chosen.

src/buffer/scroll_view.rs
```diff
@@ -1115,26 +1130,38 @@ impl State {
...
+                        self.pending_preview_exits = still_pending;
                     }
 
-                    self.pending_preview_exits = still_pending;
+                    let before = self.reply_preview_urls.len();
+                    self.reply_preview_urls
+                        .retain(|hash, _| rendered_hashes.contains(hash));
+                    if self.reply_preview_urls.len() < before {
+                        preview_changed = true;
+                    }
```
andymandias: Since we're removing urls immediately on `Message::ReplyPreviewHovered`/`Message::ReplyPreviewUnhovered`, rather than putting them into a pending exit set, can we skip checking `reply_preview_urls` for changes here?

src/buffer/message_view.rs
```diff
@@ -1583,104 +1536,383 @@ impl<'a> ChannelQueryLayout<'a> {
...
+                let el = preview_content(preview, self.config, self.theme);
+                let el: Element<_> =
+                    if matches!(preview, data::Preview::Card(..)) {
+                        container(el)
+                            .style(theme::container::hover_preview_tooltip)
+                            .into()
+                    } else {
+                        container(el).max_height(200).into()
+                    };
+                content_col = content_col.push(el);
```
andymandias: Nit: a `match` here with an arm for each `data::Preview` variant would ensure it gets flagged by the compiler if we ever add a new variant.

andymandias (commented):
Works nicely, well done!

My understanding is that the sharp corners on the tooltip are to indicate it's a view into another part of the buffer than a regular tooltip.  I think a visual cue for that distinction makes sense.  In line with that effect, assuming I'm understanding it correctly, perhaps it make sense to style in-tooltip card previews as if they were buttons, so that they appear similar to how they will in the buffer?

src/buffer/message_view.rs
furudean: I experimented a fair bit trying to make the static tooltip position work. Since the width of the content can be relatively wide, it can make the tooltip appear really far away from the mouse position (try a wide buffer with a long message). I wasn't able to find a good compromise sadly, but maybe it's possible with more iteration. The tooltip being centered around the content itself I think is hardcoded.

What do you think?

furudean:
@andymandias that's a good callout on the card previews - it had slipped past me that they aren't styled the same way. totally agreed, and it should be an easy fix.

src/buffer/scroll_view.rs
furudean: fixed in [`c642be6`](https://github.com/squidowl/halloy/pull/1953/commits/c642be629cc2e16c03a106b63af436a7320c0a6b)

src/buffer/message_view.rs
furudean: fixed

docs/configuration/buffer.md
furudean: fixed

furudean:
the card style has been implemented

casperstorm (approved):
Good work on this one 👍🏻

src/buffer/scroll_view.rs
furudean: we caught up in irc. properly fixed in [`29ac8df`](https://github.com/squidowl/halloy/pull/1953/commits/29ac8df0569860651d21f31f5c2a003fbd4367c7), i think i just misunderstood what you meant :-)

andymandias (approved):
Thank you!

PR #1955, opened 2026-05-09
Mark as read and close query user on sidebar

lcign:
All buffers and panes open are marked as read also on sidebar.
When closing a private query it also disappears in sidebar


andymandias (commented):
`sidebar.unread_indicator.show_on_open_buffers` already allows unread indicators on open buffers to be hidden.  This also throws a bunch of mark-as-read actions on various events that should be configurable if we're going to consider adding them.  I think this would benefit from a clear stating of intent, since it is currently forcing behavior that suspect is natural to you but is very much not intuitive to me (e.g. every time you click on a pane you want it to be marked as read?).

A setting to close queries when closing the associated pane could make sense, but it would definitely need to be a setting.

lcign:
Right, I might be a little bit confusing about my intents. 
My logic is: every channel pane/panes and query opened should be marked as read because you are seeing them. 
Closing a query should also be useful to make it disappead from sidebar.
I'm absolutely new to rust and just getting back programming - last time was late 90's :^)

PR #1956, opened 2026-05-09
reply: handler for `Event::GoToMessage` + visual indicator

furudean:
- you can click a reply to trigger `Event::GoToMessage`
- adds a visual highlight when `Event::GoToMessage` runs to clearly show which message was jumped to
- adds a secondary highlight when the hovered reply is already visible in the buffer, short-circuiting the tooltip preview
- fix incorrect vertical spacing caused by reply arm font size


casperstorm:
Do we respect a user’s preview include/exclude settings here, or could reply tooltips show media that regular inline previews would hide?

furudean:
@casperstorm It's not explicit, but it's using the same shared logic as for the buffer image/card previews, so there shouldn't be a difference in how its handled.

furudean:
@casperstorm I tested with a few include/exclude configs and it seems to behave the same way as regular buffer previews do. Previews get evicted from the cache on config reload so any current previews fall into an error handling path and get hidden. We could still be explicit and filter them out, but we basically get it for free already with our current cache logic.

docs/configuration/buffer.md
```diff
@@ -1670,6 +1670,19 @@ This is useful for clients that don't support replies, as they may not otherwise
 insert_nick = false
 ```
 
+### `highlight_hovered_message`
+
+When hovering a reply, highlight the target message if it visible in the buffer.
```
furudean:
```suggestion
When hovering a reply, highlight the target message if it visible in the buffer. This takes precedence over the hover preview.
```

docs/configuration/buffer.md
```diff
@@ -1670,6 +1670,19 @@ This is useful for clients that don't support replies, as they may not otherwise
 insert_nick = false
 ```
 
+### `highlight_hovered_message`
+
+When hovering a reply, highlight the target message if it visible in the buffer. This takes precedence over the hover preview.
```
furudean:
```suggestion
When hovering a reply, highlight the target message if it visible in the buffer. This takes precedence over the tooltip.
```

casperstorm (approved):

PR #1957, opened 2026-05-10
SVG Support

andymandias:
Adds support for SVGs to previews, avatars, and server icons.


luca020400:
This works perfectly, sad we need to invalidate the cache manually.

I think I'll start improving cache on that side :)

andymandias:
> This works perfectly, sad we need to invalidate the cache manually.
> 
> 
> 
> I think I'll start improving cache on that side :)

Sounds good! It could be nice to be able to clear the cache via the command bar, if that's something you'd be interested in hooking up :)

data/src/image.rs
```diff
@@ -0,0 +1,88 @@
...
+        image::guess_format(bytes).ok().map(Format::Raster)
+    }
+
+    pub fn from_mime_type(mime_type: &str) -> Option<Format> {
+        if mime_type == "image/svg+xml" {
+            Some(Format::Svg)
+        } else {
+            image::ImageFormat::from_mime_type(mime_type).map(Format::Raster)
+        }
+    }
```
casperstorm, line 28: Nit: if the content type is `content-type: image/svg+xml; charset=utf-8` the svg wont display, eg: `https://api.iconify.design/simple-icons:rust.svg`.

andymandias: Thank you for catching that!

casperstorm (approved):
Works great!

esden:
Tested it out over here too, and I can confirm that image previews as well as server icons and user avatars in SVG format all work! 🎉

PR #1959, opened 2026-05-10
Implement age based cache

luca020400:
Attempts to implement caching as specified in https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Caching


furudean:
this is still a draft, but i wanted to pitch in to suggest we also send the [`If-None-Match` header and handle 304 not modified](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/304). Many modern file servers serve this file to tell the server that the asset hasn't changed, so it can keep its cached item

luca020400:
As soon as I started to add more functionality the complexity exploded and I don't feel confident at all in its expected behaviour.

So I went ahead and decided that using a pre-existing crate was the right choice, https://github.com/squidowl/halloy/pull/1962

It's a wrapper around reqwuest that does caching and follows the RFC spec for cache age and friends (supposedly, didn't check code) via yet another crate.

PR #1962, opened 2026-05-11
Migrate to http-cache-reqwest

luca020400:


luca020400:
turns out writing cache logic is hard.

also this completely ignores user settings about size/limits, probably we may want to rename those into a generic "cache" config

luca020400:
I think what might be up for debate is whether the cached/non-cached http client makes sense.

Also this paves the way to a generic "clear cache" since it's exposed by the manager.

data/src/config/proxy.rs
```diff
@@ -120,6 +126,24 @@ pub fn build_client(
...
+    proxy: Option<&Proxy>,
+    identity: Option<reqwest::Identity>,
+) -> Result<ClientWithMiddleware, BuildError> {
+    let client = build_client(proxy, identity)?;
+
+    let cache_root = environment::cache_dir().join("http_cache");
+    let client = ClientBuilder::new(client)
+        .with(Cache(HttpCache {
+            mode: CacheMode::Default,
+            manager: CACacheManager::new(cache_root, false),
```
luca020400, line 139: If we later want to implement 'clear cache' the manager will likely need stored somewhere, and then passed down here

data/src/client.rs
```diff
@@ -233,6 +234,7 @@ pub struct Client {
     metadata_syncs: BinaryHeap<MetadataSync>,
     channel_discovery_manager: channel_discovery::Manager,
     http_client: Option<Arc<reqwest::Client>>, // Only Some if config.proxy.is_some()
+    http_cached_client: Option<Arc<ClientWithMiddleware>>, // Only Some if config.proxy.is_some()
```
luca020400, line 237: Do we want that? Or for simplicity shall we always go with cache http client?

data/src/preview.rs
```diff
@@ -175,81 +157,40 @@ enum Kind {
...
-            Err(error) => {
-                cache
-                    .save::<Preview>(&cache_key_url, &CacheState::Error)
-                    .await;
-                Err(error)
-            }
+        Kind::Avatar => {
+            load_avatar_uncached(url.clone(), client, preview_config).await
         }
     };
```
luca020400, line 195: This probably can always be cached?

data/src/server_icon.rs
```diff
@@ -139,75 +97,13 @@ impl Manager {
...
-            }
-        }
-    }
-}
-
-async fn remove(url: Url, cache: Arc<FileCache>) {
-    let cache_key_url = canonical_icon_url(&url);
-
-    cache.remove::<Image>(&cache_key_url).await;
 }
```
luca020400, line 203: I had to remove this as there's no "clear specific cache entry" in the manager, but I think with proper cache handling it's not an issue?

src/screen/dashboard.rs
```diff
@@ -5367,20 +5351,28 @@ fn http_client_from_config(config: &Config) -> Option<reqwest::Client> {
...
 
-    cache::FileCache::new(
-        root,
-        config.request.image_cache.max_size_bytes(),
-        config.request.image_cache.trim_interval,
-    )
+            None
+        }
+    }
 }
```
luca020400, line 5386: This means at the moment no config is taken into account, I think I can configure some of it via the manager-

PR #1963, opened 2026-05-11
Move websocket heartbeat into a FSM

luca020400:
Before if the tick was ready we would only
send the Ping message if the stream was
also ready.

This may not always be the case, and we would
loose the specific heartbeat...


casperstorm (approved):
LGTM, would have done it in a similar way.

PR #1965, opened 2026-05-12
Settings for Hiding URLs in Messages with Visible Previews

andymandias:
Adds settings for hiding URLs in messages when the URL's preview is visible and a condition is met.  Implemented conditions are:
- the message consists of only the URL
- the URL is the trailing fragment in the message (ignoring whitespace and hidden fragments)

### DIFF
--- a/src/buffer/message_view.rs
+++ b/src/buffer/message_view.rs
@@ -1,4 +1,4 @@
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashMap, HashSet};
 
 use chrono::{DateTime, TimeDelta, Utc};
 use data::buffer::RightAlignmentWidths;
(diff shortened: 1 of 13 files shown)


furudean:
should be a setting, maybe?

casperstorm:
> should be a setting, maybe?

we need a merch t-shirt with this slogan.

src/buffer/message_view.rs
```diff
@@ -90,12 +97,18 @@ pub struct ChannelQueryLayout<'a> {
...
+
     fn preview_hidden_for_url(
         &self,
         message: &data::Message,
         url: &str,
     ) -> Option<bool> {
-        if !self.config.preview.is_enabled(url) {
+        if self.previews_enabled(message)
+            || !self.config.preview.is_enabled(url)
+        {
```
casperstorm: Don't we want:

```rust
        if !self.previews_enabled(message)
            || !self.config.preview.is_enabled(url)
        {
            return None;
        }
```

Otherwise we can show/hide on a url.

casperstorm (approved):
Just a small nit

esden:
I finally came around to testing this out. It really helps to clean up the chat, especially in channels that have to deal with discord bridges. (Discord media links are quite something...) Well done! 🙏

src/buffer/message_view.rs
andymandias: Sure do, thank you for catching that!

PR #1966, opened 2026-05-12
Dynamic default ports for irc, ircs, ws and wss

englut:
Added dynamic default ports, when no port is set in a server config, the default port will be chosen based on the value of `use_tls` and `use_websocket`.

Default port values:

- `irc+secure`: `6697` (default)
- `irc+insecure`: `6667`
- `ws+secure`: `443`
- `ws+insecure`: `80`

I also updated the URL regex to parse `ws://` & `wss://` links. I almost added these to the url schema handlers of the various OSes, but realized these aren't IRC specific, so I left them out for now.

### DIFF
--- a/data/src/server.rs
+++ b/data/src/server.rs
@@ -163,7 +163,7 @@ impl ConfigMap {
         for (i, (server, mut config)) in iter.into_iter().enumerate() {
             if config.port.is_none() {
                 config.port =
-                    default_port(config.use_tls, config.use_websocket);
+                    Some(default_port(config.use_tls, config.use_websocket));
             }
             if let Some(pass_file) = &config.password_file {
                 if config.password.is_some()
--- a/data/src/server.rs
+++ b/data/src/server.rs
@@ -10,7 +10,9 @@ use tokio::fs;
 
 use crate::bouncer::BouncerNetwork;
 use crate::config::buffer::typing::Typing;
-use crate::config::server::{filehost, read_from_command};
+use crate::config::server::{
+    DEFAULT_UNSET_PORT, default_port, filehost, read_from_command,
+};
 use crate::config::sidebar::{OrderBy, OrderChannelsBy};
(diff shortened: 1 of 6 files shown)


data/src/server.rs
```diff
@@ -161,6 +161,10 @@ impl ConfigMap {
     ) -> Result<Self, Error> {
         let mut map = IndexMap::new();
         for (i, (server, mut config)) in iter.into_iter().enumerate() {
+            if config.port == 0 {
+                config.port =
+                    default_port(config.use_tls, config.use_websocket);
+            }
```
englut, line 167: @casperstorm as far as I can tell, this is the best way to keep a value in `port` and also allow for detecting if a user left `port` unset in their config file.

I've added a `deserialize_port` function to prevent the user from setting `0` in their config, since 0 is an invalid port this should work out nicely for our purposes.

casperstorm: Ah now i understood what the issue is. You were TOTALLY right before with the `Option`. 😅 
Let's go back to that to avoid `DEFAULT_UNSET_PORT` and perhaps we should use `port: Option<NonZeroU16>`? 🤔  Sorry for the back and forward here.

englut: No worries! My previous `Option` implementation was missing this very important piece! Without your back and forth, I would have taken longer to get there, and `NonZeroU16` is dope :fire: 

Updated with `Option<NonZeroU16>`, so much better now! Thanks @casperstorm :pray: :heart_eyes:

casperstorm:
Alrite, i just pushed a small commit to remove a unwrap, but now its good! Thank you @englut.

casperstorm (approved):

englut:
> Alrite, i just pushed a small commit to remove a unwrap, but now its good! Thank you @englut.

🔥 I'll keep an eye on that usage next time! Thank you!!

PR #1968, opened 2026-05-13
colored reaction nicks + layout

furudean:
adds colors to reacted nicks and structures them into a wrapping list instead of each having a line, which increases information density. we also get prefixes and metadata for free through `UserDisplay`

Screenshot 2026-05-13 at 9 35 33 AM
Screenshot 2026-05-13 at 9 35 26 AM
Screenshot 2026-05-13 at 9 35 24 AM
Screenshot 2026-05-13 at 9 15 33 AM

### DIFF
--- a/src/widget/reaction_row.rs
+++ b/src/widget/reaction_row.rs
@@ -80,10 +80,13 @@ where
             })
             .into();
 
-        let hidden_count =
-            nicks.len().saturating_sub(REACTION_TOOLTIP_MAX_NAMES);
-        let displayed: Vec<_> =
-            nicks.iter().take(REACTION_TOOLTIP_MAX_NAMES).collect();
+        let display_max = if nicks.len() == REACTION_TOOLTIP_MAX_NAMES + 1 {
+            REACTION_TOOLTIP_MAX_NAMES + 1
+        } else {
+            REACTION_TOOLTIP_MAX_NAMES
+        };
+        let hidden_count = nicks.len().saturating_sub(display_max);
+        let displayed: Vec<_> = nicks.iter().take(display_max).collect();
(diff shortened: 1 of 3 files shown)


src/widget/reaction_row.rs
```diff
@@ -73,25 +80,79 @@ where
...
+                false,
+            )
+            .into_element(
+                user, false, false, None, None, false, false, theme, config,
+            );
+
+            // trailing separator attached to nicks so it never starts a new line alone
+            let position = i + 1;
+            let trailing = if position + 1 == total && hidden_count == 0 {
+                Some(" and")
```
andymandias, line 124: RIP oxford comma 😔

andymandias (approved):
Looks good to me.

src/widget/reaction_row.rs
furudean: so what's the halloy style guide anyway 😂

andymandias: Consistency is primarily what I aim for, so if there's prior art (that I'm aware of) inside Halloy then I try to follow its style.  Some guidelines would probably help with that, as things definitely fall through the cracks (e.g. tooltips are sometimes title case, sometimes not :face_with_diagonal_mouth: ).  Looks like we use the oxford comma in `join_targets` (`data/src/target.rs`), probably because I wrote that and it's my default style :laughing: .  Don't see it used (or not used) elsewhere, in a quick scan of `\".* and.*\"`.  But I don't actually have a strong preference between the two, so long as the meaning of the text is clear (which is the case here).

furudean: i also don't have a strong preference, just let me know whichever you prefer and i'm happy to change it :-)

casperstorm (approved):

PR #1969, opened 2026-05-13
Sort highlight words in reverse len order

luca020400:
If you have 2 words with the same prefix it's
important to "match" the longest one first.

Yes I admit having one such entry.


andymandias (approved):
Makes sense to me.

andymandias:
Should we add a changelog fix entry?

andymandias:
Thank you!

PR #1971, opened 2026-05-14
Sidebar Fix & Tweaks

andymandias:
Fixes a bug where the maximum height of the server icon was capped for network-provided icons.  Also ensures that the server proxy is used to fetch the icon (if configured).

Added a setting for overriding the URL (`servers.<name>.icon.override_url`).  Also adds sidebar font size settings for further customization (and so that previous defaults can be reproduced).

Tweaks the defaults so that server icons & titles are more visually prominent:

Screenshot from 2026-05-14 14-23-29


casperstorm (approved):
Happy with this PR however i'd like our defaults to stay as they are as im not personally a fan of the bigger server icons and text. Its a quite big change imo, but i think it make sense to have as an option and then id like later to re-style/layout the sidebar where we can work on some new defaults. Good job on this <3

PR #1973, opened 2026-05-16
Re-send Reply Fix

andymandias:
A fix so that replies keep their reply context when being re-sent.


furudean:
i tested this fix as i reported the bug. works on my end

casperstorm (approved):

PR #1974, opened 2026-05-16
 propagate redactions to previews

furudean:
if a replied-to message gets redacted, update the preview. not sure if this is the most optimized way to index this

### DIFF
--- a/data/src/history.rs
+++ b/data/src/history.rs
@@ -1303,12 +1303,22 @@ impl History {
                 pending_redactions,
                 ..
             } => {
-                if let Some(message) =
-                    pending_messages.iter_mut().rev().find_map(|(m, _)| {
-                        (m.id.as_deref() == Some(&*id)).then_some(m)
-                    })
-                {
-                    message.redaction = Some(redaction);
+                let found_position = pending_messages
+                    .iter()
+                    .rev()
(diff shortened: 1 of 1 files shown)


data/src/history.rs
```diff
@@ -1303,12 +1303,22 @@ impl History {
...
+                if let Some(position) = found_position {
+                    pending_messages[position].0.redaction = Some(redaction);
+
+                    let updated_reply =
+                        pending_messages[position].0.as_reply_preview();
+                    for (message, _) in pending_messages.iter_mut() {
+                        if message.reply_to.as_deref() == Some(&*id) {
+                            message.reply_preview = Some(updated_reply.clone());
+                        }
+                    }
```
andymandias: We should be able to skip updating reply previews for `History::Partial`, since they will be caught by `process_messages`→`populate_reply_previews` when the buffer is opened and the full is loaded.

andymandias (commented):
Looks good, makes sense to me.  I believe we can rely on `process_messages` to simplify it a little, but otherwise I think it's good to go.

data/src/history.rs
furudean: oh that's easy enough. much less LoC changed :-) pushed a commit

andymandias (approved):
Thank you!

luca020400:
Thanks for fixing this ^^

PR #1976, opened 2026-05-18
Add Server Actions to Command Bar

andymandias:
Adds server actions to the command bar:
- Connect to server (if disconnected)
- Disconnect from server (if connected)
- Reload server icon (removes the cached server icon, then fetches it again)

### DIFF
--- a/data/src/cache.rs
+++ b/data/src/cache.rs
@@ -87,6 +87,24 @@ impl FileCache {
         let _ = fs::write(&path, &bytes).await;
     }
 
+    pub async fn remove<T>(&self, url: &Url)
+    where
+        T: CachedAsset + DeserializeOwned,
+    {
+        let path = self.state_path(url);
+
+        if let Ok(bytes) = fs::read(&path).await
+            && let Ok(state) = serde_json::from_slice::<CacheState<T>>(&bytes)
+            && let CacheState::Ok(ref asset) = state
+        {
+            let assets = asset.assets();
+
+            remove_assets(&assets).await;
+        }
+
+        let _ = fs::remove_file(path).await;
+    }
+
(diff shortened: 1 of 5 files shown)


data/src/cache.rs
```diff
@@ -87,6 +87,26 @@ impl FileCache {
         let _ = fs::write(&path, &bytes).await;
     }
 
+    pub async fn remove<T>(&self, url: &Url)
+    where
+        T: CachedAsset + DeserializeOwned,
+    {
+        let path = self.state_path(url);
+        println!("path {path:?}");
```
luca020400: Derp

data/src/cache.rs
```diff
@@ -87,6 +87,26 @@ impl FileCache {
...
+
+        if let Ok(bytes) = fs::read(&path).await
+            && let Ok(state) = serde_json::from_slice::<CacheState<T>>(&bytes)
+            && let CacheState::Ok(ref asset) = state
+        {
+            let assets = asset.assets();
+            println!("assets {assets:?}");
```
luca020400: ditto

luca020400 (commented):
cache side looks good.

I assume in the future we may want to introduce an enumeration to fetch all states to later clear cache?

andymandias:
> cache side looks good.
> 
> I assume in the future we may want to introduce an enumeration to fetch all states to later clear cache?

Thanks for taking a look at it!  (There's an additional command I want to add, so this should actually be a draft at the moment, but the cache reload part is nominally complete.)
 
I do think that would be good to add cache clear command(s) to the command bar as well.

casperstorm (approved):

andymandias:
Going to defer the additional server command to later, in the interest of making sure these commands are in for the next release.

PR #1977, opened 2026-05-18
Show tooltip on user/channel completion

### LINKED ISSUE #1964: Disambiguate nicks if there are multiple matches during autocompletion
Let's say you have `user1` and `user2` in a channel and you want to mention `user2`'s name. You type `use` and then hit <kbd>tab</kbd>. The completed nick will be the alphabetic match, `user1`, which is not what was intended.

One option to solve this would be to have a tooltip similar to `/command`s on multiple matches. There is a subtle difference here in that we can't always autocomplete when a string of characters are in the input, but it should instead trigger the matcher on a tab keystroke.

luca020400:
Fixes #1964


furudean:
clean! i didn't find any obvious usability issues

casperstorm (approved):

PR #1981, opened 2026-05-20
reply: strip redundant nicks when attached to reply

furudean:
The already-merged setting `buffer.reply.insert_nick` prepends `alice: ` to the message input when replying to `alice`. This patch adds the new setting, `buffer.reply.hide_redundant_mentions`, which hides any such redundant highlights in the buffer view.

For clients that understand replies, we can filter these out since they're already shown above the message. For clients that don't understand replies, it downgrades gracefully.  There is precedent here - goguma adds `alice: ` to its replies and parses it as a part of its reply row, separate from the text. My hope is that other clients will hop on the bandwagon with this fallback behavior.


casperstorm (approved):
LGTM, thanks for all the work.

src/buffer/message_view.rs
```diff
@@ -98,6 +98,21 @@ pub struct ChannelQueryLayout<'a> {
...
+            .hide_redundant_mentions
+            .then(|| {
+                message
+                    .reply_preview
+                    .as_ref()
+                    .and_then(|rp| rp.user.as_ref())
+                    .map(|u| u.nickname().as_str().to_owned())
+            })
+            .flatten()
+    }
```
andymandias: Do we need to convert to a `String` here?  I think if we have something like
```rust
                if self.config.buffer.reply.hide_redundant_mentions {
                    message
                        .reply_preview
                        .as_ref()
                        .and_then(|reply_preview| reply_preview.user.as_ref())
                        .map(User::as_str)
                } else {
                    None
                }
```
the borrow checker will be satisfied.

andymandias (approved):
Looks good!

src/buffer/message_view.rs
furudean: done!

PR #1982, opened 2026-05-20
Window handling

### LINKED ISSUE #1122: Opens to seconday monitor in a single-monitor setup

casperstorm:
Should fix #1122.
This also fixes so if you have a secondary monitor positioned above the main monitor it works (since before we woudn't allow negative numbers)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -305,6 +305,22 @@ pub enum Message {
 }
 
 impl Halloy {
+    fn save_main_window_settings(&self) -> Task<Message> {
+        let main_window = self.main_window;
+
+        iced::window::position(main_window.id).then(move |position| {
+            let mut main_window = main_window;
+            if let Some(position) = position {
+                main_window.position = Some(position);
+            }
+
+            Task::perform(
+                data::Window::from(main_window).save(),
+                Message::WindowSettingsSaved,
+            )
+        })
+    }
+
     fn new(
         config_load: Result<Config, config::Error>,
         window_load: Result<data::Window, window::Error>,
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -305,6 +305,22 @@ pub enum Message {
...
 impl Halloy {
+    fn save_main_window_settings(&self) -> Task<Message> {
+        let main_window = self.main_window;
+
+        iced::window::position(main_window.id).then(move |position| {
+            let mut main_window = main_window;
+            if let Some(position) = position {
+                main_window.position = Some(position);
+            }
+
```
andymandias, line 319: This might benefit from a comment explaining why it's needed even though we update `main_window.position` on  `window::Event::Moved`.  Or, at least, I'm not entirely certain why;  is it because there might be `window::Event::Moved` still in the event queue that may never be processed due to the application closing?

andymandias (approved):
I don't have a multi-monitor setup with which to test, but looks good to me.

PR #1985, opened 2026-05-20
Allow bouncer network configuration

luca020400:
Allow to override config for bouncer networks

For example
[servers.soju.bouncer_networks."Libera.Chat"]
icon.override_url = "https://libera.chat/static/img/libera-color.svg"

Will override the icon for the specified network

### DIFF
--- a/data/src/config/server.rs
+++ b/data/src/config/server.rs
@@ -11,6 +11,7 @@ use tokio::process::Command;
 use self::bouncer::BouncerConfig;
 use self::filehost::Filehost;
 use self::icon::Icon;
+use crate::bouncer::BouncerNetwork;
 use crate::config::inclusivities::{
     Inclusivities, is_target_channel_included, is_target_query_included,
 };
@@ -211,7 +212,7 @@ impl Server {
         }
     }
 
-    pub fn bouncer_config(&self) -> Self {
+    fn bouncer_config(&self) -> Self {
         Self {
             // nickserv info not relevant to the bounced network
             nick_password_file: Option::default(),
@@ -226,6 +227,18 @@ impl Server {
             ghost_sequence: Server::default().ghost_sequence,
 
             ..self.clone()
+        }
+    }
+
+    pub fn bouncer_network_config(
+        &self,
+        bouncer_network: &BouncerNetwork,
+    ) -> Self {
+        match self.bouncer_networks.as_ref() {
+            Some(bouncer_config) => {
+                bouncer_config.overlay(bouncer_network, &self.bouncer_config())
+            }
+            None => self.bouncer_config(),
         }
     }
 }
--- a/data/src/config/server.rs
+++ b/data/src/config/server.rs
@@ -147,7 +147,7 @@ pub struct Server {
     pub filehost: Filehost,
     pub metadata: HashMap<metadata::Key, String>,
     pub icon: Icon,
-    pub bouncer: Option<BouncerConfig>,
+    pub bouncer_networks: Option<BouncerConfig>,
 }
 
 impl Server {
@@ -281,7 +281,7 @@ impl Default for Server {
             filehost: Filehost::default(),
             metadata: HashMap::default(),
             icon: Icon::default(),
-            bouncer: None,
+            bouncer_networks: None,
         }
     }
 }
--- a/data/src/config/server.rs
+++ b/data/src/config/server.rs
@@ -212,7 +212,7 @@ impl Server {
         }
     }
 
-    fn bouncer_config(&self) -> Self {
+    fn default_bouncer_config(&self) -> Self {
         Self {
             // nickserv info not relevant to the bounced network
             nick_password_file: Option::default(),
@@ -233,13 +233,14 @@ impl Server {
     pub fn bouncer_network_config(
         &self,
         bouncer_network: &BouncerNetwork,
-    ) -> Self {
-        match self.bouncer_networks.as_ref() {
-            Some(bouncer_config) => {
-                bouncer_config.overlay(bouncer_network, &self.bouncer_config())
-            }
-            None => self.bouncer_config(),
+    ) -> Server {
+        let mut config = self.default_bouncer_config();
+
+        if let Some(bouncer_config) = self.bouncer_networks.as_ref() {
+            bouncer_config.overlay(bouncer_network, &mut config);
         }
+
+        config
     }
 }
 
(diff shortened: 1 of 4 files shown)


luca020400:
I think this is more or less fine, we need to decide if this approach is sane.

data/src/config/server/bouncer.rs
```diff
@@ -0,0 +1,54 @@
...
+            }
+        }
+
+        server
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Default)]
+#[serde(default)]
+pub struct NetworkConfig {
```
4e554c4c, line 43: i don't see why we shouldn't just use `config::Server` here and handle nonsense fields in the merging step.

data/src/config/server/bouncer.rs
```diff
@@ -0,0 +1,54 @@
+use indexmap::IndexMap;
+use serde::Deserialize;
+
+use crate::bouncer::BouncerNetwork;
+use crate::config::Server;
+use crate::config::server::icon::Icon;
+use crate::config::sidebar::OrderChannelsBy;
+
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Default)]
+#[serde(default)]
```
4e554c4c:
```suggestion
#[serde(default, transparent)]
```

data/src/config/server.rs
```diff
@@ -141,6 +143,7 @@ pub struct Server {
     pub filehost: Filehost,
     pub metadata: HashMap<metadata::Key, String>,
     pub icon: Icon,
+    pub bouncer: Option<BouncerConfig>,
```
4e554c4c:
```suggestion
    pub bouncer_networks: Option<BouncerConfig>,
```

4e554c4c (changes requested):
thanks! I imagined this when i wrote the bouncer neworks impl. some initial comments.

data/src/config/server/bouncer.rs
luca020400: That is probably the main point to address, I went with allowlist because it seemed easier to reason on what allow to overlay.

On the other hand allowing it all is definitely easier, but may become a whack a mole situation

casperstorm:
Bouncer networks is in a bit of a sore spot right now so it is nice that you are attacking this. However i'd like to spend a bit of time thinking about a bigger refactor; how we can support both types of network a bit better.
You want the ease of connecting to a bouncer network, but also the ability to extend each with their own configuration, ignore list, etc. We also want to clean up the code a bit, and make the seperation easier, and easier to understand.

andymandias:
To follow up on @casperstorm note a little bit, I think this is an opportunity to make sure the `bouncer-networks` interfaces are a little bit harder to misuse.

E.g. the `is_primary` function is (a) unclear to me in what it intends to return if called for a non-bouncer-network and (b) when I was testing for a PR review was not consistent in what it returned for non-bouncer-networks connections.  For that PR we wound up adding a check for whether `bouncer-networks` was acked to the problematic conditional, but left the function as-is since I'm not well versed in the spec or impl.  If it had a form like `is_primary(..) -> Option<bool>`, where it's only some if it's a `bouncer-networks` connection, that would be helpful in guiding those of us not familiar with its intricacies.

4e554c4c:
> If it had a form like `is_primary(..) -> Option<bool>`, where it's only some if it's a `bouncer-networks` connection, that would be helpful in guiding those of us not familiar with its intricacies.

i don't really understand what the issue is here. `is_primary` is supposed to return `true` if it's a bouncer network joined by halloy (with a parent), and `false` otherwise. There are other ways to bind a bouncer connection, like connecting to `pounce/libera` on a soju instance, but these are not "managed by halloy" per-se (they are distinct config entries) and don't need to have special consideration

andymandias:
> i don't really understand what the issue is here. `is_primary` is supposed to return `true` if it's a bouncer network joined by halloy (with a parent), and `false` otherwise. There are other ways to bind a bouncer connection, like connecting to `pounce/libera` on a soju instance, but these are not "managed by halloy" per-se (they are distinct config entries) and don't need to have special consideration

Well, the issue was that it wasn't clear to me whether `is_primary` is a `bouncer-networks` attribute that is not intended to have meaning outside that context, or if its return value should have meaning outside of `bouncer-networks`.  The secondary issue is that it doesn't work the way you describe;  it does return `false` for my non-`bouncer-networks` soju connections, but I'm connecting to a few Ergo servers for which `is_primary` returns `true`.  I only suggested `Option` as a way to signal whether it has a meaning because I assumed the value was not intended to be meaningful outside the `bouncer-networks` context.

4e554c4c:
sure, this is all completely as intended. `is_primary` just means that a network isn't bound in any way, so having it return `true` for ergo instances is correct

i don't really understand what is meant by a 'bouncer networks context'. i assume you want to do something like `server.bouncer_netid.is_some_and(....)` but this would really give the wrong result, since we mostly care if a connection is _not_ primary, in which case this function would return the wrong value (none) for non-`bouncer-networks` soju connections

casperstorm:
But the fact that it has to be explained means we can do better. :)

4e554c4c:
i did try to explain this in the enormous comment above the function but yes, i agree it could be more clear

but i don't think that should block this pr since it's pretty trivial

andymandias:
I think if I'm getting the both explanations:
- `is_primary` is supposed to return `true` if it's a bouncer network joined by halloy (with a parent), and `false` otherwise
- `is_primary` just means that a network isn't bound in any way, so having it return `true` for ergo instances is correct

The ergo instance is not a bouncer network joined by halloy (with a parent), so... maybe it isn't so clear what it's supposed to return?  And why is it a different value than my other connection that isn't utilizing `bouncer-networks`?

casperstorm:
This PR wont be merged until we have a clear plan for a bigger re-write of bouncer networks. @andymandias is raising some great points in this thread, and that is just the surface. We have problems; so we will fix it so that in the end we will have a even better IRC client. @4e554c4c, you wrote most a lot of the bouncer network code, but try not to be that defensive around it - it's just code after all and we need to listen and align to make this work.

4e554c4c:
> I think if I'm getting the both explanations:
> 
> * `is_primary` is supposed to return `true` if it's a bouncer network joined by halloy (with a parent), and `false` otherwise

this is false, which might be a source of the confusion. if it's a bouncer network it's never primary (sorry for mixing up `true` and `false`, im only a mathematician)

casperstorm:
Thanks for this PR. It started a great discussion which has now resulted in https://github.com/squidowl/halloy/pull/2213 which was just merged. I'll therefor close this down.

PR #1988, opened 2026-05-21
reply: render actions

furudean:
turns out `/me` actions were not rendered properly, adding redundant nicks in places and not following the established formatting in the reply preview. this patch fixes this.


src/buffer/message_view.rs
```diff
@@ -1868,14 +1888,18 @@ impl<'a> ChannelQueryLayout<'a> {
...
+                        } else {
+                            theme::font_style::primary
+                        },
                         dimmed.map(|d| {
                             move |color: Color| d.transform_color(color, bg)
                         }),
                         self.config,
                     ),
-                    true,
+                    !is_action,
```
andymandias: This should probably remain `true` until/if we don't show previews for actions.

andymandias (commented):
Tooltip previews for actions look spot on, and code looks clean.  Appreciate your commitment to detail.  Looks like the in-history preview (and reply bar shown above the input) are still styled like regular messages, is that for a particular purpose?

furudean:
No particular reason other than it being more logic, I'll see what I can do :-)

src/buffer/message_view.rs
furudean: good catch... it doesn't seem to make a lot of sense to have them but i guess there's no good reason to change it

casperstorm (approved):

PR #1992, opened 2026-05-21
Move completion overlay on top of cursor

luca020400:


casperstorm (approved):

casperstorm:
Works great in my testing.

andymandias (commented):
Works reallly nicely for emoji/word completion.  For the command tooltip, I think it would be nice if it could stay attached to the command rather than following the cursor.  And a minor nit, if the cursor is on the left side of a word, then the tooltip gets matched to the previous word (`completion.rs::get_word` unfortunately doesn't look like it'd quite work for `current_word_start`, but maybe the logic could be consolidated between the two? or not). 

Image

andymandias (approved):
Great idea, thank you!

luca020400:
This works much better, ty!

luca020400:
I think we can drop Anchor(Point) support, this is fine-ish, the only "problem" is that the tooltip is z-above the reply row, while before it was y above it instead.
We could double_pass it to get the size... but \o/

This will never end.

andymandias (approved):
Alright, the reverted behavior (completion aligned with text input, not trying to track the caret) is now working as desired by my testing.  So this is my approval of that style, and should be ready for outside review.

luca020400:
Thanks, double_pass worked like a charm

PR #1993, opened 2026-05-21
Don't autocomplete in case there's a single word

luca020400:
Annoying.


andymandias (approved):
Looks good, works well.  Thank you!

PR #1994, opened 2026-05-21
Preview Inclusion Conditions Fix

andymandias:
Fix for preview inclusion conditions requiring both channel and user/server_message be specified, when one or the other should be sufficient.


casperstorm (approved):
Great catch.

andymandias:
Well spotted by @furudean :slightly_smiling_face:

PR #1996, opened 2026-05-22
Two-Step Flush

### LINKED ISSUE #1263: Highlight messages are not displayed in channel buffer
This looks like recurring issue. Seems to happen when two (or more?) people are highlighted in single message. When there is single user (me) highlighted seems to work ok. But I cannot know for sure, because if messages are not displayed... I would know they were there in the first place.

Looks like this for example. Notice there is message in highlights and not in the channel. 
Image

andymandias:
Clear pending messages/reactions/redactions after flush has completed, instead of when the flush task is dispatched.  This is to prevent a potential race condition between history flush and load, where a history loaded with inopportune timing could result in lost message(s).  At this time the hope is this fixes #1263, but I currently don't have a way to easily reproduce the required conditions.

### DIFF
--- a/data/src/history.rs
+++ b/data/src/history.rs
@@ -419,6 +419,9 @@ pub enum History {
         pending_reactions: HashMap<message::Id, reaction::Pending>,
         pending_redactions: HashMap<message::Id, redaction::Pending>,
         show_in_sidebar: bool,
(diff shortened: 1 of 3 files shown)


casperstorm (approved):
I think this is the right way, even though i havent tested it.

Issue #1477, opened 2026-02-14
Halloy locks up DE temporarily on launch


ashrama:
New install of Halloy.
Config as per below
When Halloy is launched, or switched to after being minimized the desktop environment is locked up with no input possible and no GUI display redraws for quite a period of time (10-30 seconds).
A similar issue happens when ALT-TAB is used to switch windows and you bypass Halloy to switch to another window. It causes a delay and no screen redraws for some period of time but not as bad as switching to the window as described above.

Halloy 2026.1.1
Manjaro KDE Plasma stable branch

```
# Halloy config.
#
# For a complete list of available options,
# please visit https://halloy.chat/configuration.html

[servers.liberachat]
nickname = "hidden_nickname"
server = "irc.libera.chat"
use_tls = true
port = 6697
channels = ["#halloy"]

[servers.IRCHighWay]
nickname = "hidden_nickname"
server = "irc.irchighway.net"
use_tls = false
port = 6667
channels = ["#matrix"]

[font]
family = "Hack Nerd Font Mono"
size = 16
weight = "light"
bold_weight = "semibold"
only_emojis_size = 18

```

casperstorm:
Hmm, this smells like a GPU driver related issue 🤔

ashrama:
> Hmm, this smells like a GPU driver related issue

What can I do to diagnose further?
nvtop says halloy is running on the dGPU


```
inxi -G                                                                                                           
Graphics:
  Device-1: Intel Kaby Lake-H GT2 [HD Graphics 630] driver: i915 v: kernel
  Device-2: Advanced Micro Devices [AMD/ATI] Polaris 22 XL [Radeon RX Vega
    M GL] driver: amdgpu v: kernel
  Device-3: Realtek Integrated_Webcam_HD driver: uvcvideo type: USB
  Display: wayland server: X.org v: 1.21.1.21 with: Xwayland v: 24.1.9
    compositor: kwin_wayland driver: X: loaded: amdgpu,modesetting
    dri: iris,radeonsi gpu: i915 resolution: 2560x1440~60Hz
  API: EGL v: 1.5 drivers: iris,kms_swrast,radeonsi,swrast
    platforms: gbm,wayland,x11,surfaceless,device
  API: OpenGL v: 4.6 compat-v: 4.5 vendor: intel mesa v: 25.3.4-arch1.1
    renderer: Mesa Intel HD Graphics 630 (KBL GT2)
[...]

Issue #1220, opened 2025-10-04
Can't start halloy


wlallemand:
I can't start halloy, it is crashing at startup.

With the binary download from github `halloy-2025.9-x86_64-linux.tar.gz` I got:

```
% RUST_BACKTRACE=full ./bin/halloy

thread 'main' panicked at /home/runner/.cargo/git/checkouts/winit-865c6d5afb3ec9c3/05b8ff1/src/platform_impl/linux/x11/window.rs:655:37:
Failed to change window theme: XError { description: "BadMatch (invalid parameter attributes)", error_code: 8, request_code: 149, minor_code: 4 }
stack backtrace:
   0:     0x5e4e05ebdf32 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::hf435e8e9347709a8
   1:     0x5e4e05eeb723 - core::fmt::write::h0a51fad3804c5e7c
   2:     0x5e4e05eb97e3 - std::io::Write::write_fmt::h9759e4151bf4a45e
   3:     0x5e4e05ebdd82 - std::sys::backtrace::BacktraceLock::print::h1ec5ce5bb8ee285e
   4:     0x5e4e05ebfc26 - std::panicking::default_hook::{{closure}}::h5ffefe997a3c75e4
   5:     0x5e4e05ebfa29 - std::panicking::default_hook::h820c77ba0601d6bb
   6:     0x5e4e05ec05b2 - std::panicking::rust_panic_with_hook::h8b29cbe181d50030
   7:     0x5e4e05ec036a - std::panicking::begin_panic_handler::{{closure}}::h9f5b6f6dc6fde83e
   8:     0x5e4e05ebe439 - std::sys::backtrace::__rust_end_short_backtrace::hd7b0c344383b0b61
   9:     0x5e4e05ebfffd - __rustc[5224e6b81cd82a8f]::rust_begin_unwind
  10:     0x5e4e046e2830 - core::panicking::panic_fmt::hc49fc28484033487
  11:     0x5e4e046e2c56 - core::result::unwrap_failed::h9e4c136384b1cfa3
  12:     0x5e4e04c41887 - winit::window::Window::set_theme::h8717facbe99bbb4d
  13:     0x5e4e04bf34ab - iced_winit::run_instance::{{closure}}::hb60f221f4c21e273
  14:     0x5e4e04bfae0f - iced_winit::run::Runner<Message,F>::process_event::hc96ecd023cc7d6cd
[...]

PR #2001, opened 2026-05-22
Runtime backend settings

### LINKED ISSUE #669: Option to turn off hardware accelration
Feature request: 

An option to turn off GPU hardware acceleration, which is on by default.
If running application, eg. games that are heavy depended on GPU, offloading ICED to use CPU rather then GPU would be preferable. 

Also with hardware acceleration off, Halloy is not killed if the graphics card drivers crashes and Nvidia overlay notifications are not shown when opening Halloy.

casperstorm:
Fixes https://github.com/squidowl/halloy/issues/669.

```toml
[runtime]
backend = "best" | "hardware" | "software"
antialiasing = true | false
vsync = true | false
```

todos:
- [x] update about after config refresh
- [ ] figure out why changing vsync/aa crashes on wayland

### DIFF
--- a/data/src/config/runtime.rs
+++ b/data/src/config/runtime.rs
@@ -1,0 +1,28 @@
+use serde::Deserialize;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Deserialize)]
+#[serde(default)]
+pub struct Runtime {
+    pub backend: Backend,
+    pub vsync: bool,
+    pub antialiasing: bool,
+}
+
+impl Default for Runtime {
+    fn default() -> Self {
+        Self {
+            backend: Backend::default(),
+            vsync: true,
+            antialiasing: false,
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub enum Backend {
+    #[default]
+    Best,
+    Hardware,
+    Software,
+}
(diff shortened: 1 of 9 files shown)


KaiKorla:
Any way to choose DX12 over VULKAN on Windows?

luca020400:
> Any way to choose DX12 over VULKAN on Windows?

Can try https://gist.github.com/luca020400/868e0e1c9dd1beb855c54c0008b5a712

casperstorm:
> Any way to choose DX12 over VULKAN on Windows?

I'll make it a option, for now you just do software, hardware or best - however ill extend so you can define it yourself before this is ready.

data/src/config/runtime.rs
```diff
@@ -0,0 +1,76 @@
...
+
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub enum HardwareApi {
+    #[default]
+    Best,
+    Vulkan,
+    Metal,
+    DirectX12,
+    OpenGL,
```
luca020400, line 31: With kebab case this gets translated to open-g-l
Ditto for the rest

casperstorm: Yeah good catch.

casperstorm:
> Any way to choose DX12 over VULKAN on Windows?

`backend = { hardware = "directx12" }`

andymandias (approved):
LGTM!

I think we could potentially say #1220 and #1477 are as solved as they can be from our side, but we could leave them open to track the upstream issue.

PR #2005, opened 2026-05-23
data: allow scheme://host URLs with no TLD (localhost, IPs)

achille:
URLs without a public TLD — `http://localhost/`, `http://intranet/`, IPs without dots, etc. — used to be left as plain text because the regex required a `.<tld>` suffix on every host. This PR makes the TLD optional in the `scheme://host` form, so common local-development URLs linkify.

- Also drops the bare `www.X` shortcut from the regex (and the matching `https://` auto-prefix in the caller) and inferring `https://` is a guess.
 
- Added 4 `fragments_parsing` test cases for `http://localhost/foo`, `http://localhost:8080/api`, `https://localhost/secure`, and `ws://localhost:9000/socket`.

### DIFF
--- a/data/src/message.rs
+++ b/data/src/message.rs
@@ -43,8 +43,11 @@ const URL_PATH_RESERVED: &str = r#":?@!&'()*+,;=\[\]"#;
 
 static URL_REGEX: LazyLock<Regex> = LazyLock::new(|| {
     RegexBuilder::new(concatcp!(
-        r#"(?i)(((https?|ircs?|wss?):\/\/|www\.)[\p{Letter}\p{Number}\-@:%._+~#=]{1,256}\.[\p{Letter}\p{Number}]{1,63}\b"#,
-        r#"(?:"#,
+        r#"(?i)("#,
+        r#"(?:https?|ircs?|wss?):\/\/"#,
+        r#"[\p{Letter}\p{Number}\-@:%._+~#=]{1,256}"#,
+        r#"(?:\.[\p{Letter}\p{Number}]{1,63})?"#,
+        r#"\b(?:"#,
         r#"["#, URL_PATH_UNRESERVED, URL_PATH_RESERVED, r#"%\/#]"#,
         r#"*)|halloy:\/\/[^ ]*)"#
     ))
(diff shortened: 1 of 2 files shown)


furudean:
Not sure if I'm reading this right, but a pattern of `www.host` seems overly broad to convert into links in text. Text in buffers may contain a variety of syntax - it's not always standard English. You could imagine there would be a lot of false positives. Very few chat apps treat strings like this like URLs.

achille:
@furudean oh ok, good point, I've updated it (force pushed it) so its a bit more simpler and minimal to review. 

But thinking about this -- different groups will have needs to support different url schemes. Would make sense to maybe expose https://crates.io/crates/linkify (see: https://robinst.github.io/linkify/ ) as a configurable options. 

E.g so folks could just configure their protocols for whatever they need.

data/src/message.rs
```diff
@@ -43,8 +43,11 @@ const URL_PATH_RESERVED: &str = r#":?@!&'()*+,;=\[\]"#;
 
 static URL_REGEX: LazyLock<Regex> = LazyLock::new(|| {
     RegexBuilder::new(concatcp!(
-        r#"(?i)(((https?|ircs?|wss?):\/\/|www\.)[\p{Letter}\p{Number}\-@:%._+~#=]{1,256}\.[\p{Letter}\p{Number}]{1,63}\b"#,
-        r#"(?:"#,
+        r#"(?i)("#,
+        r#"(?:https?|ircs?|wss?):\/\/"#,
+        r#"[\p{Letter}\p{Number}\-@:%._+~#=]{1,256}"#,
+        r#"(?:\.[\p{Letter}\p{Number}]{1,63})?"#,
+        r#"\b(?:"#,
```
englut, line 50: I think to achieve what you want (linkifying: `http://localhost:8080/api`), all you would need to do is this:

```
r#"(?i)(((https?|ircs?|wss?):\/\/|www\.)[\p{Letter}\p{Number}\-@:%._+~#=]{1,256}(\.[\p{Letter}\p{Number}]{1,63})?\b"#,
```

That last group just needs to be optional. Everything else should stay the same.

I do agree that `www.<domain>` linkifying can be an odd experience ~~, but it is already functionality that we have. I am wary about changing existing behaviour that could be expected by users~~ .

englut: Just had a chat with one of the maintainers about this. I think you're good to remove the `www.<domain>` scenario. @furudean is right, it's not common practice and it's not explicit.

In this case, I think all you need is this:

```
r#"(?i)(((https?|ircs?|wss?):\/\/)[\p{Letter}\p{Number}\-@:%._+~#=]{1,256}(\.[\p{Letter}\p{Number}]{1,63})?\b"#,
```

Which... I think you do have something very similar to this already, just including the `\b`. So all good from me! Just be sure to address the clippy issues.

Thanks for taking this on @achille 🙏

englut (approved):

casperstorm (approved):

andymandias:
Thank you!

PR #2013, opened 2026-05-24
Allow to configure broadcast

luca020400:


luca020400:
Those can be handled via `internal_messages`!

PR #2014, opened 2026-05-24
Use `UserDisplay` for quoting replies of actions

englut:
Looks like when displaying the quoted reply of an action, the metadata color is being used regardless of if `display.nickname` has `color` in in the config.

### before
Screenshot_20260524_132218

### after, without color metadata enabled
Screenshot_20260524_132229

### after, with color metadata enabled
image


andymandias (approved):
Looks good to me.  Thank you!

PR #2015, opened 2026-05-25
Allow to add internal buffers to sidebar

luca020400:


luca020400:
Currently the main annoyance is that they use Test<> icons so the alignment is wrong.
What's left is deciding if we want to use the same configs as servers icon for size/show

~~Ideally we'd allow to configure the position (before/after) servers?~~

casperstorm:
Screenshot 2026-06-11 at 19 52 01

I think this is close @luca020400, only thing i would like is that icons follow server icons 🤔.

luca020400:
> Screenshot 2026-06-11 at 19 52 01
> I think this is close @luca020400, only thing i would like is that icons follow server icons 🤔.

The logic now matches upstream_buffer, ideally it should be cleaned up...

luca020400:
image

Now in case `server_icon = "hidden"` the text isn't aligned anymore

casperstorm:
Other than what @luca020400 also brought up, this PR now looks good.

andymandias:
For now I updated it so that internal buffer icons are also controlled by the `sidebar.server_icon` setting.  I assume that's how most users would want it to function (and we don't need a separate toggle for each type of icon in the sidebar).  We could potentially alias that setting to `sidebar.icon` if that's the approach we stick with.

luca020400:
Works great, thanks for the cleanup!

casperstorm (approved):

PR #2029, opened 2026-05-27
Align Unread/Highlight Indicators and Server Badges

andymandias:
Align unread/highlight indicators and server unread/highlight badge.
Shrink default highlight_icon_size such that it matches the server badge size by default.


casperstorm (approved):
Thank you for this andy, my eyes can rest now!

PR #2033, opened 2026-05-28
More Sidebar Tweaks

andymandias:
Adjust sidebar padding/spacing to avoid obscuring sidebar buttons with the scrollbar.

Fix button alignment when sidebar is in top/bottom position.


casperstorm (approved):
Nice

PR #2044, opened 2026-05-31
Force disable auto format if channel has c mode

### LINKED ISSUE #2043: FR: Don't send formatted messages into channels that don't allow that
the `c` mode on channels blocks formatting. halloy attempts to send formatted messages to these channels anyway which causes them to look like plain text to the end user.

Halloy should ideally disable auto-format or ignore `/format` before sending messages to a `+c` channel.

luca020400:
Fixes #2043

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Unreleased
 
+Fixed:
+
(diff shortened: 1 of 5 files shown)


casperstorm (commented):
Can you add CHANGELOG entry? Code looks great.

casperstorm (approved):

Issue #2045, opened 2026-05-31
Username Highlighting In Contractions


sco1:
It seems like if there is a channel member whose name is the same as the second part of a contraction, e.g. username `t`, then messages that contain words like `can't`, `isn't`, etc. get treated as user mentions.

For example:

<img width="788" height="164" alt="Image" src="https://github.com/user-attachments/assets/f252d168-7c24-48ac-b710-9ef1ad8afcf9" />

Here's my version info:
```
Version: 2026.7
Commit: b41dde61
GPU Backend: Vulkan
GPU Adapter: NVIDIA GeForce RTX 4070 Ti SUPER
VSync: Enabled
Antialiasing: Disabled
```

andymandias:
This should be improved as of 2026.7.1+.

PR #2046, opened 2026-05-31
Redo user regex following https://modern.ircdocs.horse/#clients

luca020400:
Solves #2045


andymandias (commented):
Looks good to me. Can we get a test to cover the fixed issue? I find regex in particular is prone to unintended regressions.

andymandias (approved):
Thank you!

PR #1911, opened 2026-04-29
Tooltip to indicate if query user is offline

englut:
Fixed user's "offline" status.

Added a tooltip to show `offline` for scenarios where the colour chosen for `buffer-nickname-offline` is not super obvious (especially when nothing to contrast to).

image

PR #2047, opened 2026-05-31
Query offline indicator

englut:
Re-imagining #1911

Smarter approach for indicating if user in query buffer is offline.

Thanks to @andymandias for the idea! Adding monitoring to nicks when you open a query. Monitoring is cleaned up when query buffer is closed as well.

The various calls checks to see if monitoring is supported by the server.

Screenshot_20260531_113757

### DIFF
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -2900,6 +2900,15 @@ impl Client {
                         }
                         if should_notify { Some(user) } else { None }
                     })
+                    .filter_map(|user| {
+                        self.config
+                            .monitor
+                            .iter()
+                            .any(|monitored_user| {
+                                monitored_user == user.as_normalized_str()
+                            })
+                            .then_some(user)
+                    })
                     .collect::<Vec<_>>();
 
                 return Ok(vec![
@@ -2939,6 +2948,15 @@ impl Client {
                         }
 
                         if should_notify { Some(nick) } else { None }
+                    })
+                    .filter_map(|nick| {
+                        self.config
+                            .monitor
+                            .iter()
+                            .any(|monitored_user| {
+                                monitored_user == nick.as_normalized_str()
+                            })
+                            .then_some(nick)
                     })
                     .collect::<Vec<_>>();
 
@@ -4556,6 +4574,12 @@ impl Client {
 
     pub fn multiline_limits(&self) -> Option<MultilineLimits> {
         self.capabilities.multiline_limits()
+    }
+
+    pub fn is_monitored_user_online(&self, user: &User) -> bool {
+        self.monitored_users
+            .get(user)
+            .is_some_and(|monitored_user| monitored_user.online)
     }
 }
 
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -693,6 +693,8 @@ impl Client {
         let labeled_response_context =
             self.set_labeled_response_context(buffer, &mut message);
 
+        let mut restore_automated_monitored_users: Vec<String> = vec![];
+
         if matches!(priority, TokenPriority::User) {
             match &message.command {
                 Command::LIST(..) => {
@@ -793,7 +795,12 @@ impl Client {
                             }
                         }
                         (_, "C" | "c") => {
-                            self.monitored_users.clear();
+                            for (user, _) in self.monitored_users.iter().filter(
+                                |(_, monitored_user)| monitored_user.automated,
+                            ) {
+                                restore_automated_monitored_users
+                                    .push(user.as_normalized_str().to_owned());
+                            }
                         }
                         _ => (),
                     }
@@ -806,6 +813,26 @@ impl Client {
             anti_flood.add_token(message, priority);
         } else if let Err(e) = self.handle.try_send(message.into()) {
             log::warn!("[{}] Error sending message: {e}", self.server);
+        }
+
+        if !restore_automated_monitored_users.is_empty()
+            && let Some(isupport::Parameter::MONITOR(monitor_limit)) =
+                self.isupport.get(&isupport::Kind::MONITOR)
+        {
+            let messages = group_monitors(
+                &restore_automated_monitored_users,
+                *monitor_limit,
+                find_target_limit(&self.isupport, "MONITOR"),
+                &self.server,
+                None,
+            );
+            for message in messages {
+                if let Some(ref mut anti_flood) = self.anti_flood {
+                    anti_flood.add_token(message.into(), priority);
+                } else if let Err(e) = self.handle.try_send(message) {
+                    log::warn!("[{}] Error sending message: {e}", self.server);
+                }
+            }
         }
 
         labeled_response_context
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -2883,38 +2883,32 @@ impl Client {
                             User::parse(target, casemapping, prefix).unwrap_or(
                                 User::from(Nick::from_str(target, casemapping)),
                             );
-
-                        if let Some(monitored_user) =
-                            self.monitored_users.get_mut(&user)
-                        {
-                            monitored_user.online = true;
-                            if !monitored_user.query {
-                                Some(user)
-                            } else {
-                                None
-                            }
-                        } else {
-                            self.monitored_users.insert(
-                                user.clone(),
-                                MonitoredUser {
-                                    online: true,
-                                    query: false,
-                                },
-                            );
-                            Some(user)
-                        }
+                        let entry = self
+                            .monitored_users
+                            .entry(user.clone())
+                            .and_modify(|monitored_user| {
+                                monitored_user.online = true;
+                            })
+                            .or_insert_with(|| MonitoredUser {
+                                online: true,
+                                query: false,
+                            });
+                        (!entry.query).then_some(user)
                     })
                     .collect::<Vec<_>>();
 
-                let mut events = vec![];
-                if !targets.is_empty() {
-                    events.push(Event::Single {
-                        message: message.clone(),
-                        our_nick: self.nickname().to_owned(),
-                        deduplicate: false,
-                    });
-                    events.push(Event::MonitoredOnline(targets));
-                }
+                let events = if !targets.is_empty() {
+                    vec![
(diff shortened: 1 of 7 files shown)


src/screen/dashboard/pane.rs
```diff
@@ -561,11 +561,27 @@ fn query_title<'a>(
...
+    let is_user_offline = config.buffer.nickname.offline.is_offline(
+        shared_channels.is_empty()
+            && current_user.is_none()
+            && current_user.and_then(User::accountname).is_none()
+            && !clients
+                .client(server)
+                .is_some_and(|client| client.is_monitored_user_online(&user)),
+    );
+
+    let state = if is_user_offline {
```
englut, line 594: 🤔 Maybe should only show `(Offline)` here if the server client has ISUPPORT Monitor available.. ~Will add the conditional when I am back at my desk.~ 🙏

Edit: updated via SSH + nvim. Will test when I am back at my desk 🫡

englut: Done! Works great :saluting_face:

data/src/client.rs
```diff
@@ -2870,36 +2878,88 @@ impl Client {
...
+                        } else {
+                            self.monitored_users.insert(
+                                user.clone(),
+                                MonitoredUser {
+                                    online: false,
+                                    query: false,
+                                },
+                            );
+                            Some(nick)
+                        }
```
luca020400:
```suggestion
                        let entry = self
                            .monitored_users
                            .entry(user.clone())
                            .and_modify(|monitored_user| {
                                monitored_user.online = false;
                            })
                            .or_insert_with(|| MonitoredUser {
                                online: false,
                                query: false,
                            });

                        (!entry.query).then(|| nick)
```

Might be a bit easier to read, got lost trying to understand `query` was the deciding factor

englut: Nice! Great suggestion. I've updated to take this approach! Thank you @luca020400 🙏

src/screen/dashboard.rs
```diff
@@ -2979,6 +2980,22 @@ impl Dashboard {
...
+            && let Some(query) = upstream.query()
+            && let Some(client) = clients.client_mut(server)
+            && let user =
+                User::from(Nick::from_str(query.as_str(), client.casemapping()))
+            && let Some(server_config) = config.servers.get(server)
+            && !server_config
+                .monitor
+                .iter()
+                .any(|nick| nick == user.as_normalized_str())
+            && !client.is_monitored_user(&user)
```
andymandias: Can we use
```rust
        if let Some(buffer::Upstream::Query(server, query)) = buffer.upstream()
            && let Some(client) = clients.client_mut(server)
            && let user = User::from(Nick::from(query))
            && !client.is_monitored_user(&user)
```
here, since `is_monitored_user` should be `true` for manually requested/configured monitoring?

src/screen/dashboard.rs
```diff
@@ -3223,6 +3240,16 @@ impl Dashboard {
                 (Task::batch(tasks), None)
             }
             buffer::Upstream::Query(server, nick) => {
+                if let Some(client) = clients.client_mut(&server)
+                    && let user = User::from(Nick::from_str(
+                        nick.as_str(),
+                        client.casemapping(),
+                    ))
```
andymandias: We should be able to use:
```rust
                if let Some(client) = clients.client_mut(&server)
                    && let user = User::from(Nick::from(&query))
```

src/main.rs
```diff
@@ -1893,6 +1893,15 @@ fn handle_client_events(
             }
             Event::OnConnect(on_connect) => {
                 let server = server.clone();
+                for query in dashboard.open_pane_server_queries(&server) {
+                    if let Some(client) = clients.client_mut(&server) {
+                        let user = User::from(Nick::from_str(
+                            query.as_str(),
+                            client.casemapping(),
+                        ));
```
andymandias: We can use:
```rust
                        let user = User::from(Nick::from(query));
```

src/screen/dashboard/pane.rs
```diff
@@ -584,11 +584,27 @@ fn query_title<'a>(
...
         .is_away(current_user.is_some_and(User::is_away));
-    // It's generally not possible to tell whether another user is offline or
-    // just not in the any shared channels.  Unless/until user offline status is
-    // monitored (via IRCv3 MONITOR or otherwise), do not display users as
-    // offline in query titles as it may be misleading.
-    let is_user_offline = config.buffer.nickname.offline.is_offline(false);
+    let is_user_offline = config.buffer.nickname.offline.is_offline(
+        shared_channels.is_empty()
+            && current_user.is_none()
+            && current_user.and_then(User::accountname).is_none()
```
andymandias: Do we need these two checks? I believe`Client::get_user_channels` will return any & all channels shared with the user, so I would expect `current_user.is_none()` to always match `shared_channels.is_empty()`.

data/src/client.rs
```diff
@@ -4507,6 +4679,57 @@ impl Client {
...
+        self.monitored_users.remove(user);
+        if self.has_isupport_monitor() {
+            self.send(
+                None,
+                command!("MONITOR", "-", user.as_normalized_str()).into(),
+                TokenPriority::Low,
+            );
+        }
+    }
 }
```
andymandias, line 4736: Maybe we should add to & remove from `monitored_users` only when `self.has_isupport_monitor()`?  That way `monitored_users` will reflect the users that are (expected to be) monitored.

data/src/client.rs
```diff
@@ -363,21 +371,62 @@ impl Client {
...
+                        .filter(|existing_monitor| {
+                            let user = User::parse(
+                                existing_monitor,
+                                casemapping,
+                                prefix,
+                            )
+                            .unwrap_or(User::from(Nick::from_str(
+                                existing_monitor,
+                                casemapping,
+                            )));
```
andymandias: We're using this construction in a few places, and based on my testing on various servers I think it's the right construction.  So, maybe we should make a function for it; something like, `User::parse_or_force`?

data/src/client.rs
```diff
@@ -363,21 +371,62 @@ impl Client {
...
+                                && !self.is_monitored_user_automated(&user)
+                        })
+                        .cloned()
+                        .collect::<Vec<_>>();
+
+                    let monitored_users_keys: Vec<User> =
+                        self.monitored_users.keys().cloned().collect();
+                    for user in monitored_users_keys {
+                        if remove_monitors.iter().any(|remove_monitor| {
+                            remove_monitor == user.as_normalized_str()
```
andymandias: `remove_monitor` here may not be normalized, since it's coming from the user's config.  Maybe we could `filter_map` above, and return a `Vec<User>` since we've already constructed a `User` from each target?

data/src/client.rs
andymandias, line 6034: You've done well updating the correctness of the log message, so we should also fix the imposed limit (which was assuming all monitoring would be enacted via config).  Something like:
```rust
    if let Some(monitor_limit) = monitor_limit.map(usize::from) {
        let mut total_monitor_count = users.len();
        if let Some(num_existing_monitors) = num_existing_monitors {
            total_monitor_count += num_existing_monitors;
        }

        let remove_target_count = total_monitor_count.saturating_sub(monitor_limit);

        if remove_target_count > 0 {
            log::warn!(
                "[{server}] More users in monitor list than permitted by the server \
                      ({total_monitor_count} users in monitor list, {monitor_limit} permitted)",
            );
        }

        users.iter().take(users.len().saturating_sub(remove_target_count))
    } else {
        users.iter().take(users.len())
    }
```

data/src/client.rs
```diff
@@ -5788,16 +6011,19 @@ fn group_monitors<'a>(
     monitor_limit: Option<u16>,
     target_limit: Option<u16>,
     server: &Server,
+    num_existing_monitors: Option<usize>,
```
andymandias: Does `None` have a meaning different from `0`?

data/src/client.rs
```diff
@@ -695,6 +746,65 @@ impl Client {
...
+                        }
+                        (_, "C" | "c") => {
+                            for (user, _) in self.monitored_users.iter().filter(
+                                |(_, monitored_user)| monitored_user.automated,
+                            ) {
+                                restore_automated_monitored_users
+                                    .push(user.as_normalized_str().to_owned());
+                            }
+                        }
+                        _ => (),
```
andymandias, line 809: Should we `self.monitored_users.clear()` here after we're done constructing `restore_automated_monitored_users`?  If so, we can use `take(self.monitored_users).into_iter()`.

andymandias (commented):
I think this is basically good to go, though I have a few minor questions/nits.  Nice work!

data/src/client.rs
englut: Yes it does! I could just pass 0, figured `None` is more idiomatic though

src/screen/dashboard/pane.rs
englut: Hmmmmm you're right. I was hoping to do the whole.. "if they're authenticated, they're online". But it's the same as shared_channels.

data/src/client.rs
andymandias: Sorry, you're absolutely correct they are different  What I meant to ask was, in the context of this argument do they carry a different meaning?  E.g. the `target_limit` argument means "there is no limit" if `None`, which doesn't cleanly map onto any `u16` value (`u16::MAX` is similar, but not really the same).  The place I see `None` being passed in is in `Client::send` where my impression is that the monitor list was just cleared and `num_existing_monitors` is expected to be zero, but I may be overlooking the semantic distinction.

englut: I think in this scenario, passing 0 and having None are essentially the same. I can't think of there being a difference here... 🤔

andymandias (approved):
I think this is a really nice quality of life improvement for queries.  Thank you!

englut:
> I think this is a really nice quality of life improvement for queries. Thank you!

As always, really appreciate your time and keen reviewing eyes 😍🫶

casperstorm (approved):
AWESOME addition! Great thinking both.

PR #2058, opened 2026-06-02
Fix for thread 'main' panic due to a char boundary.

aniketkotal:
i had the following error, causing halloy to crash and not open if the buffer with the message is in view

```
thread 'main' (2569419) panicked at src/widget/message_content.rs:408:16:
end byte index 5 is not a char boundary; it is inside '\u{fe0f}' (bytes 3..6) of `☁️ 06:00 15°C │ 🌧 09:00 20°C │ ☁️ 12:00 21°C`
```
this pr fixes a panic in `strip_leading_nick` caused by slicing `text` at raw byte offsets, which broke when a message started with a multi-byte character like an emoji and `nick.len()` landed mid-character.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,14 @@
 
 Fixed:
 
+- Panic when a reply contains a multibyte character near the start of message
 - Ensure messages don't get formatted if channel has +c mode (colour filter)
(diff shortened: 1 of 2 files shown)


andymandias (approved):
Thank you!

PR #2061, opened 2026-06-04
skip `buffer.reply.insert_nick` for query buffers and self-reply

furudean:
fixes two cases where inserting a nick does not make sense:
- in query buffers there are only two users, so the need for insert_nick is lessened
- since the purpose of the text is to notify a user, there is no need to insert it for your own messages


andymandias (approved):
Nice work, thank you!

PR #2064, opened 2026-06-04
support RPL_METADATASYNCLATER

### LINKED ISSUE #2063: Postponed metadata synchronization is not supported
Per the [metadata spec](https://ircv3.net/specs/extensions/metadata#postponed-synchronization), the server may respond with `RPL_METADATASYNCLATER` on certain events such as joining a channel, indicating the client needs to manually issue a `METADATA SYNC` command to retrieve metadata. Halloy does not do this. In instances where this numeric is sent, this results in halloy not properly receiving metadata.

Steps to reproduce:

1. Have a client (let's call them "A") set a display-name on themselves
2. Have a different halloy instance (let's call them "B") join the channel, using an ircd that returns `RPL_METADATASYNCLATER`
3. B never sees that A has a display-name set, because halloy doesn't attempt to automatically issue a SYNC command in such cases

casperstorm:
Fixes #2063.


data/src/client.rs
```diff
@@ -224,6 +224,7 @@ pub struct Client {
     anti_flood: Option<TokenBucket<message::Encoded>>,
     mode_requests: Vec<ModeRequest>,
     metadata_sub_requests: HashSet<String>,
+    metadata_syncs: Vec<MetadataSync>,
```
4e554c4c:
```suggestion
    metadata_syncs: BinaryHeap<MetadataSync>,
```

data/src/client.rs
```diff
@@ -4559,6 +4599,26 @@ impl Client {
...
         }
 
+        let (ready, pending): (Vec<_>, Vec<_>) = self
+            .metadata_syncs
+            .drain(..)
+            .partition(|sync| now >= sync.ready_at);
+
+        self.metadata_syncs = pending;
+
+        for sync in ready {
```
4e554c4c:
```suggestion
        while let Some(sync) = heap.peek_mut() && sync.ready_at < now {
            let sync = PeekMut::pop(sync);
```

4e554c4c (changes requested):
i don't like `tick` so much but we should use it since we already have it. just a nit: maybe a `BinaryHeap` would be better than a `Vec`

4e554c4c (approved):

PR #2066, opened 2026-06-04
Nightly Version

andymandias:
Keep nightly version on same major & minor version as latest release to ensure it is ordered before the next stable release (rather than having it nominally be "overwritten" by the next release, since that likely will not happen).


KaiKorla:
Did we really need the NIGHTLY then? Just put it into VERSION and remove NIGHTLY?

andymandias:
We don't *need* it in either case, I was simply trying to make the deviation from the existing build process as small as possible to avoid complicating it or spending much time on it.

Edit to add/clarify: I'm not trying to shut down the possibility of a more automated nightly build process, just trying to elaborate on why this one is the way it is.

casperstorm (approved):

PR #2073, opened 2026-06-07
More Conservative Handling of Message Server Time for Chathistory

andymandias:
Record whether server-time was received as a tag with the message or was generated from local time, since presence cannot be simply inferred from non-tag contents of the message.  If generated, then it is not used as a chathistory reference.  When joining a channel or server, the corresponding JOIN/RPL_LOGGEDIN is utilized as a reference;  in that case, if there is no server-time tag on the message then the generated server-time is offset into the future in case the local time is before the server's time (any repeat messages that might result from the offset should be handled by chathistory deduplication).

### DIFF
--- a/data/src/capabilities.rs
+++ b/data/src/capabilities.rs
@@ -3,7 +3,7 @@ use std::str::FromStr;
 use std::string::ToString;
 use std::sync::LazyLock;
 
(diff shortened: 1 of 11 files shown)


casperstorm (approved):
LGTM

PR #2075, opened 2026-06-07
Add keyring support for secrets

### LINKED ISSUE #681: support reading passwords using org.freedesktop.secrets spec (a.k.a, from keyring)
_(Storing passwords in plain text may not be a good idea.)_

It would be quite easier if the app reads from the keyring (https://www.freedesktop.org/wiki/Specifications/secret-storage-spec/) using config like:

```toml
[servers.liberachat.sasl.plain]
username = "foobar"
use_keyring = true
```

If the keyring doesn't have the password, then the password can be taken from the user and stored in the keyring.

crate [libsecret](https://crates.io/crates/libsecret) can be used for this purpose

### LINKED ISSUE #855: FR: Secure secrets storage
I saw the external file secrets feature, but as it doesn't support relative paths or expansions it's a pain for portability, and it still means leaving secrets in places that are trivial to steal.

How would you feel about a PR that integrates a crate like https://crates.io/crates/keyring for cross-platform secret storage and recovery?

I'd imagine adding a configuration field such as `keyring = true` which means to lookup a secret in the secret store, and a small piece of dialog UI for accepting and updating the keyring contents.

KaiKorla:
This PR is introducing to allow secrets to be stored outside `config.toml` by adding system keyring support (Linux/FreeBSD, macOS, Windows) for SASL passwords and other passwords/secrets in the Halloy configuration file.

**It's need testing for all the cases X {Linux/FreeBSD, macOS, Windows}. See last topic**

## Add/Changes

- Added `keyring-core` with platform-specific stores for macOS, Linux/FreeBSD, and Windows.
- Added keyring config options for:
  - server passwords
  - NickServ passwords
  - channel keys
  - SASL PLAIN passwords
  - filehost PLAIN credentials
  - global and per-server proxy passwords
- Added automatic key names, for example:
  - `servers.libera.password`
  - `servers.libera.nick_password`
  - `servers.libera.sasl.plain.password`
  - `proxy.http.password`
- Added a generic password input modal for missing keyring entries.
- Extended duplicate secret-source validation and missing keyring-entry errors.
- Updated configuration docs and changelog.
- Added small unit tests for keyring config parsing and default key names.

## Things To Watch
- This does not migrate existing plaintext passwords automatically. Users need to enable `*_keyring = true` and remove the plaintext secret.
- Secrets are still resolved into the existing config structs at load time. This keeps the change small, but does not change the full in-memory secret lifecycle.
- `password`, `password_file`, `password_command`, and `password_keyring` are mutually exclusive.
- `channel_keys` and `channel_keys_keyring` cannot both be set for the same channel.
- Automatic channel key names include the channel prefix, e.g. `servers.libera.channel_keys.#halloy`.

## Request for Review & Testing
- Confirm that the default key names are intuitive.
- Confirm that the keyring service name `chat.halloy` is appropriate.
- Test the missing-keyring-entry flow:
  - set `password_keyring = true`
  - start Halloy
  - enter the password in the modal
  - verify that the config reloads successfully
- Test platform keyring behavior:
  - macOS Keychain
  - Linux Secret Service
  - Windows Credential Manager
- Test proxy passwords:
  - global proxy with `password_keyring`
  - per-server proxy with `password_keyring`
- Check reload behavior when changing a server’s `password_keyring`.

**As always: Feedback Welcome**

FR: #855 #681

### DIFF
--- a/data/src/config/keyring.rs
+++ b/data/src/config/keyring.rs
@@ -86,7 +86,34 @@ pub fn server_proxy_password_key(server: &str, kind: &str) -> String {
     format!("servers.{server}.proxy.{kind}.password")
 }
 
-pub fn get_password(key: &str) -> Result<Option<String>, Error> {
+pub async fn get_password(key: &str) -> Result<Option<String>, Error> {
+    let key = key.to_string();
+    let task_key = key.clone();
+
+    tokio::task::spawn_blocking(move || get_password_blocking(&task_key))
+        .await
+        .map_err(|error| Error::Keyring {
+            key,
+            error: error.to_string(),
+        })?
+}
+
+pub async fn set_password(key: &str, password: &str) -> Result<(), Error> {
+    let key = key.to_string();
+    let password = password.to_string();
+    let task_key = key.clone();
+
+    tokio::task::spawn_blocking(move || {
+        set_password_blocking(&task_key, &password)
+    })
+    .await
+    .map_err(|error| Error::Keyring {
+        key,
+        error: error.to_string(),
+    })?
+}
+
+fn get_password_blocking(key: &str) -> Result<Option<String>, Error> {
     ensure_default_store(key)?;
 
     let entry = keyring_core::Entry::new(SERVICE, key).map_err(|error| {
@@ -106,7 +133,7 @@ pub fn get_password(key: &str) -> Result<Option<String>, Error> {
     }
 }
 
-pub fn set_password(key: &str, password: &str) -> Result<(), Error> {
+fn set_password_blocking(key: &str, password: &str) -> Result<(), Error> {
     ensure_default_store(key)?;
 
     let entry = keyring_core::Entry::new(SERVICE, key).map_err(|error| {
--- a/data/src/config/keyring.rs
+++ b/data/src/config/keyring.rs
@@ -220,6 +220,14 @@ mod tests {
     }
 
     #[test]
+    fn password_deserializes_disabled_boolean() {
+        let fixture: PasswordFixture =
+            toml::from_str("password_keyring = false").unwrap();
+
+        assert_eq!(fixture.password_keyring, Password::Disabled);
+    }
+
+    #[test]
     fn password_deserializes_custom_key() {
         let fixture: PasswordFixture =
             toml::from_str(r#"password_keyring = "custom.secret""#).unwrap();
@@ -227,6 +235,36 @@ mod tests {
         assert_eq!(
             fixture.password_keyring,
             Password::Key("custom.secret".to_string())
+        );
+    }
+
+    #[test]
+    fn password_enabled_uses_default_key() {
+        let password = Password::Enabled;
+
+        assert_eq!(
+            password.key_or_default(|| "default.secret".to_string()),
+            Some("default.secret".to_string())
+        );
+    }
+
+    #[test]
+    fn password_custom_key_overrides_default_key() {
+        let password = Password::Key("custom.secret".to_string());
+
+        assert_eq!(
+            password.key_or_default(|| "default.secret".to_string()),
+            Some("custom.secret".to_string())
+        );
+    }
+
+    #[test]
+    fn password_disabled_uses_no_key() {
+        let password = Password::Disabled;
+
+        assert_eq!(
+            password.key_or_default(|| "default.secret".to_string()),
+            None
         );
     }
 
--- a/data/src/config/keyring.rs
+++ b/data/src/config/keyring.rs
@@ -1,0 +1,257 @@
+use serde::{Deserialize, Deserializer};
+
+use super::Error;
+
+const SERVICE: &str = "chat.halloy";
+
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub enum Password {
+    #[default]
+    Disabled,
+    Enabled,
+    Key(String),
+}
+
+impl Password {
+    pub fn is_enabled(&self) -> bool {
+        !matches!(self, Password::Disabled)
+    }
+
+    pub fn key_or_default(
+        &self,
+        default: impl FnOnce() -> String,
+    ) -> Option<String> {
+        match self {
+            Password::Disabled => None,
+            Password::Enabled => Some(default()),
+            Password::Key(key) => Some(key.clone()),
+        }
+    }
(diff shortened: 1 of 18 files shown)


luca020400:
On Linux I see

```
thread 'main' (424357) panicked at /home/luca/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.52.3/src/runtime/scheduler/multi_thread/mod.rs:91:9:
Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.
```

Full trace
```
  11: zbus_secret_service_keyring_store::store::Store::new
             at /home/luca/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/zbus-secret-service-keyring-store-1.0.0/src/store.rs:32:9
  12: data::config::keyring::set_platform_default_store
             at ./data/src/config/keyring.rs:150:37
  13: data::config::keyring::ensure_default_store
             at ./data/src/config/keyring.rs:132:5
  14: data::config::keyring::get_password
             at ./data/src/config/keyring.rs:90:5
  15: data::config::server::Sasl::set_password::{{closure}}
             at ./data/src/config/server.rs:402:21
  16: data::server::ConfigMap::new::{{closure}}
             at ./data/src/server.rs:304:18
  17: data::config::Config::load::{{closure}}
             at ./data/src/config.rs:578:10
  18: halloy::main::{{closure}}
             at ./src/main.rs:99:41
```

With

```toml
[servers.soju.sasl.plain]
username = "luca020400"
password_keyring = true
```

KaiKorla:
Hey,

thanks for testing on Linux.

I got the same on my fresh installed Linux development VM and started to investigate. I think i found the problem and it's fixed. On Linux, the Secret Service keyring backend uses a blocking API internally. Since Halloy loads config asynchronously, calling it directly can cause a nested Tokio runtime panic. The last commit is to run keyring reads/writes via tokio::task::spawn_blocking. Give it a try again.

luca020400:
> Hey,
> 
> thanks for testing on Linux.
> 
> I got the same on my fresh installed Linux development VM and started to investigate. I think i found the problem and it's fixed. On Linux, the Secret Service keyring backend uses a blocking API internally. Since Halloy loads config asynchronously, calling it directly can cause a nested Tokio runtime panic. The last commit is to run keyring reads/writes via tokio::task::spawn_blocking. Give it a try again.

Works just fine, here's the entry in case you're interested in kwallet

```json
    "Secret Service/keyring:servers.soju.sasl.plain.password@chat.halloy": {
        "$fdo_created": "1780857116",
        "$fdo_mime_type": "application/octet-stream",
        "$fdo_modified": "1780857116",
        "attributes": {
            "service": "chat.halloy",
            "username": "servers.soju.sasl.plain.password"
        }
    },
```

englut:
Works great in flatpak! Tested with kwallet & keepassxc :saluting_face:

casperstorm (approved):
Confirmed working on macOS and Windows. Did a small clean-up otherwise this works nicely.

data/src/config/keyring.rs
```diff
@@ -0,0 +1,306 @@
...
+    format!("servers.{server}.filehost.credentials.plain.password")
+}
+
+pub fn proxy_password_key(kind: &str) -> String {
+    format!("proxy.{kind}.password")
+}
+
+pub fn server_proxy_password_key(server: &str, kind: &str) -> String {
+    format!("servers.{server}.proxy.{kind}.password")
+}
```
andymandias, line 89: Server names, and to a lesser extent channel names, can contain quite a range of characters.  [`Entry::new`](https://docs.rs/keyring-core/latest/keyring_core/struct.Entry.html#method.new)'s documentation mentions `service`/`user` values could be invalid, but is vague about what constitutes a valid key name.  Understandably so, I think, since it's probably up to the keyring stores themselves.  The only restriction I was able to find in [`zbus-secret-service-keyring-store`](https://docs.rs/zbus-secret-service-keyring-store/latest/zbus_secret_service_keyring_store/), [`windows-native-keyring-store`](https://docs.rs/windows-native-keyring-store/latest/windows_native_keyring_store/), and [`apple_native_keyring_store`](https://docs.rs/apple-native-keyring-store/latest/apple_native_keyring_store/) was no empty `service` or `user`.  Is that the only limitation we expect? (Seems like it might be, from `Password::Deserialize`, but wanted to verify if I was interpreting that correctly.)

Testing on Ubuntu 26.04 LTS (GNOME) there's no issue with `' '`, `'/'`, Japanese, or Korean in the server name.

KaiKorla: Thats a good point, also. I'll think about a robust implementation for this.

KaiKorla: Hey @andymandias,

thanks for pointing this out. I took a closer look at the constraints of the backends we use:

* On Linux, credentials are handled through the Freedesktop.org Secret Service API. Its attributes are regular case-sensitive strings, so spaces, `/`, `.`, `#`, and Unicode should not be an issue.
* For Apple Keychain, the main restriction is that `service` and `account` must not be empty. These values are case-sensitive as well.
* The relevant difference is on Windows: Windows Credential Manager treats the credential `TargetName` as case-insensitive. Therefore, two keyring identifiers that differ only in letter casing refer to the same entry. The store also writes Halloy’s identifier to the native `UserName` field, which is limited to 513 UTF-16 code units. This is unlikely to affect automatically generated Halloy identifiers, but custom identifiers can exceed this limit. Halloy now validates it on Windows and returns a clear error.

I would still prefer to keep the identifiers human-readable rather than hashing or normalizing them. This makes it easier for users to manage entries through the native tools and keeps custom identifiers unchanged.

The Windows case-insensitivity is now documented as a platform limitation, with a note that portable keyring identifiers should not differ only by letter casing. I also documented the platform-specific constraints in the server and proxy configuration documentation and added tests covering typical special characters, mixed casing, and Unicode.

I hope this addresses your concern.

Thanks,
Kai

andymandias: I agree with documenting the Windows limitation rather than converting to an opaque identifier.  My expectation is vanishingly few users will run into the limitation, while many (at least, relatively speaking) will benefit from the legibility of the key names.

PR #2079, opened 2026-06-08
only show tooltip for completing nicks if there is more than one option

furudean:
it seems redundant to show it if we aren't disambiguating


luca020400 (approved):

casperstorm (approved):
Agree on this one.

PR #2082, opened 2026-06-08
Display version badge next to title on mobile

englut:
Added a badge to display the version of the docs on mobile:

Screenshot_20260608_132052

Also added `@types/node` as a dev dependency so that node types are properly used in IDE.

Reasoning is that it's not obvious what version the docs are for when viewing from mobile.

Currently, you can only view the version by opening up the hamburger menu:
image


casperstorm (approved):

casperstorm:
Thank you!

PR #2086, opened 2026-06-09
Nick highlight improvements

englut:
- Nick highlight improvements
  - Handle nick possessives `<nick>'s` (`<nick>` should be parsed)
  - Skip parsing nicks inside abbreviations (`e.g.` shouldn't have highlights)

Current state:
`quaff's` does not parse `quaff`
`e.g.` parses `g` (a user in `#halloy`) as nick.

PR fixes both scenarios (and the scenario where `e` is a user as well):
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Unreleased
 
+Fixed:
+
+- Nick highlight improvements
+  - Handle nick possessives `<nick>'s` (`<nick>` should be parsed)
+  - Skip parsing nicks inside abbreviations (`e.g.` shouldn't have highlights)
+
+Thanks:
+
+- Contributions: @englut
+
 # 2026.7.2 (2026-06-08)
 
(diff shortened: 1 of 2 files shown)


luca020400 (approved):
Looks fine on my side, tests are happy ^^

casperstorm:
Looks good, however two things i noticed while testing.

1. `'s` in channels are now broken  `join #foo's` splits `#foo's` into `#foo` + `'s`.
2. `casper... you there?` breaks as `casper` no longer highlights.

englut:
>     1. `'s` in channels are now broken  `join #foo's` splits `#foo's` into `#foo` + `'s`.

>     2. `casper... you there?` breaks as `casper` no longer highlights.

Doh. Good catch, @casperstorm :pray:. Fixed both of these and added tests.

image

casperstorm (approved):
Great improvement!

PR #2087, opened 2026-06-09
Decode html strings in preview card descriptions

englut:
added decoding html strings to preview card descriptions

| Before | After |
|--------|--------|
| Screenshot_20260608_190347 | Screenshot_20260608_190507 |


englut:
Ah. This is by design. Problem is GH passing encoded strings to `og:description`.

Closing this since this is a GH problem. :saluting_face:

PR #2088, opened 2026-06-09
Keyboard-driven message actions

### LINKED ISSUE #2154: Shortcuts for +reply
It'd be nice to have some keyboard shortcuts to quickly reply to messages :D 

in my eyes it'd make sense to copy what a few other programs (telegram iirc?) do and use something like Ctrl+<kbd>↓</kbd>/<kbd>↑</kbd> to move around messages, initiating the reply-selection using Ctrl+↑ while focused on the buffer's message input. Though, we may have to use a different modifier than <kbd>Ctrl</kbd> by default as that's already taken by scrolling to the top/bottom of a buffer. Maybe <kbd>Alt</kbd>? <kbd>Shift</kbd>?

Halloy can already highlight replied-to messages on hover, that same effect could be used to highlight the message you're *about* to reply to, so that when moving up/down with your reply it's easier to follow wtf you're doing.

and uh, I think that covers my thoughts here :)

furudean:
Adds a message focus feature, allowing messages and links to be navigated and acted on from the keyboard.

Individual URLs and `#channel`s can be selected in messages.

There is a vague idea to expand this system to other views as well, though that isn't as clear cut and/or useful, so let's keep this in buffers/queries for now.

## User-facing behavior

- `alt+up` / `alt+down` focuses nearest target in chat from that direction
- Once in focus mode, bare `up` / `down` / `left` / `right`  may also be used to navigate
- `alt+right`, `right`, `enter`, `space` opens a keyboard-navigable action menu that is analogous to the right-click menu
	- Alternatively there are direct shortcuts skip the menu: `r` reply, `=` react, `backspace` redact, `enter/space` open link, `ctrl+c` copy
- `alt+left`, `left` `shift+enter`, `shift+space` opens the nick menu
- Navigating down past the last message exits focus mode
- `esc` closes the action menu or exits focus

These keybinds are all configurable, except for esc and arrow keys inside of focus mode, which are considered built in. The configuration language for these can not easily be expressed with our current keybinding configs.

## Theming

A new `buffer.focus` theme color draws the focus border, falling back to `border_selected` if not present. The default theme was updated to carry this new property.

## Architecture considerations

A hidden widget is selected when in focus mode to take the cursor out of the input field, as that would be visually confusing. As far as I know there is no way to "unfocus" in Iced, so this is somewhat needed.

The enum for the right-click context menu and keyboard action menu was unified so they don't drift.

## Other changes

`Message::ScrollTo` will now consider what's partially visible before performing a reveal and nudge in that direction


luca020400:
~~With this PR it seems like ctrl+alt+{up,down} doesn't work anymore~~

~~I then tried~~
```toml
[keyboard]
cycle_previous_buffer = ["alt+up"]
cycle_next_buffer = ["alt+down"]
focus_message_up = ["ctrl+alt+up"]
focus_message_down = ["ctrl+alt+down"]
```

~~and with that alt+{up,down} doesn't work :confused:~~

~~Nevermind, weird shenanigans on KDE probably~~

Everything works as expected!

englut:
This is great! I have one issue, and one question!

The issue I am having:
When trying to remap `focus_*` to any non arrow key, I'm noticing the non arrow key binds do not work:
```
[keyboard]
focus_up = ["alt+shift+k", "alt+up"]
focus_down = ["alt+shift+j", "alt+down"]
focus_left = ["alt+shift+h", "alt+left"]
focus_right = ["alt+shift+l", "alt+right"]
```

In the above example, `alt+shift+<hjkl>` does not trigger focus mode. I've tried this with any combination of mods and keys. Only seems to work with `up/down/left/right`.

---

My one question is regarding overriding keybinds when in focus mode:

Should the following work?

```
[keyboard]
move_up = "alt+k"
move_down = "alt+j"
move_left = "alt+h"
move_right = "alt+l"
focus_up = ["alt+up", "alt+k"]
focus_down = ["alt+down", "alt+j"]
focus_left = ["alt+left", "alt+h"]
focus_right = ["alt+right", "alt+l"]
```
I have `alt+<hjkl>` mapped to `move_*`. Would be cool if once I'm in focus message mode, I can use `alt+<hjkl>` to navigate (override the non-focus message mode).

It's okay if this doesn't work, but for some reason I thought it would and I'm finding it doesn't.

casperstorm (approved):

andymandias (approved):
Great addition, thank you!

PR #2090, opened 2026-06-09
Decode html strings in preview card descriptions

englut:
Re-opening #2087 with a new config option:

added ability to decode html strings in preview card descriptions

```toml
# Type: boolean
# Values: true, false
# Default: false

[preview.card]
description_decode_html = true
```

Only decode html in preview descriptions for matching URLs:

```toml
[preview.card]
description_decode_html = [
    '''https?://(www\.)?github\.com/.*''',
]
```
| Before | After |
|--------|--------|
| Screenshot_20260608_190347 | Screenshot_20260608_190507 |

### DIFF
--- a/docs/configuration/preview.md
+++ b/docs/configuration/preview.md
@@ -26,7 +26,7 @@ Without literal strings, you'd have to write the above as `"\\bfoo'd\\b"`
 ```toml
 [preview]
 enabled = [
-    '''https?://(www\.)?imgur\.com/.*''', 
+    '''https?://(www\.)?imgur\.com/.*''',
     '''https?://(www\.)?dr\.dk/.*'''
 ]
 ```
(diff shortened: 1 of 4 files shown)


docs/configuration/preview.md
```diff
@@ -198,6 +198,36 @@ explicitly excluded, so this setting is only relevant when combined with the
 include = { users = ["BridgeBot"] }
 ```
 
+### `description_decode_html`
+
+The `og:description` for webpages are meant to be strings, but sometimes you may
```
luca020400: probably worth re-writing this.

og:description is always a string, but due to it being in HTML it needs to be escaped. Sometimes you may get double-escaped entries in this field, this setting will allow you to handle this scenario.

note: it may break purposely created og:description entries meant to include HTML escape codes, see below for ...domain based whitelist?

englut: Nice! Good call. I added a little warning tip box. 🙏

Stitchcraft_20260610-084222

casperstorm (approved):

PR #2093, opened 2026-06-12
Miscellaneous Fixes/Tweaks

andymandias:
A collection of minor fixes/tweaks:
- Improved legibility of the returned values of `MONITOR` list (`/monitor L`)
- Do not send duplicate `JOIN` message when using `/join`
- Do not send `MARKREAD` for non-`PRIVMSG`/`NOTICE` messages when the server does not support echoes
- Do not show unread/highlight indicators for ignored messages
- `change_mode` server messages are categorized as `actions` (i.e. `actions_dimmed` controls whether they are dimmed by default)
- Updates nightly build with `-nightly` naming across all assets

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -27,7 +27,7 @@ jobs:
             os: macos-latest
             make: bash scripts/build-macos.sh
             artifact_path: |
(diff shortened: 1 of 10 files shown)


casperstorm (approved):
Love these fixes. Good job.

PR #2097, opened 2026-06-14
Config editor buffer

### LINKED ISSUE #211: [feature request] text input ctrl + z/y undo/redo
like the title says, pressing CTRL + Z when the text input is focused should undo your last input change. pressing again should undo the second last one, and so on

CTRL + Y would be the opposite

- If your last action was deleting something: Undo the delete
- If your last action was typing: Delete the last word

casperstorm:
This introduces a config editor pane for easy edit of the config file.
You can `ctrl(cmd)+s` to save, and you can also press `refresh` or `save` in pane header.

Screenshot 2026-06-14 at 15 51 25


src/buffer/input_view.rs
```diff
@@ -397,6 +310,30 @@ pub fn view<'a>(
...
+                    }
+                    iced::keyboard::Key::Character("y") => {
+                        return Some(text_editor::Binding::Custom(
+                            Message::Redo,
+                        ));
+                    }
+                    _ => {}
+                }
+            }
+
```
andymandias: Should these be in `text_editor_key_bindings`?

src/buffer/input_view.rs
andymandias, line 198: Would it make sense to put these inside the `text_editor_key_bindings` module, so they could be re-used in the config editor?

src/buffer.rs
```diff
@@ -785,7 +819,8 @@ impl Buffer {
         match self {
             Buffer::Empty
             | Buffer::FileTransfers(_)
-            | Buffer::ChannelDiscovery(_) => Task::none(),
+            | Buffer::ChannelDiscovery(_)
+            | Buffer::ConfigEditor(_) => Task::none(),
```
andymandias: I think we should pass `PageUp` to the config editor.  Similarly with `scroll_down_page`.  Otherwise, `PageUp`/`PageDown` are inoperable in the Config Editor by default (in Linux/Windows).

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 Added:
 
 - Internal buffers (logs, highlights, etc) can be added to sidebar
+- Config editor pane for editing the config file in-app
```
andymandias, line 6: We should add an entry for undo/redo.  Very nice feature addition that needs a shout out.

src/buffer.rs
```diff
@@ -891,7 +928,8 @@ impl Buffer {
...
```
andymandias: Similar to page-up/down, I think this should work in the config editor to scroll to the bottom of the file (and `scroll_to_start` should scroll to the top of the file).

andymandias (approved):
I think this is really neat, should reduce friction for users who would rather not open a separate text editor (hopefully meaningfully so for such users), and is well integrated so we can use any benefits brought to it into the input box as well.  Case in point, the undo/redo feature added as a side quest to this PR.

I have a few nits, but I think this is a very solid initial implementation.

PR #2101, opened 2026-06-14
Don't dup message bytes before parsing

luca020400:
The first allocation may happen if the slice
isn't a valid UTF-8 string or when we run
the nom parser over it.


andymandias (approved):
Looks good to me.  Thank you!

PR #2103, opened 2026-06-15
Move highlights button out of hamburger menu

### LINKED ISSUE #1987: Move highlights out of hamburger menu
I've been using Halloy for a few months (it's fantastic 🙏) but I only discovered today that it is possible to get a buffer with all highlights in it by pressing `Ctrl + I` or selecting `Highlights` from the hamburger menu in the bottom left. I had previously assumed this was not possible, and had been manually scrolling back through chat logs whenever I reconnected to my bouncer and saw highlights on a given channel(s). I had used hilightphrase (I think? It's been a while) in the past when I used Hexchat.

Would it be reasonable/possible to move this link out of the hamburger menu and place is somewhere more prominent, like next to the hamburger menu? Even better, could we get a small visual identifier if there are unread highlights? The latter might duplicate the visual identifier we already get for each individual buffer but I still think it would be useful.

stephenfin:
Move the highlights button out of the hamburger menu and place it beside it instead. Add a visual notifier (based on the one used for channels with unread highlights) when there are unread highlights.

Fixes #1987

> [!NOTE]
> I am not a Rust developer by trade, so I can't say whether this is the best way to approach this. I'm happy to iterate on this with feedback. I'm also happy to close it out if it's particularly bad or the request is not a priority.

### DIFF
--- a/data/src/config/sidebar.rs
+++ b/data/src/config/sidebar.rs
@@ -31,6 +31,7 @@ pub struct Sidebar {
     #[serde(alias = "server_font_size")]
     pub primary_font_size: Option<u8>,
     pub user_menu: UserMenu,
+    pub highlights_button: HighlightsButton,
     pub padding: Padding,
     pub spacing: Spacing,
     pub order_channels_by: OrderChannelsBy,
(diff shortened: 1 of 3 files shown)


luca020400:
https://github.com/squidowl/halloy/pull/2015 how would that work for you?

stephenfin:
> #2015 how would that work for you?

Good timing. That's a simpler approach that works for me :+1:

luca020400:
> > #2015 how would that work for you?
> 
> Good timing. That's a simpler approach that works for me 👍

I was checking if what I did for myself also covered what you needed in some way or another :)
Didn't really want you to close the PR, that's up to maintainers to decide ^^

Issue #1132, opened 2025-08-20
Hide and/or Mute Buffers


deepspaceaxolotl:
Glad to see excluding buffers from notifications implemented (#733)! It would be great to also be able to hide buffers, or at least have a way to prevent them from showing an unread message indicator. Particularly useful in the case of bots that message a user every time they connect to a channel, cluttering up the channel list and distracting with unread indicators even when they're excluded from notifications.

If a hide action could be done as a right click menu option, I wonder if it could also be accompanied by a separate mute option that hides just the unread indicator as well as adding it to the exclude list for direct messages and file transfer requests (not necessarily highlights without doing it manually, to replicate the behaviour of messaging software such as Discord that allow one to mute a channel or server while still receiving notifications for direct mentions).

englut:
@deepspaceaxolotl with the latest version, there's an ignore feature:
https://halloy.chat/configuration/servers.html?highlight=ignore#filtersignore

This will also hide the messages from the bots though. So maybe it doesn't completely cover what you want, if you just want to mute the alerts but keep the chat messages.

deepspaceaxolotl:
I was more so thinking about buffers as a whole rather than specific users! Whether for temporary reasons or because I don't want to get a direct message every time I reconnect.

deepspaceaxolotl:
I was more so thinking about buffers as a whole rather than specific users, but that does seem to work for direct messages at least, so thank you!

englut:
> I was more so thinking about buffers as a whole rather than specific users, but that does seem to work for direct messages at least, so thank you!

For buffer specific excluding, it would make sense to add an exclude to this [unread_indicator](https://halloy.chat/configuration/sidebar.html#unread_indicator) section.

andymandias:
I closed this as unread indicators can be hidden via exclude/include in the unread_indicator settings (in the next release).  If more fine-grained control is desired, or additional functionality (not entirely sure what hiding a channel would entail), please feel free to re-open or open a new issue.

deepspaceaxolotl:
By hiding a buffer (channel? I may be using the wrong language here, forgive me), I meant hiding it from the list in the sidebar, so that it doesn't clutter it up. This would be particularly useful in the case I mentioned above since I can't find a way to stop receiving messages from some bots entirely, so they will always be there, but I will never need to read them or respond. I hope that makes it clearer? (I would reopen this issue, but I don't seem to be able to.)

[...]

PR #2105, opened 2026-06-15
Allow to show internal buffers only if unread

luca020400:


andymandias:
We have received a request for similar functionality on regular buffers (https://github.com/squidowl/halloy/issues/1132).  While we do not need to add that functionality in this PR, I do think it is worth considering a bit in the hopes of crafting a UX that can be used in both cases.  For muting channels/queries I suspect we'd want to do so by name (e.g. `"mute = [ "#halloy", "#libera" ]"` in the server configuration), or a global configuration that uses `Inclusivities`.  A similar UX could be adopted here, in which case internal buffers that do not have an unread state would be invalid entries for the setting (providing feedback at time of setting whether the setting will work how they hope it does).

luca020400:
> We have received a request for similar functionality on regular buffers (#1132). While we do not need to add that functionality in this PR, I do think it is worth considering a bit in the hopes of crafting a UX that can be used in both cases. For muting channels/queries I suspect we'd want to do so by name (e.g. `"mute = [ "#halloy", "#libera" ]"` in the server configuration), or a global configuration that uses `Inclusivities`. A similar UX could be adopted here, in which case internal buffers that do not have an unread state would be invalid entries for the setting (providing feedback at time of setting whether the setting will work how they hope it does).

I think it would be a great UX addition, may work on it in the future ^^

casperstorm:
Hmm, one small thing is that with this you could end up in a situtation where logs is gone, without the user is able to open it from either sidebar or menu.

luca020400:
> Hmm, one small thing is that with this you could end up in a situtation where logs is gone, without the user is able to open it from either sidebar or menu.

Yeah that's possible, this PR was made a bit before #2110 was uploaded so I didn't really take into account that scenario.
With that said I think it may get rather messy to add yet another piece of logic tied to history for the menu, up to you ^^

andymandias:
I believe what Discord does is have a menu entry (somewhere) to show/hide muted channels.  Perhaps we could add something like that to the sidebar menu and/or command bar?

andymandias:
> I believe what Discord does is have a menu entry (somewhere) to show/hide muted channels. Perhaps we could add something like that to the sidebar menu and/or command bar?

Added these controls, which also meant renaming from `show` to `mute`.  I think that naming will make a bit more sense once non-internal buffers can be muted.

docs/configuration/sidebar.md
```diff
@@ -167,6 +167,24 @@ Internal buffers shown in the sidebar.
 buffers = ["logs", "highlights"]
 ```
 
+### `mute`
+
+Controls when the internal buffers are muted (hidden from the sidebar).
+
+- `"never"`: Never mute the internal buffers (i.e. always internal buffers are visible).
```
luca020400: internal buffers are always visible?

luca020400 (commented):
Indeed replacing show with mute makes more sense then having both, thanks for going back to this!

docs/configuration/sidebar.md
andymandias: Sure should be, good catch.  Thanks!

luca020400:
works perfectly, thanks!

andymandias (approved):

PR #2106, opened 2026-06-15
cap token bucket refill at capacity

KaiKorla:
Use min instead of max when replenishing permits so the bucket refills gradually and never exceeds its configured capacity.

Add test for capacity clamping.


andymandias (approved):
Thank you!

PR #2108, opened 2026-06-15
Add explicit portable mode

TheDcoder:
Just a quick change to add support for explicitly enabling portable mode by setting a data directory, do we need to consider anything else for this feature? If no, does this need any more polishing or is it GTM as-is?

### DIFF
--- a/data/src/environment.rs
+++ b/data/src/environment.rs
@@ -38,9 +38,19 @@ pub fn cache_dir() -> PathBuf {
         .join("halloy")
 }
 
-/// Checks if a config file exists in the same directory as the executable.
+/// Checks if a portable dir is explicitly set or if a config file
+/// exists in the same directory as the executable.
 /// If so, it'll use that directory for both config & data dirs.
 fn portable_dir() -> Option<PathBuf> {
+    if let Some(path) = env::var_os("HALLOY_PORTABLE_DIR") {
+        let path = PathBuf::from(path);
+        if path.is_dir() {
+            return Some(path);
+        } else {
+            panic!("Given portable directory isn't valid!");
+        }
+    }
+
     let exe = env::current_exe().ok()?;
     let dir = exe.parent()?;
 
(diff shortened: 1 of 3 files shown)


data/src/environment.rs
```diff
@@ -38,9 +38,14 @@ pub fn cache_dir() -> PathBuf {
...
 }
 
-/// Checks if a config file exists in the same directory as the executable.
+/// Checks if a portable dir is explicitly set or if a config file
+/// exists in the same directory as the executable.
 /// If so, it'll use that directory for both config & data dirs.
 fn portable_dir() -> Option<PathBuf> {
+    if let Some(path) = env::var_os("HALLOY_DATA_DIR") {
+        return Some(PathBuf::from(path));
+    }
```
andymandias: Since this is overriding both config and data directory (as is done in portable mode) maybe the environment variable should be something like `HALLOY_PORTABLE_DIR`?  Or, we could have separate environment variables for overriding data and config directories, if you think that's preferable.

I also think it would be good to produce an error if there is no directory or there's no config file in the provided directory.  And probably halt, otherwise Halloy will create the potentially-mistyped directory and put a number of files into it.  An `expect` might suffice in this case, if we're only really expecting the feature to be used for development.

andymandias (changes requested):
I think this can be a useful feature for development work, particularly while we don't yet have command-line arguments.  A couple minor tweaks and I think it'll be good to go.  We should make a note of the changes in `CHANGELOG.md` as well.

data/src/environment.rs
TheDcoder: I agree with the variable name, I thought of using it first but decided to go with `DATA_DIR` because it's shorter, but I didn't take into account the fact that in normal mode the data dir is separate from the config dir :smile:.

Have added a `panic!` when the directory is not valid as well.

andymandias (approved):
Looks good to me.  Anything you'd like to change, @casperstorm?

casperstorm (approved):
LGTM, just rebased on main to resolve changelog conflict.

PR #2115, opened 2026-06-17
Internal buffers summer cleanup

luca020400:
Main change is `Add content menu to internal buffers`

The rest are misc improvements in code & UX

### DIFF
--- a/src/screen/dashboard/command_bar.rs
+++ b/src/screen/dashboard/command_bar.rs
@@ -416,7 +416,7 @@ impl std::fmt::Display for Buffer {
                         write!(f, "Open highlights")
                     }
                     buffer::Internal::ChannelDiscovery(_) => {
-                        write!(f, "Open Channel discovery")
+                        write!(f, "Open channel discovery")
                     }
                 },
                 data::Buffer::Upstream(upstream) => match upstream {
--- a/src/screen/dashboard/command_bar.rs
+++ b/src/screen/dashboard/command_bar.rs
@@ -169,10 +169,9 @@ pub enum Buffer {
     NewHorizontal,
     NewVertical,
     Close,
-    Replace(buffer::Upstream),
+    Replace(data::Buffer),
     Popout,
     Merge,
-    ToggleInternal(buffer::Internal),
 }
 
 #[derive(Debug, Clone)]
@@ -295,7 +294,7 @@ impl Buffer {
             buffer::Internal::ALL
                 .iter()
                 .cloned()
-                .map(Buffer::ToggleInternal),
+                .map(|buffer| Buffer::Replace(buffer.into())),
         );
 
         list.push(Buffer::Close);
@@ -312,7 +311,12 @@ impl Buffer {
             list.push(Buffer::Merge);
         }
 
-        list.extend(buffers.iter().cloned().map(Buffer::Replace));
+        list.extend(
+            buffers
+                .iter()
+                .cloned()
+                .map(|buffer| Buffer::Replace(buffer.into())),
+        );
 
         list
     }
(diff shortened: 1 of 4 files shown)


src/screen/dashboard/command_bar.rs
```diff
@@ -402,26 +406,30 @@ impl std::fmt::Display for Buffer {
...
+                data::Buffer::Internal(internal) => match internal {
+                    buffer::Internal::FileTransfers => {
+                        write!(f, "Open file transfers")
+                    }
+                    buffer::Internal::Logs => write!(f, "Open logs"),
+                    buffer::Internal::Highlights => {
+                        write!(f, "Open highlights")
+                    }
+                    buffer::Internal::ChannelDiscovery(_) => {
+                        write!(f, "Open Channel discovery")
```
andymandias: This was like this way from before, but I'm noticing now that it's the only one that's capitalized the buffer name (i.e. `"Open Channel discovery"` instead of `"Open channel discovery"`).  Might be worth making it consistent while we're looking at it.

src/screen/dashboard/sidebar.rs
```diff
@@ -1511,39 +1510,147 @@ fn internal_buffer_button<'a>(
...
+                    },
+                    Entry::CloseAllQueries
+                    | Entry::Connect
+                    | Entry::MarkServerAsRead
+                    | Entry::Detach
+                    | Entry::Leave
+                    | Entry::Remove => {
+                        unreachable!(
+                            "Internal buffers should not have these entries"
+                        )
```
andymandias: In `src/buffer/context_menu.rs` we use `row![].into()` as an empty placeholder for entries/conditionals we expect to be unreachable.  I think that is slightly preferable, since these are unreachable by convention rather than logic (i.e. could potentially become reachable in the future, as unlikely as that is) and we'd like to avoid panics as much as possible.  If you're up for it, this might be a good opportunity to formalize that convention in an `unexpected_entry` function (or a better name than that if you've got one)?

src/screen/dashboard/sidebar.rs
```diff
@@ -891,17 +884,22 @@ impl Entry {
...
+                    buffer::Upstream::Channel(_, _) => vec![],
+                    buffer::Upstream::Query(_, _) => vec![],
+                },
+                buffer::Buffer::Internal(_) => vec![],
+            }
+            .into_iter()
+            .chain(vec![Context, HorizontalRule])
+            .collect_vec(),
             match open {
                 None => vec![MarkAsRead, NewPane, Popout, Replace],
```
andymandias: This is another one that's not your doing, but I'm noticing now as part of the review.  Since we now allow unread markers on open buffers (configurable), it seems to me that we should move `MarkAsRead` into the `vec![]`s above (i.e. show the Mark as read entry on open buffers).

Also, we could hide the `MarkAsRead` entry when `history::Kind::from_buffer(buffer).is_none()` so it won't show for internal buffers that can't be unread.  My suggestion would be to create `history = history::Kind::from_buffer(buffer)` in `internal_buffer_button` where it can be used for `has_read` determination then later pass in `history.is_none()` to `Entry::list`, but you may have a better idea.

andymandias (approved):
Really cleanly done, I like how you subsumed `ToggleInternal` into `Replace`.  I have a couple of nits (some of which are pre-existing conditions), but this could be merged as-is.  Excellent work, thank you!

src/screen/dashboard/sidebar.rs
luca020400: I may do that as a side thing, for now row![] it is.

casperstorm (approved):

Issue #2122, opened 2026-06-19
Context menu overlay mouse click passthrough


luca020400:
If a context_menu is open at the moment if any element within the overlay don't have mouse interaction set on them we can 'hit' the element below the overlay

PR #2123, opened 2026-06-19
Properly handle mouse_interaction in menu overlay

luca020400:
- **Always use the same logic to access first children**
- **Properly handle mouse_interaction in menu overlay**


luca020400:
Fixes #2122

andymandias (approved):
Excellent, thank you!

PR #2128, opened 2026-06-23
Indicate in example docs that `sasl.external` can be used for `SSL + CertFP`

englut:
OFTC supports `sasl.external` but they call it `SSL + CertFP`. Updating the docs so this is a little more obvious for OFTC users.


casperstorm (approved):
Thanks!

PR #2134, opened 2026-06-26
Update keyboard configuration docs

edwardloveall:
When updating my own keyboard configuration, I found I wanted a few things from the existing docs. So I:

- Specified to put the keyboard config
- A list of valid modifier keys
- Rearranged the page to have a more clear (IMO) hierarchy
- Added a sentence about how to combine modifiers

I tested all the modifiers on macOS. I don't have an easy linux or windows box to test on.

### DIFF
--- a/docs/configuration/keyboard.md
+++ b/docs/configuration/keyboard.md
@@ -2,7 +2,7 @@
 
 # Keyboard
 
-Customize keyboard shortcuts. Below is a list of all actions which can be mapped.
+Customize keyboard shortcuts by adding a `[keyboard]` section to your [config.toml](../configuration.md).
 
 ```toml
 [keyboard]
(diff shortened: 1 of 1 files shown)


edwardloveall:
I'm also realizing it might be useful to specify all the [named keys](https://github.com/squidowl/halloy/blob/bb29666186dd548e74fac401bc78407b88c4ef7b/data/src/shortcut.rs#L589-L655). There's a lot, though. I could link to the source, or just do nothing for now. Let me know if you have a preference.

docs/configuration/keyboard.md
```diff
@@ -2,7 +2,7 @@
 
 # Keyboard
 
-Customize keyboard shortcuts. Below is a list of all actions which can be mapped.
+Customize keyboard shortcuts by adding a `[keyboard]` section to your [config.toml](../configuration.md). You can disable a keybind by setting it to `"noop"` or `"none"`. Each shortcut accepts either a single keybind string or an array of keybind strings.
```
englut: > You can disable a keybind by setting it to `"noop"` or `"none"`. Each shortcut accepts either a single keybind string or an array of keybind strings.

I think this makes sense to be placed after the example. That way the first sentence leads directly to an example config of `[keyboard]`

docs/configuration/keyboard.md
```diff
@@ -14,10 +14,25 @@ quit_application = "alt+q"
...
-Each shortcut accepts either a single keybind string or an array of keybind strings.
+## Modifiers
 
-## Types
+Specify what modifier keys must be pressed to invoke the action with these special strings:
+
+- `shift`: The Shift key
+- `ctrl`: The Control key
+- `alt`, `option`, `opt`: The Alt key, or Option (macOS)
+- `command`, `cmd`: The Command/⌘ key (macOS) or Control on other platforms
```
englut: > or Control on other platforms

This is redundant, the actions section does display `Default Other`..

This is more me thinking out loud.. but maybe we can change `Other` there to `Linux / Windows` since right now the `Other` kind of implies that `macOS` is first class 👀

edwardloveall: Oh, yeah, I don't want to make it seem like macOS is the dominant platform. I like your idea to use `Linux / Windows`.

englut: Ah, I was thinking more of the column header for "Default Other" underneath that. but that works there too. I think this highlights the difference for anyone who jumps between platforms (I do!) :pray: Thank you!

casperstorm (approved):
Thank you so much, this is a great addition.

PR #2136, opened 2026-06-28
fix: preserve own away state when rejoining a channel

### LINKED ISSUE #1984: Away bug after rejoining
Image

mvanhorn:
## Summary

After marking yourself away with `/away` and then leaving and rejoining a channel, your own away indicator showed you as no longer away. This preserves the client's own away state across a rejoin.

## Why this matters

Own away state was only applied to users already present in the joined channels. On rejoin the channel is recreated as `Channel::default()` (self-JOIN branch in `data/src/client.rs`) and the self user is re-added by the names list with `away: false`, so the indicator stayed stale until a WHO reply arrived (and some servers omit self away in WHO). This matches the report in #1984: (1) toggle away, (2) leave and rejoin.

The fix tracks the client's own away state on the `Client` struct, sets it in the `AWAY` / `RPL_NOWAWAY` / `RPL_UNAWAY` handlers, and re-applies it to the self user when the names list re-adds it after a rejoin (`RPL_NAMREPLY`).

## Testing

Added two unit tests in `data/src/client.rs` using the existing `test_client` harness:
- `own_away_state_survives_channel_rejoin`: self stays away after leave + rejoin, and the tracking does not leak onto other users.
- `own_unaway_state_applied_on_rejoin`: self is not away after clearing away and rejoining.

`cargo +nightly fmt --check -p data`, `cargo clippy -p data --lib --tests`, and `cargo test -p data --lib` all pass.

Fixes #1984

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Fixed:
 - Show unread indicators for panes hidden by maximize
 - Allow commands (e.g. `/join`) to be used in the input box of not-joined channels
 - Disabled menu entries now block clicks from passing through
+- Preserve own away state when leaving and rejoining a channel
 
 Changed:
 
--- a/data/src/client.rs
+++ b/data/src/client.rs
@@ -213,6 +213,7 @@ pub struct Client {
     batches: HashMap<String, Batch>,
     reroute_responses_to: Option<buffer::Upstream>,
     logged_in: bool,
+    away: bool,
     registration_step: RegistrationStep,
     capabilities: Capabilities,
     features: Features,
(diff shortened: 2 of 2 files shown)


casperstorm (approved):
This looks good to me, thanks for the contribution!

andymandias:
> Own away state was only applied to users already present in the joined channels. On rejoin the channel is recreated as `Channel::default()` (self-JOIN branch in `data/src/client.rs`) and the self user is re-added by the names list with `away: false`, so the indicator stayed stale until a WHO reply arrived (and some servers omit self away in WHO). This matches the report in #1984: (1) toggle away, (2) leave and rejoin.

Can you shed any light on which server(s) omit self away? I'd like to keep track of bugs we're working around.

PR #2143, opened 2026-07-02
Remove Image Size Check

andymandias:
Remove image size check as it is now handled by iced.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1594,7 +1594,6 @@ dependencies = [
  "html-escape",
  "iced",
  "iced_core",
- "iced_wgpu",
  "idna",
  "image",
  "indexmap 2.14.0",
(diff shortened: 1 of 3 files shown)


casperstorm (approved):

casperstorm:
@hecrj does this look ok to you as iced is now handling size check?

andymandias:
To add a few more details: previously we saw an issue where an image that was bigger than the max buffer size for wgpu would result in a panic.  We added a size check on images to avoid that scenario, but now it seems that the panic is no longer triggered.

hecrj:
It may now be handled by `wgpu`; but I don't think there have been any meaningful changes in `iced` that would explicitly handle this.

I think you should resize the preview image instead to some maximum dimensions.

PR #2145, opened 2026-07-02
Update iced

luca020400:
This moves us to an iced fork that is closer to upstream
that integrates primary clipboard handling via arboard
instead of the current revert.

https://github.com/iced-rs/iced/pull/3195 is a "feature" downgrade unless we add them back, if we want

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -926,38 +926,13 @@ dependencies = [
 ]
 
 [[package]]
-name = "calloop"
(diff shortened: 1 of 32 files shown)


andymandias (approved):
Looks good to me, and works well in my testing.  ~~@casperstorm, when time permits can you double-check that copy-paste filehost upload still works on macOS?~~  @furudean confirmed, thank you!

PR #2146, opened 2026-07-02
Set window icon also on Linux

City-busz:
The window icon is commonly used by window managers and taskbars on X11.


casperstorm:
Can you update CHANGELOG with this minor change? Then it looks good.

City-busz:
Done.

casperstorm (approved):

PR #2153, opened 2026-07-03
Add example password files to docs

englut:


docs/guides/example-server-configurations.md
```diff
@@ -28,7 +28,7 @@ nickname = "registered-user" # your name on the server
 channels = ["#registered-users-clubhouse"] # channel(s) joined on launch
 
 sasl.plain.username = "registered-user" # name used for authentication; often the same as nickname
-sasl.plain.password_file = "super-secret-password" # password used for authentication
+sasl.plain.password_file = "/path/to/super-secret-password-file" # password used for authentication
```
luca020400: Probably comment should be something like "File containing password used for authentication"

englut: Good call! Done :)

casperstorm (approved):

PR #2160, opened 2026-07-04
add/highlight_indicators

### LINKED ISSUE #1445: Exception for Highlights from Sidebar Unread Indicator Exclude
It seems that excluding a buffer from showing an unread indicator will also not show one when highlighted. It would be great to have an option to "mute" the buffer but not for highlights (as can be done with notifications, from what I can tell).

rtmongold:
Closes #1445
Decouple sidebar highlight indicators from unread exclude.  First PR here, I think I found everything that I needed to update.


andymandias (approved):
Looks good to me, thank you!  Rebased only to handle CHANGELOG conflicts.

PR #2166, opened 2026-07-06
add click_channel_discovery action config

### LINKED ISSUE #1548: Add action config for channel discovery pane
Would be a nice addition to be able to configure the action for clicking on a channel name in the `/list` channel discovery pane.

Currently it uses the buffer action buffer for `click_channel_name`, which when set as `"replace-pane"` results in the channel discovery pane being replaced and the user having to input the `/list` command again if they wish to keep browsing.

If possible it would be ideal to have something like the following with the default value being `"new-pane"`

**Example**
```
[actions.buffer]
click_channel_discovery = "new-window"
click_channel_name = "replace-pane"
click_highlight = "replace-pane"
click_username = "new-window"
```

rtmongold:
Closes #1548
Change channel discovery to use click_channel_discovery. Default click_channel_discovery to new-pane so /list stays open.


andymandias (approved):
Looks good to me, thank you!

PR #2172, opened 2026-07-09
Only-Contract Setting

andymandias:
Adds setting, on by default, to only contract an expanded message when clicking anywhere on the message (previously, clicking on any link/nickname in the message trigger the corresponding action instead).

Also changes it so the context menu for the underlying link is displayed when right-clicking on a message where link/nickname behavior is being overridden with an expand/contract action.


casperstorm (approved):
Seems good to me! Thanks.

PR #2173, opened 2026-07-09
Restructure Indicator Settings

andymandias:
Restructure indicator settings to avoid populating `sidebar.unread_indicator` with too many `highlight_*` settings.

Also adds independent `title` and `show_on_open_buffers` settings for highlights.


casperstorm (approved):
Great change.

PR #2178, opened 2026-07-12
Sidebar Tweaks

andymandias:
A few minor tweaks/fixes re: the sidebar:
- Adds a distinct icon for the "About Halloy" user menu entry
- Update the log "unread" indicator to `entypo-cancel.svg` rather than the regular unread indicator in order to more clearly signal there is a warning or error in the logs.  Also color the indicator with the warning/error color based on the highest severity log entry.
- Fix a regression with the `sidebar.max_width` setting


casperstorm (approved):
Really nice with these bug fixes

PR #2180, opened 2026-07-14
iced Update

andymandias:
Pulls in a missing commit from the previous `arboard-full-patch` branch (fixes a regression in input selection).  (Also re-orders the patch branch commits a bit to group commits by feature).


luca020400 (approved):
Oops, thanks.

casperstorm (approved):

PR #2182, opened 2026-07-14
Fix mode argument pairing for mode strings with multiple signs

zsigisti:
`mode::parse` split the mode string on the first `+` and the first `-` only, and always processed the added modes before the removed ones. Both of those break on strings servers commonly send:

- `MODE #chan -o+v alice bob` paired the arguments in the wrong order, so Halloy applied `+v alice` and `-o bob` instead of de-opping alice and voicing bob. Since the parsed modes feed the nicklist, the wrong users ended up with op/voice badges until the next WHO.
- With more than two sign groups (`+o-v+t`), everything after the second group was folded into it, so `+` was treated as a literal (unknown) mode char and `t` was removed instead of added.

Parsing is now a single left-to-right scan that tracks the current sign, so modes and their arguments stay in the order the server sent them. This also means parsed modes are no longer reordered adds-first, which only mattered to the old splitting logic.

Extended the mode test with cases for argument pairing across signs, repeated signs, and type A/B/C argument handling.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ Fixed:
 - Only play one sound at a time (if two sounds would overlap, then the second sound is skipped)
 - Allow sound files to be symbolic links
 - Queries specified in configuration correctly added to the sidebar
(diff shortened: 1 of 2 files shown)


casperstorm (approved):
Great PR, thank you so much.

PR #2183, opened 2026-07-14
Do not reconnect when the server closes the connection after a requested quit

### LINKED ISSUE #2126: Connects automatically when the error code from manual disconnect
This bug when I disconnect manually and on chat has connection lost with error and several seconds connects automatically

zsigisti:
When a quit is requested, Halloy waits for the server to acknowledge it. An `ERROR` reply is already handled (`quit_requested` is checked and autoconnect is disabled), but many servers just close the connection after `QUIT`, or the close beats the `ERROR` reply. That path went through the generic connection-error arm, which unconditionally re-enabled autoconnect — so a manual quit showed a "connection lost" error and then reconnected a few seconds later.

The connection-error arm now checks `quit_requested` the same way the `ERROR` arm does: treat the close as a valid acknowledgement, report a clean disconnect, and don't reconnect.

A clean EOF without an error was already covered by the quit-request timeout, so this only changes the case where the closing connection surfaces as an error (connection reset and the like).

Closes #2126

Note: this touches the same match arm as #2181 refactors (which keeps `autoconnect = true` there), so whichever lands second will need a trivial rebase.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ Fixed:
 - Only play one sound at a time (if two sounds would overlap, then the second sound is skipped)
 - Allow sound files to be symbolic links
 - Queries specified in configuration correctly added to the sidebar
+- Do not reconnect automatically when the server closes the connection in response to a requested quit
 
 Changed:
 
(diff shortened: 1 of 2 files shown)


andymandias (approved):
Thank you!

PR #2184, opened 2026-07-14
Update file upload guide and channel notifications docs

englut:
- Updated fileupload guide to use proper casing required for ergo's isupport 
- Updated `channel` description for notification types to be a little more visually prominent


andymandias (approved):
Thank you!

PR #2188, opened 2026-07-16
Specify Accept Header when Requesting Previews/Icons

andymandias:
Sets accept headers for preview and server icon fetch requests, favoring IANA-registered image types that are known to be supported by iced.


luca020400 (approved):
Makes sense to me

casperstorm (approved):
LGTM

PR #2189, opened 2026-07-16
IRC Protocol Logging

andymandias:
Initial implementation of IRC protocol message logging.  Logging is controllable per-server, written to plain text files in directories named for the server.  Log files are named with the local date the message was received (`YYYY-MM-DD.log`).

### DIFF
--- a/data/src/stream.rs
+++ b/data/src/stream.rs
@@ -606,7 +606,7 @@ async fn connect(
     proxy: Option<config::Proxy>,
 ) -> Result<(Stream, Client), connection::Error> {
     let logger = if config.log_irc_protocol {
-        let (sender, receiver) = tokio::sync::mpsc::channel(100);
+        let (sender, receiver) = tokio::sync::mpsc::unbounded_channel();
 
         tokio::task::spawn(log_codec(server.clone(), receiver));
 
@@ -637,15 +637,15 @@ async fn connect(
 
 async fn log_codec(
     server: Server,
-    mut receiver: tokio::sync::mpsc::Receiver<CodecLog>,
+    mut receiver: tokio::sync::mpsc::UnboundedReceiver<CodecLog>,
 ) {
     let log_writer = LogWriter::new(&server).await;
 
     match log_writer {
         Ok(mut log_writer) => {
-            let mut timeout_receive = false;
+            let mut set_timeout_to_flush = false;
 
-            while let Some(action) = if timeout_receive {
+            while let Some(action) = if set_timeout_to_flush {
                 tokio::time::timeout(
                     Duration::from_millis(500),
                     receiver.recv(),
@@ -658,11 +658,11 @@ async fn log_codec(
                 match action {
                     Ok(message) => {
                         log_writer.write(message).await;
-                        timeout_receive = true;
+                        set_timeout_to_flush = true;
                     }
                     Err(_) => {
                         log_writer.flush().await;
-                        timeout_receive = false;
+                        set_timeout_to_flush = false;
                     }
                 }
             }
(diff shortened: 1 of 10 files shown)


data/src/stream.rs
```diff
@@ -522,8 +527,19 @@ async fn connect(
     config: Arc<config::Server>,
     proxy: Option<config::Proxy>,
 ) -> Result<(Stream, Client), connection::Error> {
+    let logger = if config.log_irc_protocol {
+        let (sender, receiver) = tokio::sync::mpsc::channel(100);
```
casperstorm: hmm, don't we want to use an unbounded_channel here, to avoid silently dropping log entries if the channel fills up?

irc/src/codec.rs
```diff
@@ -24,12 +35,26 @@ impl Decoder for Codec {
             // Guard against a peer that never sends CRLF: without a cap the framed stream would
             // buffer the "line" without bound, exhausting memory from a single connection.
             if src.len() > MAX_LINE_LENGTH {
+                if let Some(logger) = &self.logger {
+                    let _ = logger.try_send(CodecLog::Received(
```
casperstorm: and if we had unbounded sender we could `send(...)` here instead.

data/src/stream.rs
```diff
@@ -541,6 +557,130 @@ async fn connect(
...
+            .open(self.log_dir.join(log_file))
+            .await
+            .ok()
+            .map(|writer| (date, BufWriter::new(writer)));
+    }
+
+    fn format(message: CodecLog, received_at: DateTime<Local>) -> String {
+        let (message_content, direction) = match message {
+            CodecLog::Received(content) => (content, "RECEIVED"),
+            CodecLog::Sent(content) => (content, "  SENT  "),
```
4e554c4c:
```suggestion
            CodecLog::Received(content) => (content, "<-"),
            CodecLog::Sent(content) => (content, "->"),
```

data/src/stream.rs
```diff
@@ -541,6 +557,130 @@ async fn connect(
...
+        };
+
+        format!(
+            "{} {} -- {}",
```
4e554c4c:
```suggestion
            "{} {} {}",
```

4e554c4c (commented):
just a nit: i like how goguma logs these messages

thanks! this will help debugging a lot.

data/src/stream.rs
casperstorm: I like it verbose now that we are talking about logs, as @andymandias commited.

andymandias:
> just a nit: i like how goguma logs these messages

Now configurable!

At some point it might be nice to include a ZNC formatting style as well, but since a format specification didn't immediately show up in a search on their wiki I've put it off for now.

4e554c4c:
i'd prefer not using ZNC since it's designed to only log some kinds of messages, and only some parts of messages (e.g. prefixes, tags, WHO, etc are unlogged). if you want to go ahead with this i think reading the soju code for its znc parser is pretty easy, and i [used to maintain a branch that supports tagmsg/ircv3 tags](https://git.sr.ht/~pounce/soju/tree/taglog/item/msgstore/znclog/writer.go) but it still loses a lot of info.

fwiw the znc devs are open to logging tags i just haven't gotten around to sending them a pr, but either way i don't think it's the best for this use-case.

casperstorm (approved):

andymandias:
ZNC-style plain text storage of messages has been requested by users for a while, even with the coming conversion of the message store to a database.  Since we are not planning to allow configuration of the message store format, providing plain text storage feature would need to be a secondary output.  While I agree it's a little weird to fit it in this setting, I think its implementation should (a) make it very clear that it's not the message store and (b) be decoupled from regular history so we don't need to consider it at all in when restructuring for the database conversion.  Both of those criteria are met here, so to me it's a candidate, but let's leave it for another time regardless.  Appreciate the notes on reference implementations to check out.

PR #2191, opened 2026-07-17
Server sidebar visibility, config, toggles and indicators

### LINKED ISSUE #2175: Collapsible networks
I'm on a couple of networks with many many channels. And I'm finding myself wanting to collapse the network so all the channels for that network are hidden to have a cleaner sidebar.

I'm not too familiar with iced's UI elements yet, so I'm unsure if there's a collapse/expand functionality the way other UI frameworks would. Hopefully it does have it and it's not too difficult to implement.

Considerations:
- when collapsed, we probably need to show indicators for `unread` and `highlights` if there are _any_ channels, queries or server buffers in that network that have `unread` or `highlights`.
- click to collapse/expand: either double click, or a dedicated chevron icon button (up & down, or left & down?) right aligned on the network entry?
- I can see someone wanting to set whether or not the network is initially collapsed / expanded via config. would mean either adding `collapsed = true` to `[server.<network>]` or maybe having a `sidebar.collapsed = ["<network>"]`. probably server specific? 
- maybe the whole ability to collapse/expand should be configurable too (enabled/disabled, if disabled, everything is expanded). in case people want a cleaner sidebar without all this.
- unsure how horizontal sidebars would work tho. maybe can still collapse/expand but chevron buttons have different directions (left and right)?

englut:
Addresses #2175

[Screencast_20260716_184746.webm](https://github.com/user-attachments/assets/c8c837e8-0c93-4b0a-ab30-3cceca7aa779)

Not in this PR, saving for next iteration:
- Collapse/expand icons to indicate the collapsed state of the server
- Allow these icons to be buttons to toggle the collapsed state
- Perhaps these buttons should light up based on if there's unread or highlights in the server


src/screen/dashboard/sidebar.rs
```diff
@@ -87,11 +91,14 @@ pub enum Event {
     Connect(Server),
     Remove(Server),
     ShowMutedBuffers(bool),
+    CollapseServer(Server),
```
luca020400: Maybe ToggleCollapse?

The action will be unified and would only depend on current status (ofc if that's easily available, otherwise this will do)

englut: Good call. I've merged Collapse and expand together!

Also, cleaned up the UI a little too:

[Screencast_20260717_124440.webm](https://github.com/user-attachments/assets/bdbe0365-34de-4f1a-b789-ccd9618c3e78)

casperstorm (approved):

src/screen/dashboard/sidebar.rs
```diff
@@ -989,90 +1040,234 @@ impl Entry {
         }
 
         entries.sort();
```
luca020400, line 1063: Wouldn't it be better to stop using the Enum order as the source of truth and start handling it completely from code here (i.e. no .sort())

It felt weird when I first touched this code, and I'm pretty sure I made sure the order the entries are added is the "proper" one already

englut: Yeah I purposely avoided using this. Am not sure if the order is needed elsewhere, so I left it. I wouldn't mind just having the order be the vec order, and dropping the sort altogether.

andymandias (approved):
Great work you two!

PR #2192, opened 2026-07-17
Reroute follow ups

casperstorm:
This is a follow up, based on suggestions from @mgdudas.

* It is now possible to add `follow = true` to a reroute rule to reroute the message to the focused buffer.

eg
```toml
[servers.home.reroute]
notice = [
  { user = "ChanServ", target = { channel = "#foo" }, follow = true },
]
```

This will reroute ChanServ messages into your active buffer IF the buffer is on the same server - if not it will "fallback" to your target.

* Rerouted message has a small icon to indicate they are rerouted (similar to bots). The color of the message is (still) tertiary.

image


src/widget/user_display.rs
```diff
@@ -340,20 +359,33 @@ impl UserDisplayData {
...
+        if self.bot_icon || self.reroute_icon {
             row![
                 text_piece(self.left, font.clone()),
-                icon,
+                self.bot_icon.then(|| icon_piece('\u{1F916}')),
+                self.reroute_icon.then(|| {
+                    icon::reroute()
+                        .color_maybe(style.color)
+                        .line_height(line_height)
+                }),
```
andymandias: Should we use `icon_piece` here?

andymandias (approved):
Looks good to me!

src/widget/user_display.rs
casperstorm: Good point, i switched to `icon_piece`

PR #2193, opened 2026-07-18
Minor Log Files Tweaks

andymandias:
Adds setting `logs.max_file_count` to allow the user to control how many Halloy log files are kept.  Default is set to four log files (i.e. logs for the four most recent times Halloy was open) to make it more likely a log file will be available when users report issues in `#halloy`.  Also makes some minor changes to log message formatting to improve scannability.

### DIFF
--- a/data/src/log.rs
+++ b/data/src/log.rs
@@ -2,14 +2,18 @@ use std::cmp::Ordering;
 use std::path::PathBuf;
 use std::{fs, io};
 
-use chrono::{DateTime, Utc};
+use chrono::{DateTime, Local, Utc};
 use serde::{Deserialize, Serialize};
 
 use crate::config::logs::LevelFilter;
 use crate::environment;
 
 pub fn file() -> Result<fs::File, Error> {
-    let path = path()?;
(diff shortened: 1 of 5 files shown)


data/src/log.rs
```diff
@@ -19,14 +23,37 @@ pub fn file() -> Result<fs::File, Error> {
...
+        fs::create_dir_all(&dir)?;
     }
 
-    Ok(parent.join("halloy.log"))
+    Ok(dir)
+}
+
+pub fn clear(number_of_logs_to_keep: usize) {
+    if let Ok(dir) = dir() {
+        for (index, dir_entry) in walkdir::WalkDir::new(dir)
```
luca020400, line 38: This seems the perfect job for glob, if one of our deps pulls it in maybe we can use it instead?

andymandias: `cargo tree --target all -i glob` turns up nothing, unfortunately.

luca020400: Welp I tried.

casperstorm (approved):

PR #2194, opened 2026-07-18
Fix Regression with Config Editor

andymandias:
Fixes regression with config editor that prevented it from being focused.  Also fixes focus transfer when opening the config editor as a new pane.


casperstorm (approved):
Thanks Andy

PR #2196, opened 2026-07-20
Suppress connection status messages during macOS sleep

casperstorm:
This suppresses connection status message while macOS is sleeping

Screenshot 2026-07-20 at 17 20 46

### DIFF
--- a/src/system/macos.rs
+++ b/src/system/macos.rs
@@ -51,6 +51,8 @@ struct Registration(Retained<Observer>);
 impl Registration {
     fn new(sender: mpsc::UnboundedSender<Event>) -> Self {
         let observer = Observer::new(sender);
+        // Workspace associated with the process, shared by all threads
+        // of the app.
         let notification_center =
             NSWorkspace::sharedWorkspace().notificationCenter();
 
--- a/src/system/macos.rs
+++ b/src/system/macos.rs
@@ -19,6 +19,8 @@ define_class! {
     // SAFETY: NSObject has no subclassing requirements and the ivars are Send.
     #[unsafe(super(NSObject))]
     #[name = "HalloyPowerObserver"]
+    // An instance variable (ivar) is a variable that exists and holds its
+    // value for the life of the NSObject.
     #[ivars = Ivars]
     struct Observer;
 
--- a/src/system/macos.rs
+++ b/src/system/macos.rs
@@ -1,0 +1,99 @@
+use futures::channel::mpsc;
+use futures::{SinkExt, StreamExt};
+use iced::Subscription;
+use objc2::rc::Retained;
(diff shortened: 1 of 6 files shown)


src/system/macos.rs
```diff
@@ -0,0 +1,99 @@
...
+    }
+}
+
+struct Registration(Retained<Observer>);
+
+impl Registration {
+    fn new(sender: mpsc::UnboundedSender<Event>) -> Self {
+        let observer = Observer::new(sender);
+        let notification_center =
+            NSWorkspace::sharedWorkspace().notificationCenter();
```
andymandias, line 57: From <https://developer.apple.com/documentation/appkit/nsworkspace/shared>, which I presume this lines up with, maybe this would help describe in what way the workspace is shared?
```suggestion
        // Workspace associated with the process, shared by all threads
        // of the app.
        let notification_center =
            NSWorkspace::sharedWorkspace().notificationCenter();
```

src/system/macos.rs
```diff
@@ -0,0 +1,99 @@
...
+        Self(observer)
+    }
+}
+
+impl Drop for Registration {
+    fn drop(&mut self) {
+        let notification_center =
+            NSWorkspace::sharedWorkspace().notificationCenter();
+
+        // SAFETY: This observer was registered with this notification center.
```
andymandias:
```suggestion
        // SAFETY: This observer was registered with this notification center in
        // Registration::new.
```

src/system/macos.rs
```diff
@@ -0,0 +1,99 @@
...
+
+struct Ivars {
+    sender: mpsc::UnboundedSender<Event>,
+}
+
+define_class! {
+    // SAFETY: NSObject has no subclassing requirements and the ivars are Send.
+    #[unsafe(super(NSObject))]
+    #[name = "HalloyPowerObserver"]
+    #[ivars = Ivars]
```
andymandias, line 24: Maybe of mild use, from [Encapsulating Data](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/EncapsulatingData/EncapsulatingData.html), as it seems to describe the expected lifetime?
```suggestion
    // An instance variable (ivar) is a variable that exists and holds its
    // value for the life of the NSObject.
    #[ivars = Ivars]
```

andymandias (approved):
Not familiar with Objective C and unable to test, but looks like it makes sense from what I can glean from what Objective C documentation I've found.  Wrote a few potential comments as potential signposts for other folks like myself unfamiliar with Objective C, but I leave it to your judgment whether they're accurate/descriptive enough to use.

All the rest looks good to me!

PR #2202, opened 2026-07-23
Muteable Channels & Queries

### LINKED ISSUE #1132: Hide and/or Mute Buffers

andymandias:
Expands the `channels` and `queries` settings to allow upstream buffers to be muted.  Modifies the internal buffers `mute` setting to allow individual buffers to be muted.

### DIFF
--- a/data/src/config/actions.rs
+++ b/data/src/config/actions.rs
@@ -52,6 +52,7 @@ pub struct Sidebar {
     pub channel: Option<BufferAction>,
     pub query: Option<BufferAction>,
     pub focused_buffer: Option<BufferFocusedAction>,
+    pub cycle: CycleAction,
 }
 
 #[derive(Debug, Default, Clone, Deserialize)]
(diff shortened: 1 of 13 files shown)


casperstorm (approved):
Great addition

data/src/config/actions.rs
```diff
@@ -191,3 +192,17 @@ pub enum ImageClickAction {
...
     Preview,
 }
+
+#[derive(Debug, Copy, Clone, Default, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub enum CycleAction {
+    #[default]
+    IntoCollapsed,
+    SkipCollapsed,
+}
```
englut, line 202: 🔥 this is awesome, I was thinking the cycling is kind of odd, and you've already fleshed it out!

PR #2203, opened 2026-07-23
Treat WHOX "0"/"*" account sentinel as no account in from_whoreply

luca020400:
`User::with_accountname` already maps the WHOX no-account sentinels ("*" and
"0") to `None`, but `User::from_whoreply` (used for the InitialJoin WHO poll,
`%tcuhnfa,999`) stored the account field verbatim. As a result a logged-out
user whose WHOX account field is "0" is shown as if authenticated
("Authenticated as 0", with the padlock).

This only shows when the server/bouncer advertises `no-implicit-names`,
which is what makes Halloy use the InitialJoin poll and route the reply through
`from_whoreply` instead of `with_accountname`.

This patch normalises the sentinel in `from_whoreply` the same way `with_accountname`
does.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,7 +78,7 @@ Changed:
 
 Thanks:
 
-- Contributions: @englut, @luca020400, @classabbyamp, @KaiKorla, @TheDcoder, @City-busz, @rtmongold, @Gelbpunkt, @zsigisti
+- Contributions: @englut, @luca020400, @classabbyamp, @KaiKorla, @TheDcoder, @City-busz, @rtmongold, @Gelbpunkt, @zsigisti, quark
 - Bug reports: @luca020400, agent314, @FlooferLand, @englut, @rexbinary
 - Feature requests: @FlooferLand, quark, @daniiooo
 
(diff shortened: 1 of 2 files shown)


englut (approved):
I am not able to reproduce this. But I don't see the harm in adding the guard.

Fastidious:
> I am not able to reproduce this. But I don't see the harm in adding the guard.

This happens only when the server or bouncer advertises no-implicit-names. That is what makes Halloy use the InitialJoin poll (WHO ... %tcuhnfa,999), and its reply is handled by from_whoreply(Some(account)) rather than with_accountname(). Against a server that doesn't advertise no-implicit-names, Halloy uses the %tcnfa,99 poll, which goes through with_accountname() and all is fine.

englut:
I've managed to reproduce! It only appears in query titles for logged out users that are in a channel you've joined. Not a scenario I tested :sweat_smile: 

This PR fixes the issue. Thanks for catching this before we released `no-implicit-names` :pray:

andymandias (approved):
Thank you!

PR #2216, opened 2026-07-26
Configurable avatar size

### LINKED ISSUE #2215: Avatar rendering too small
I'm implementing an IRC Server that supports user avatars. Sure they get displayed in Halloy but the size of them is incredibly small, which wouldn't be a problem in itself but the scaling doesn't smoothe it out. It looks like this currently:

Image

It looks really bad compared to say, Discord, for instance. Like this its practically unusable unless people upload tiny images.

casperstorm:
Fixes #2215.

Adds `metadata.avatar.size` to configure avatar size. Increased the default size is 64 pixels.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ Added:
 Fixed:
 
 - Remove blank space above the input after marking a buffer as read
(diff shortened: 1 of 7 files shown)


andymandias (approved):
Looks good to me!

PR #2218, opened 2026-07-26
Remember sidebar visibility

### LINKED ISSUE #2156: [feature] Sidebar starts hidden
We have a lot of customization and it is amazing.
Sidebar can current be hide/toggle using command "Toggle Sidebar" / `toggle_sidebar` (`⌘ + ⌥ + b`, `ctrl + alt + b`). 
But sidebar will always start opened. I'd like to be able to set the status, so it can start the app with sidebar hidden instead.
Or, it could be statefull (like nicklist seems to be): relaunch the app will always use/remember last state.

casperstorm:
Fixes #2156

### DIFF
--- a/src/screen/dashboard.rs
+++ b/src/screen/dashboard.rs
@@ -175,10 +175,16 @@ impl Dashboard {
     }
 
     pub fn restore(
-        dashboard: data::Dashboard,
+        mut dashboard: data::Dashboard,
         config: &Config,
         main_window: &Window,
     ) -> (Self, Task<Message>) {
+        if !config.pane.restore_on_launch {
+            dashboard.pane = data::Pane::Empty;
+            dashboard.popout_panes.clear();
+            dashboard.focus_buffer = None;
+        }
+
         let (mut dashboard, task) =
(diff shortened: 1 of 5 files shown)


src/screen/dashboard.rs
```diff
@@ -133,7 +133,7 @@ impl Dashboard {
         let (main_panes, pane) =
             pane_grid::State::new(Pane::new(Buffer::Empty));
 
-        let (sidebar, sidebar_task) = Sidebar::new();
+        let (sidebar, sidebar_task) = Sidebar::new(false);
```
andymandias, line 136: We currently use `Dashboard::empty` in `src/main.rs` when `config.pane.restore_on_launch` is set to `false`.  I think using `false` here is correct, but we should use `Dashboard::restore` whenever we are able to load the `Dashboard` state then skip pane restoration within `Dashboard::restore` when configured as such.  That way we won't have to remember anything beyond able to load → `restore` and unable to load → `empty`.

andymandias (approved):

PR #2221, opened 2026-07-27
Fix blank space above input after buffer is marked as read

rollecode:
Scrolling to the bottom of a buffer with unread messages leaves a gap above the input that is not there when a message simply arrives.

`Column::push` skips void children, but an empty `row![]` or `column()` is not void, so it still claims a `spacing(line_spacing)` gap. With unread messages the backlog divider and the `new` column both have content; marking as read empties them, and their spacing stays behind as blank space at the end of the buffer.

Parts are now only pushed when they render something.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+Fixed:
+
+- Remove blank space above the input after marking a buffer as read
+
 # 2026.8 (2026-07-24)
 
 Added:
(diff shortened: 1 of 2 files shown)


furudean:
I got a blank space baby, and I'll write your name
-- Taylor Swift

casperstorm (approved):
Thanks looks good. Corrected CI and updated CHANGELOG.

PR #2222, opened 2026-07-27
Add config file path in config editor

luca020400:
Always better know which one you're touching!

### DIFF
--- a/src/buffer/config_editor.rs
+++ b/src/buffer/config_editor.rs
@@ -265,6 +265,10 @@ pub fn view<'a>(
     config: &'a Config,
     theme: &'a Theme,
 ) -> Element<'a, Message> {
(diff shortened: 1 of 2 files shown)


src/buffer/config_editor.rs
```diff
@@ -297,11 +301,15 @@ pub fn view<'a>(
...
     };
 
-    let mut info =
-        row![position, error_or_section, Space::new().width(Length::Fill)]
-            .spacing(8)
-            .padding(padding::bottom(6))
-            .align_y(iced::Alignment::Center);
+    let mut info = row![
+        file_path,
+        position,
```
casperstorm, line 308: Kinda feel like position should be first, or perhaps two rows:

```
[file_path - fill_space - fill_space - position]
[error_or_section]
```

casperstorm (approved):

PR #2226, opened 2026-07-27
Add buffer.nickname.offline = "dimmed"

rtmongold:
Fixes #1783

`buffer.nickname.offline` can be `"dimmed"` / `{ dimmed = ... }` the same way as `away`. Offline nicks no longer pick up `away` styling.

### DIFF
--- a/data/src/config/buffer/nickname.rs
+++ b/data/src/config/buffer/nickname.rs
@@ -1,12 +1,14 @@
-use serde::Deserialize;
+use serde::{Deserialize, Deserializer};
 
 use crate::buffer::{Alignment, Brackets, Color};
-use crate::config::buffer::{AccessLevelFormat, Away, HideConsecutive};
+use crate::config::buffer::{
+    AccessLevelFormat, Alpha, Dimmed, HideConsecutive,
+};
 
 #[derive(Debug, Clone, Deserialize)]
 #[serde(default)]
 pub struct Nickname {
-    pub away: Away,
+    pub away: Alpha,
     pub offline: Offline,
     pub color: Color,
     pub brackets: Brackets,
@@ -21,7 +23,7 @@ pub struct Nickname {
 impl Default for Nickname {
     fn default() -> Self {
         Self {
-            away: Away::default(),
+            away: Alpha::default(),
             offline: Offline::default(),
             color: Color::default(),
             brackets: Brackets::default(),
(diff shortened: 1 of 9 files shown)


data/src/config/buffer/nickname.rs
```diff
@@ -35,17 +35,55 @@ impl Default for Nickname {
...
 
-#[derive(Debug, Clone, Copy, Default, Deserialize)]
-#[serde(rename_all = "kebab-case")]
+#[derive(Debug, Clone, Copy, Default)]
 pub enum Offline {
+    Dimmed(Dimmed),
     #[default]
     Solid,
     None,
 }
```
andymandias, line 44: I believe we wind up losing one of the possibilities here, where offline nicknames are both dimmed and their color is overridden by the theme color.  (In my testing, at least, it's one or the other, and based on the enum I think we'd have to infer dimming or color from elsewhere.)  We could fit that information into the enum, but maybe it'd be cleaner to have it `Offline` be a struct with `color` and `alpha` fields?  Then `color` could be `from-theme` or `from-nickname` maybe?  That would also allow us to re-use the `Away` enum for the `alpha` field (in which case it would probably make sense to rename the `Away` enum in `Alpha`).  How does that sound?

src/appearance/theme/selectable_text.rs
```diff
@@ -138,18 +138,20 @@ pub fn nickname(
...
+        let background = theme.styles().buffer.background;
+        let color = match is_offline {
+            Some(buffer::Offline::Solid) => metadata_color,
+            Some(buffer::Offline::Dimmed(dimmed)) => {
+                dimmed.transform_color(metadata_color, background)
+            }
+            Some(buffer::Offline::None) | None => {
+                nickname_alpha(metadata_color, is_away, background)
+            }
+        };
```
andymandias, line 159: Here we're not overriding the nickname color for the `Offline::Solid` variant.  That's what was previously done so not your fault, but I think that's an oversight.  I.e. `metadata_color` should be treated the same as any other nickname color.

src/screen/dashboard/pane.rs
```diff
@@ -843,8 +843,10 @@ fn query_title<'a>(
                 .client(server)
                 .is_some_and(|client| client.is_monitored_user_online(&user)),
     );
+    let show_offline = is_user_offline
+        .is_some_and(|o| !matches!(o, data::config::buffer::Offline::None));
 
-    let state = if is_user_offline {
+    let state = if show_offline {
```
andymandias: Appreciate the attention to detail, but I want to note: if we do restructure the settings as suggest, then I think we can revert this.  When the naming for `offline` was `"none"` I can see maybe removing all hint of the user as offline, but once the scope is reduced to color I think we can keep the `Offline` parenthetical.

src/widget/user_display.rs
```diff
@@ -321,7 +321,14 @@ impl UserDisplayData {
...
-        self.render(style, is_offline, size, selectable, theme, line_height)
+        let show_offline = config
+            .buffer
+            .nickname
+            .offline
+            .is_offline(is_offline)
+            .is_some_and(|o| !matches!(o, data::config::buffer::Offline::None));
+
+        self.render(style, show_offline, size, selectable, theme, line_height)
     }
```
andymandias, line 325: Same note as in `src/screen/dashboard/pane.rs`.

andymandias (commented):
Thank you for taking this on!  I have a couple suggestions, but overall looking good!

data/src/config/buffer/nickname.rs
rtmongold: Yeah that makes sense — that'll make this a lot more flexible. I'll start working on the changes. Thanks,

src/widget/user_display.rs
rtmongold: removed.

src/screen/dashboard/pane.rs
rtmongold: Reverted to is_user_offline.

src/appearance/theme/selectable_text.rs
rtmongold: Done — metadata path now uses theme nickname_offline (with metadata fallback) when offline color is theme, same as other nicknames.

data/src/config/buffer/nickname.rs
rtmongold: Done. - offline is now { color, alpha } (theme / nickname + reused Alpha, formerly Away). Default "solid" stays theme color with no dim; you can combine with e.g. offline = { color = "theme", alpha = "dimmed" }.

andymandias:
Looks great, cleanly done.  Thank you!

andymandias (approved):

ncfavier:
Thanks!

PR #2230, opened 2026-07-28
Instructions for keyring access when running flatpak

englut:
Adding the permissions to make our keyring feature work has been [rejected by flathub reviewers](https://github.com/flathub/org.squidowl.halloy/issues/48).

I understand their concerns, and at some point, I am hoping that [keyring-rs](https://github.com/open-source-cooperative/keyring-rs) supports the portal pathway rather than the direct secrets access it currently provides (via the [zbus-secret-service-keyring-store](https://github.com/open-source-cooperative/zbus-secret-service-keyring-store) store we use.

In the mean time, I am hoping this blurb in the keyring guide is enough to steer flatpak users in the right direction.

TL;DR: run the following to fix permissions for keyring access:
```bash
flatpak override org.squidowl.halloy --talk-name=org.freedesktop.secrets --user
```

### DIFF
--- a/assets/flatpak/org.squidowl.halloy.json
+++ b/assets/flatpak/org.squidowl.halloy.json
@@ -14,8 +14,7 @@
         "--socket=fallback-x11",
         "--socket=pulseaudio",
         "--socket=wayland",
-        "--talk-name=org.freedesktop.Notifications",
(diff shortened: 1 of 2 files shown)


andymandias (approved):
Thank you!

PR #2236, opened 2026-07-30
Include ignored nicknames in autocomplete

### LINKED ISSUE #1920: Add ability to autocomplete ignored nicknames.
As per title. Current behaviour is that if nickname is on ignore list it's not possible to tab-complete it when writing a message.
It would be nice if the autocomplete still worked on those nicks.

rtmongold:
Fixes #1920

Nickname Tab-completion no longer skips users on the ignore list. Ignore filters still hide their messages (and typing): only completion changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Fixed:
 Thanks:
 
 - Contributions: @rollecode, @luca020400, @rtmongold
+- Feature requests: @daniiooo
 
 # 2026.8 (2026-07-24)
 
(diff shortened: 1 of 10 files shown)


andymandias:
Thank you for the PR. Since the primary function of ignore is to filter out user(s) I think we should make this behavior optional/configurable.

rtmongold:
I agree I think that is a better approach — added buffer.text_input.autocomplete.include_ignored (default false).

andymandias (approved):
Nicely done, thank you!  Rebased to clear CHANGELOG conflict and add thanks for the feature request, then I took the opportunity to clean up some pre-existing function names a bit.

PR #2247, opened 2026-08-01
iced Update

andymandias:
Includes new patch commit that prevents a text editor from being horizontally scrolled when the cursor is moved before it is shown.  I.e. prevents text in input box being scrolled/clipped 2px on the left on pane load or when navigating history.


casperstorm (approved):
Awesome!

PR #2252, opened 2026-08-01
Configurable server character encoding (support ISO-8859-15)

murapadev:
## Summary

- Adds a `servers.<name>.encoding` config setting so a server's connection can be decoded/encoded with any [WHATWG encoding label](https://encoding.spec.whatwg.org/#names-and-labels) instead of always assuming UTF-8. Defaults to `"utf-8"`, so existing configs are unaffected.
- Primary motivation is `iso-8859-15` support for legacy/regional IRC networks that don't speak UTF-8, but any label recognized by `encoding_rs` (e.g. `windows-1252`) works.
- `irc::Codec` now decodes incoming bytes and encodes outgoing messages through the configured `&'static encoding_rs::Encoding` before/after the existing `proto` parser/formatter, which are untouched and keep working on `String`/`&str`.
- Documented in `docs/configuration/servers.md` under the Connection section, alongside `root_cert_path`.
- Treated as a connection setting: changing it triggers the same reconnect path as other options in `has_same_connection_settings`.

## Test plan

- [x] `cargo test -p irc -p data` — all existing tests pass, plus a new `irc::codec::tests::encodes_and_decodes_iso_8859_15` round-trip test (encodes a message containing `é`, `à`, `€` as ISO-8859-15 bytes, then decodes it back and asserts equality with the original).
- [x] `cargo clippy -p irc -p data --all-targets` — clean.
- [x] `cargo fmt --check -p irc -p data` — clean.
- [x] `cargo check --workspace` — builds successfully.


classabbyamp:
https://halloy.chat/contributing#ai

Issue #2259, opened 2026-08-03
Feature request: modifier key to open buffers in new window


dpedu:
It would be convenient if holding a modifier key when clicking a buffer would cause the buffer to open in a new window instead of the default behavior of opening in the main window.

I think this behavior would be convenient and feel natural to users as it mirrors some behaviors already present on major desktop operating systems.

For example: I use Halloy on macOs. On macOs, holding `Command` key as the open-in-new-window modifier is a familiar pattern as, for example, all major browsers on the platform (Chrome, Safari, Firefox) use it as follows: holding this key and clicking a link or submitting a form cause the browser to do so in a new window.

I believe Windows and many desktop Linuxes use the `Control` key for this behavior.

Edit: some other places that would make sense to do this at the same time:
- clicking a user's name in the chat panel (query the user in a new window instead of a new pane in the same window)

PR #2261, opened 2026-08-03
expand actions.sidebar behavior options with modifier key variants 

dpedu:
Halloy already has a set of options under [actions.sidebar](https://halloy.chat/configuration/actions#sidebar) that controls what happens when clicking a generic buffer, channel, query, or focused_buffer listed in the sidebar.

This PR adds similar `*_with_modifier` options which configures what behavior happens when performing the above action while holding a modifier key. With these changes, the default behavior will be such that the clicked buffer/channel/query/etc is opened in a new window if a modifier key is held while clicking its name. The original behavior (click without modifier key) is retained.

The "modifier key" is the key that iced's [Modifiers.command](https://docs.rs/iced/latest/iced/keyboard/struct.Modifiers.html#method.command) defaults to.

This partially satisfies #2259. 

I still need to ~~update the docs~~ and other meta-items (am I supposed to add an entry to the changelog?).

### DIFF
--- a/data/src/config/actions.rs
+++ b/data/src/config/actions.rs
@@ -53,6 +53,12 @@ pub struct Sidebar {
     pub query: Option<BufferAction>,
     pub focused_buffer: Option<BufferFocusedAction>,
     pub cycle: CycleAction,
+    #[serde(default = "default_channel_with_modifier")]
+    pub channel_with_modifier: BufferAction,
+}
+
+fn default_channel_with_modifier() -> BufferAction {
+    BufferAction::NewWindow
 }
 
 #[derive(Debug, Default, Clone, Deserialize)]
--- a/data/src/config/actions.rs
+++ b/data/src/config/actions.rs
@@ -53,11 +53,14 @@ pub struct Sidebar {
     pub query: Option<BufferAction>,
     pub focused_buffer: Option<BufferFocusedAction>,
     pub cycle: CycleAction,
-    #[serde(default = "default_channel_with_modifier")]
-    pub channel_with_modifier: BufferAction,
+    #[serde(default = "default_buffer_with_modifier")]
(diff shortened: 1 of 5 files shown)


casperstorm:
Thank you for this PR!
Looking at the [actions documentation](https://halloy.chat/configuration/actions), I wonder if adding fields like `buffer_with_modifier` and `channel_with_modifier` will become difficult to maintain in the future. Already it feels like theres a disconnect naming wise in our docs :D

What about model clicks and their modifiers directly instead?

```toml
[[actions.sidebar.click]]
buffer = "replace-pane"
channel = "replace-pane"

[[actions.sidebar.click]]
modifiers = ["command"]
buffer = "new-window"
```

We could use the same structure for `buffer` and `nicklist` clicks as well

```toml
[[actions.buffer.click]]
channel_name = "replace-pane"

[[actions.buffer.click]]
modifiers = ["command"]
channel_name = "new-window"

[[actions.nicklist.click]]
modifiers = ["command"]
username = { open-query = "new-window" }
```

This would be a larger configuration change, but it seems easier to extend than adding a new `*_with_modifier` option for every action. What do you think?

dpedu:
I like the idea! It definitely appears cleaner. 

I see you used an array for the `modifiers` value. To me, this implies that multiple modifier keys could be listed here if the user wanted to hold multiple keys to perform an action. Also, the user could list `[[actions.buffer.click]]` as many times as needed with different sets of modifier keys for various action too.

This direction looks much more customizable and flexible too. I'll work on this.

casperstorm:
Thanks, just ping if there is any questions! Awesome than you are up for the task.
I don't know if we want `modifiers` as an array, i just thought that perhaps it could unlock even more customization if needed.

Issue #2260, opened 2026-08-03
Nickname highlighting defeated by brackets


ncfavier:
Neither `(ncf)` nor `<ncf>` highlight me, `ncf`. Those characters cannot be part of a nickname, so they should.

englut:
I hear ya. Unfortunately, if I am reading it correctly, the [specs](https://modern.ircdocs.horse/#nick-message) indicate that the `<` or `>` characters are valid for nicks. How many servers actually allow them, is another matter :/

I am open to making an exception, but I would like to others to weigh in before implementing one.

ncfavier:
I am not aware of any servers that allow parentheses or angle brackets in nicknames. If supporting this is desired, the list of allowed characters could be configured per-server.

englut:
@ncfavier chatted with @andymandias, and I have a PR open that addresses this issue, while still respecting the valid nick chars from the spec!

When you get the chance, could you test out the PR? 🙏

ncfavier:
Closed by https://github.com/squidowl/halloy/pull/2262

PR #2262, opened 2026-08-04
Restrict commonly used (`()<>"`) chars from being used to highlight nicks

englut:
Addresses #2260.

UPDATE: Alrighty, round two of this PR. I have dropped the reparsing logic and update the REGEX.

We now rely on the USER_REGEX to handle highlighting nicks. Because of this, we are also restricting these commonly used characters from being part of highlighted nicks: `(`, `)`, `<`, `>`, `"` (the `'` case is still handled by the possessive stripping logic).

image

---
~~Because [specs](https://modern.ircdocs.horse/#nick-message) state that `<` and `>` are valid nick chars, we don't strip surrounding brackets from nicks. To make this work, after parsing for user fragments, I take these reparse them again to strip any wrapping delimiters around nicks. Seems to work okay. The bonus for this is we get any wrapping delimiters (both paired and symmetrical delimiters).~~

~~I do worry I am adding too much logic to parsing fragments, which was mostly regex. From my testing this has been performant. But perhaps someone else could test this out and see if they notice any performance degradation :pray:~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Fixed:
 - Text in input box being scrolled/clipped 2px on the left on pane load or when navigating history
 - Expand `$HOME` or `%AppData%` when first component in a path setting
 - Don't show another preview from the same message after hiding one
+- Restrict commonly used (`()<>"`) chars from being used to highlight nicks
 
 Changed:
 
@@ -30,7 +31,7 @@ Changed:
 Thanks:
 
 - Contributions: @rollecode, @luca020400, @rtmongold, @tranzystorekk, @englut
-- Bug reports: sebbu, @dpedu, kurwavidae
+- Bug reports: sebbu, @dpedu, kurwavidae, @ncfavier
 - Feature requests: @daniiooo, @fabricionaweb, @RoboDanjal, @AbandonedCranium, tbo
 
 # 2026.8 (2026-07-24)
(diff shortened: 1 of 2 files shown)


ncfavier:
I think `(quaff foo)` should match as well (and `>quaff<`, and so on).

4e554c4c:
rfc2812 allowed special characters in nicks are "[", "]", "\", "`", "_", "^", "{", "|", "}", parentheses and angle brackets are technically not allowed

obviously we should be generous with what we receive protocol wise (as the horse docs say) but im ok if we break highlighting for nicks with those characters when including them creates so many special cases

englut:
@ncfavier 
> I think `(quaff foo)` should match as well (and `>quaff<`, and so on).

Ah, definitely for the first one, but the second one.. depends on if those are invalid chars or not.

> rfc2812 allowed special characters in nicks are "[", "]", "", "`", "_", "^", "{", "|", "}", parentheses and angle brackets are technically not allowed
> 
> obviously we should be generous with what we receive protocol wise (as the horse docs say) but im ok if we break highlighting for nicks with those characters when including them creates so many special cases

RFC2812, does seem to indicate they're invalid. But the horse docs add in an extra "MAY use".. practically, I've not seen this to be in use on any servers.. but that's likely anecdotal. I am open to changing the USER_REGEX to just reject those chars from nick highlights wholesale. Would definitely make this a lot simplier.

englut:
Okay, I've updated this PR to prevent common chars from being included in highlighting nicks.

So far, we're only restricting these chars:

- `(`
- `)`
- `<`
- `>`
- `"`

@ncfavier your different scenarios should be addressed fully now.

Let me know if there's other chars that should be restricted from being used in highlighting.

@4e554c4c I have also updated the USER_REGEX so it's clear which restricted chars are due to specs vs our highlighting.

ncfavier (approved):
This works for me, the implementation is much simpler, and this should match what other IRC clients do. 👍

andymandias (approved):
Thank you for looking after this!

PR #2268, opened 2026-08-06
better support for larger font sizes in modals

### LINKED ISSUE #2265: Raised font size causes text to be mashed together or cut off
As can be seen in the screenshots below, having text size set to (in my case, 22, with comic code, but the same happens with default font) will cause the text to either mash into other text on the same line, or be cut off

Allowing larger font sizes is a pretty important accessibility feature, so the ui being able to handle it would be quite nice

Screenshot of the About popup where the contact key and emails are mashed together into an unreadable mess

Screenshot of the theme editor, where the dropdown menu for keys has values cut off

casperstorm:
Fixes #2265

### DIFF
--- a/src/screen/dashboard/theme_editor.rs
+++ b/src/screen/dashboard/theme_editor.rs
@@ -59,10 +59,11 @@ impl ThemeEditor {
         main_window: &Window,
         config: &Config,
     ) -> (Self, Task<window::Id>) {
+        let size = iced::Size::new(555.0, 300.0)
+            * theme::scale_for_font_size(&config.font);
         let (window, task) = window::open(window::Settings {
-            // Just big enough to show all components in combobox
-            size: iced::Size::new(555.0, 300.0),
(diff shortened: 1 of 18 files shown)


src/screen/dashboard/theme_editor.rs
```diff
@@ -59,10 +59,11 @@ impl ThemeEditor {
...
         config: &Config,
     ) -> (Self, Task<window::Id>) {
+        let size = iced::Size::new(555.0, 300.0)
+            * theme::scale_for_font_size(&config.font);
         let (window, task) = window::open(window::Settings {
-            // Just big enough to show all components in combobox
-            size: iced::Size::new(555.0, 300.0),
-            resizable: false,
+            size,
+            resizable: true,
```
andymandias, line 67: Would it make sense to set a `min_size` now that it's resizable?

andymandias (approved):

Issue #1505, opened 2026-02-18
Add gif preview support


BigOjisan:
Add the ability to preview gifs along with the already implemented jpg and png images.

Also add a way to reopen the preview, if closed accidentally and to open the preview for images when clicking on links.
 Use case being not having the inline image take up space on smaller screens but still getting a preview without switching to the browser, maybe with shift+click or some other, ideally configurable, keybind.

sebbu2:
maybe it should also support png (yes it's standard since june 2025) or webp animations (if not already supported, I haven't checked).

casperstorm:
I removed myself from this as it is focused on gif, however the hide/unhide previews has been merged to main.

englut:
Could we use @tarkah's [iced_gif](https://github.com/tarkah/iced_gif) to animate gifs? I will try to implement this and consider some sane playback UX & settings 🙏

PR #2272, opened 2026-08-09
Animate gifs for previews

englut:
Addresses #1505.

Made use of @tarkah's [iced_gif](https://github.com/tarkah/iced_gif) library. But because we have forked iced, we also need `iced_gif` to use our fork, so this PR relies on the [PR to the squidowl fork](https://github.com/squidowl/iced_gif/pull/1) of `iced_gif`.

---

Here is a screencast of an animated gif with my config for `preview.image.animation` set to `"animate-on-preview-action"`.

[Screencast_20260820_103122.webm](https://github.com/user-attachments/assets/85833f24-0775-4c5a-8d59-191ee48f2e0b)


src/buffer/message_view.rs
```diff
@@ -147,6 +147,29 @@ impl<'a> ChannelQueryLayout<'a> {
...
+            return false;
+        }
+
+        let Some(parsed) = url::Url::parse(url).ok() else {
+            return false;
+        };
+
+        self.previews.get(&parsed).is_some_and(|state| match state {
+            preview::State::Loaded(preview) => {
+                preview.clone().is_gif()
```
luca020400: Cloning the whole image just to see if it's a gif sounds expansive?

src/screen/dashboard.rs
```diff
@@ -1661,10 +1668,11 @@ impl Dashboard {
             },
             Message::LoadPreview((url, Ok(preview))) => {
                 log::trace!("Preview loaded for {url}");
+
                 if let hash_map::Entry::Occupied(mut entry) =
-                    self.previews.entry(url)
+                    self.previews.entry(url.clone())
                 {
-                    *entry.get_mut() = preview::State::Loaded(preview);
+                    *entry.get_mut() = preview::State::Loaded(preview.clone());
```
luca020400: Ditto here

src/widget.rs
```diff
@@ -178,20 +178,47 @@ pub mod button {
...
                 .border_radius(if round_corners { 4 } else { 0 })
                 .content_fit(content_fit)
-                .into(),
+                .into()
+        };
+
+        match data.format {
+            data::image::Format::Raster(image_format) => match image_format {
+                image::ImageFormat::Gif => {
+                    gif(data.path.clone(), max_animation_loops)
```
luca020400: Can probably do data.path.as_path() and avoid the clone

luca020400:
<3

furudean:
I'm not sure the `preview.image.max_animation_loops` setting makes sense. Gifs define their own looping behavior as a part of the file format. What is the use case for overwriting it?

If this is an accessibility accommodation, I would suggest making gif playback on hover an option (or the default even)? This is a pretty standard pattern to deal with that that users expect.

englut:
> I'm not sure the `preview.image.max_animation_loops` setting makes sense. Gifs define their own looping behavior as a part of the file format. What is the use case for overwriting it?
> 
> If this is an accessibility accommodation, I would suggest making gif playback on hover an option (or the default even)? This is a pretty standard pattern to deal with that that users expect.

This arose cause I can see people wanting to disable animated gifs totally. It was suggested that we introduce a max loop concept to have the gifs loop until that number is reached, then allow replay. Using `0` for the disabled state I am wanting to introduce.

Personally, I like the idea you suggest better, remove max_animation_loops, and when animations are disabled (`preview.image.animation_enabled`?), only allow playback on hover. 

I would like one of the maintainers to weigh in on this though before I go ahead with this change.

andymandias:
As someone who would turn off GIF animations (or *maybe* set `max_loops` to 1), I would consider autoplay-on-hover to be only slightly less user-hostile than autoplaying forever because the GIF author said to.  I am probably oversensitive because a certain large corporation that popularized that method keeps turning that "functionality" back on after I turn it off, but I do not like having to weave the pointer between images to avoid potentially triggering animations.  Not opposed to it being available functionality, but it needs an off switch.

englut:
~~So that leaves us with three states:~~

~~- animate~~
~~- animate-on-hover~~
~~- disabled~~

~~if this is agreeable, I will make the change :saluting_face:~~

Actually. When clicking the gif, we can have the preview animate. So maybe we don't even need a animate-on-hover? I guess we'd need a way for the user to know it's an animation.

luca020400:
I like the idea of
- disabled
- animate
- animate-on-click

For the click user affordance a simple "play" button as an overlay when the user hovers with the mouse on top of the image?
And you'd stop the anim with the "same" action, just display the stop icon?

englut:
Okay, I will wait a day or so, to give people time to chime in before I implement. Here's what I am proposing, for animation config options (and functionality tied to those options):

- `disabled` (completely; avatars too)
- `animate-on-modal` (disabled in buffer, animate in modal; avatars still animate)
- `animate` (default; animate everywhere)

This next one is TDB. depending on how easy it is to set up a hovered trigger (maybe it's super simple. I haven't looked yet):
- `animate-on-hover` (disabled in buffer until it's hovered; modal & avatar still animate)

The options `animate-on-modal` and `animate-on-hover` will have an overlay indicator that this is an animation.. I am open to ideas how. Maybe `GIF` or something small. I prefer not to have a play/stop button overlay, as that means there are two different click actions for the gif (open modal, click to play/stop). I am not super firm on this though, so if people feel strongly about wanting a click to play/stop, I can implement that instead.

furudean:
I like the direction of this so far.

I wonder if message focus should be treated the same way as hover? Maybe this is what "modal" refers to, I'm not completely sure I understand the meaning in this context.

englut:
> I like the direction of this so far.
> 
> I wonder if message focus should be treated the same way as hover? Maybe this is what "modal" refers to, I'm not completely sure I understand the meaning in this context.

I am referring to the modal that pops up when you click an image preview. You can see me doing it around the end of the video I posted in the PR description.

You would need [`preview.image.action`](https://halloy.chat/configuration/preview#action) to be set to `preview`.

Maybe I should use `animate-on-preview-action`.

englut:
image

Updated! Will test this for a bit before I am satisfied it's ready for PR :saluting_face:

englut:
Waiting on https://github.com/squidowl/iced_gif/pull/2 to take advantage of iced rebase.

src/buffer/context_menu.rs
```diff
@@ -644,7 +644,7 @@ impl Entry {
                 Entry::ShowPreview,
                 Context::Url(UrlContext { url, message, .. }),
             ) => Some((
-                "Show preview".to_string(),
+                "Show Preview".to_string(),
```
englut: Maybe this was not relevant to the PR... But I had noticed that some entries, like these two preview entries, are capitalized case and others are sentence case. Should we be using capitalized case for entries?

data/src/config/preview.rs
```diff
@@ -310,7 +310,7 @@ impl Card {
 #[serde(rename_all = "kebab-case")]
 pub enum ImageAnimation {
     #[default]
-    Animate,
```
englut: Is the option to animate on display undesirable?

data/src/config/preview.rs
```diff
@@ -306,6 +306,15 @@ impl Card {
     }
 }
 
+#[derive(Debug, Clone, Deserialize, Default)]
+#[serde(rename_all = "kebab-case")]
+pub enum ImageAnimation {
+    #[default]
+    AnimateOnHover,
```
englut, line 313: Makes sense to me to have `AnimateOnHover` as default 🫡

But was the option to animate on display (`Animate`) undesirable?

casperstorm (approved):

englut (commented):
Wow. Thanks for the awesome refactor @casperstorm! This is a much cleaner animation approach than using the `iced_gif` lib 🙏 

I am seeing that avatars aren't animating anymore. Not sure if that was on purpose, but just wanted to call that out. I don't want to step on any toes, so I will wait before making anymore changes. 🫡

casperstorm:
> I am seeing that avatars aren't animating anymore. Not sure if that was on purpose, but just wanted to call that out. I don't want to step on any toes, so I will wait before making anymore changes. 🫡

I think we keep the focus on just preview for now, and then extend for avatars, and so on later. Will be easy.

src/buffer/context_menu.rs
casperstorm: Hehe i just corrected when i saw, didnt check others. Will do that now. Thanks!

casperstorm:
> > I am seeing that avatars aren't animating anymore. Not sure if that was on purpose, but just wanted to call that out. I don't want to step on any toes, so I will wait before making anymore changes. 🫡
> 
> I think we keep the focus on just preview for now, and then extend for avatars, and so on later. Will be easy.

@englut tbh, if you are up feel free to just add it on top of this branch and we can add it shortly after!

data/src/config/preview.rs
casperstorm: I removed it because it was spiking resources. We can add it later if we want.

englut: Ahhh. That is a good reason to hold off on adding it now. Thanks @casperstorm :pray:

PR #2288, opened 2026-08-15
add configurable nickname color overrides

### LINKED ISSUE #2283: Selectively color nicknames based on rules
I like everyone to have the same color nick but my close friends to use different nick colors. Some might prefer to set specific users to use specific nick colors with also the current randomizing behavior.

It would be nice if nick colors for users could be set manually in the config. Additionally, REGEX rules could maybe be useful to use too, so that nicknames that follow a specific format get caught (eg. in the Indieweb channels, Slack users have `[]` around their nickname, which would be useful to catch).

casperstorm:
Fixes #2283.


andymandias (approved):
Very nice as always, really like the configuration scheme!

PR #2293, opened 2026-08-16
Add `triggers_unread` Setting for Server Messages

andymandias:
Drop server message categories (action/passive), and make category properties (whether dimmed by default and whether triggers unread) explicit user-configurable settings.  Keeps mode change server messages dimmed value as `false`, but reverts their triggers_unread status to `false`.


casperstorm:
Doesn’t the unread state depend on whether history is loaded? If we change `triggers_unread`, open buffers use the new setting, but what happens to closed buffers?

casperstorm:
Should we mention we remove  `passive_dimmed` and `actions_dimmed` in changelog?

andymandias:
> Doesn’t the unread state depend on whether history is loaded? If we change `triggers_unread`, open buffers use the new setting, but what happens to closed buffers?

Currently we don't go back and retroactively apply changes to triggers unread on `History::Partial` (e.g. if a user sends a message in an unloaded buffer it's marked as unread, then the user is set to be ignored → the buffer remains marked as unread).  I think this is something we should change in general, but might be better tackled after history restructuring than before.

casperstorm (approved):

PR #2294, opened 2026-08-17
Fix parsing of RELAYMSG nicknames

### LINKED ISSUE #2151: Some (RELAYMSG'd) lines are sometimes shown like NOTICEs even when they're not
Alright so, quick rundown:
- some users' messages that are bridged from another network using (a fork of) MatterBridge show up as if they're `NOTICE` messages, i.e without a nick and just a • instead.
- You can observe this behaviour in #queerjunction on either `ircs://transirc.chat` or `ircs://starlightnet.work` when certain users talk
- This is not replicable on `senpai`, `gamja` and `goguma`
- The user in question has a RELAYMSG'd nick of `Lilith-[MaskedWitch]-(she/they)/QSENC`

and a screenshot for good measure:

Image

senpai on the left, Halloy version `2026.7.1` on the right

## steps to replicate:
- join `#queerjunction` on `ircs://starlightnet.work`
- observe issue (with lines from CHATHISTORY or in real-time)

casperstorm:
Fixes #2151. @cyrneko can you confirm this works?


irc/proto/src/parse.rs
```diff
@@ -153,22 +151,13 @@ fn user(input: &str) -> IResult<&str, User> {
...
-        |s: &str| s.contains(':') && s.contains('.'),
-    );
-    let nickname = alt((expanded_nick, strict_nick));
+    // A `!` separates the nickname from the user. Bridge nicknames may use
+    // characters outside the regular nickname set.
+    let prefixed_nick = recognize(terminated(
+        many1_count(none_of("\0\r\n !")),
+        peek(char('!')),
+    ));
+    let nickname = alt((prefixed_nick, strict_nick));
```
cyrneko: is the assumption here that `!` will always be used for RELAYMSG? if so, I'm not 100% sure if that's the case because my default ergo configuration uses `/` for example, e.g `Username/XMPP`.

casperstorm: `/` and `!` have different roles here. `/` is part of the relayed nickname (Username/XMPP), while `!` separates the nickname from the user in `Username/XMPP!user@host`. So `/` is supported. I should update the comment a bit i think

cyrneko: oh! yea no I was only haphazardly skimming through while waiting for Terra to do its thing (it didn't end up doing its thing) so I mis-interpreted what it said

whoops

cyrneko:
image

Seems to still be happening. Not reproducible with `senpai` for comparison

casperstorm:
Alrite, ill dig into it. Thanks for testing @cyrneko.

casperstorm:
Should be fixed now after last commit @cyrneko!

cyrneko:
aight, lemme see for myself :)

cyrneko:
okay, very interesting, I don't see it in the history of `ircs://irc.transirc.chat`, though I do see it on the much older messages from my bouncer??

I must say I'm a little confused. I'm not sure whether we can reproduce a scenario where someone's nick is exactly `Ashtrid-In-Paperboat-<3/QSENC` which is one of the failure cases I still see; That said the failure case you were testing in #queerjunction seems to indeed be fixed (the one with parantheses and pronouns in the nick) :D

data/src/message.rs
```diff
@@ -383,6 +387,7 @@ pub struct Message {
     pub hash: Hash,
     pub hidden_urls: HashSet<Url>,
     pub is_echo: bool, // Only relevant if direction == Direction::Received
+    pub relayed: bool,
```
andymandias: Not sure if it'd add too much to this PR, but I think it could be nice for this to be `relayed_by: Option<Nick>`, and then the relaying bot can be surfaced in the UI in a tooltip, context menu, or elsewhere.

andymandias (approved):
Haven't tested directly with relaymsg, but it looks good to me.  One idea for a possible enhancement, but that could be a follow-up.

PR #2296, opened 2026-08-17
Fix WebSocket connection shutdown

casperstorm:
Gracefull shutdown of websockets.


irc/src/connection/websocket.rs
```diff
@@ -98,11 +100,22 @@ impl<Codec> WebSocketConnection<Codec> {
...
+    pub async fn shutdown(&mut self) -> Result<(), Error> {
+        if self.close_received {
+            self.stream.flush().await?;
+        } else {
+            self.stream
+                .close(Some(CloseFrame {
+                    code: CloseCode::Normal,
+                    reason: Utf8Bytes::default(),
+                }))
+                .await?;
```
xcfrg: don't we need to wait for the peer’s Close response?

Issue #2302, opened 2026-08-19
Decouple `+draft/react` and `+draft/unreact` capabilities


englut:
I have run across a scenario where the server implemented an older version of the draft spec for reacts which did not have `+draft/unreact`. Because of this though, other clients can react on this server, but Halloy can not.

We just received a report in #halloy by @edwardloveall, that UnrealIRCd has a reaction [plugin](https://github.com/unrealircd/unrealircd-contrib/blob/unreal6/files/react.c) that similarly implements `+draft/react` but not `+draft/unreact`.

If it is acceptable to decouple `+draft/react` and `+draft/unreact` from allowing Halloy to add reactions, I can work on this 🙏 

The one question is: do we gate unreacting behind requiring `+draft/unreact`?

PR #2304, opened 2026-08-19
Decouple `+draft/unreact` from `+draft/react`

englut:
Addresses #2302.

Some servers implement an older version of `+draft/react` where there was no `+draft/unreact` yet. These implementations allow for react & unreact. Decoupling the two in our codebase allows reacting to work for these servers.

Found two so far:

[reaction plugin for UnrealIRCd](https://github.com/unrealircd/unrealircd-contrib/blob/unreal6/files/react.c)
[matrix2051](https://github.com/progval/matrix2051/issues/92)


edwardloveall:
FWIW I just built and tested this and it indeed works on a server that supports `+draft/react` but not `+draft/unreact`! With the caveat also that I cannot "take back" a reaction, as expected.

casperstorm (approved):
Great addition friend.

PR #2309, opened 2026-08-21
Update for rebased iced

### LINKED ISSUE #2280: Cursor gets stuck in caret state and does not switch into pointer
**Test Case:**
- open Channel Discovery for liberachat server
- do not scroll the list
- hover over channel text description - cursor switches into caret/text cursor
- hover over channel name or link in channel description - cursor switches into pointer
- scroll the list with mouse scroll
- hover over channel name or link in channel description - cursor stays in caret/text state

halloy 2026.8
macOS 15.7.9

https://github.com/user-attachments/assets/3809dc59-d723-40fc-8b3d-f3d1c5a6e13f

### LINKED ISSUE #2308: Config editor doesn't "out-scroll" cursor
Scrolling requires the _final_ cursor position to still be within the content otherwise it gets blocked.

e.g. on first load the cursor is at 0:0 and scrolling the content up doesn't work, since the content would now start at line 1+, but cursor is still at line 0

text_editor example in iced-rs works fine

luca020400:
Most of the work is around the new Editor generic iface
Drop our custom combo_box implementation and use iced


luca020400:
This still needs some love around middle click handling on the iced part, ~~and a check as to why we had a custom combo_box is likely needed (update on that front, seems like it was originally added for arrow+tab navigation of the entries)~~

~~We lose focus handling in combo_box (via upstream)~~

luca020400:
Ok focus handling for command bar is added back, since combo_box lost it via its state handling

~~, what's left is middle click improvements~~

andymandias (approved):

englut:
Tested with flatpak and seems to work okay. But only under KDE and wayland. Haven't tried with Sway or Gnome yet.

casperstorm (approved):

casperstorm:
Rebase and we are good @luca020400

PR #2310, opened 2026-08-22
Automatically Save Before Reloading Config

andymandias:
Automatically save configuration in editor to disk (if editor is open, with unsaved changes) before reloading configuration.  To make for a more intuitive experience when editing the configuration and using "Reload config file" from the sidebar menu, the reload configuration shortcut, or reloading the configuration via the command bar.

Also moves the existing logic for limiting to one reload in-flight from the sidebar into the dashboard, and applies it to all configuration reloads initiated via the dashboard or its children.


casperstorm (approved):
Nice!

PR #2317, opened 2026-08-24
Change websocket ping interval behavior

luca020400:
In the default `Burst` mode `::Interval` replays every missed tick
back-to-back as soon as it's next polled.

If we miss the window to send a Ping (e.g. the process is suspended),
we'd fire a burst of Message::Ping on resume, or fail to send them.

`Delay` instead resets the window if it notices we missed an interval
and starts as normal with a new cycle.


andymandias (approved):
Thanks!

PR #2318, opened 2026-08-25
Emoji aliases

luca020400:
Adds support for emoji aliases via completion support.

Configuration is done via TOML
```toml
[buffer.emojis.aliases]
aliases = {
    ":)" = ":slightly_smiling_face:",
    ":(" = ":frowning:",
    ":D" = "😃",
    ":3" = "😺",
}
```

It supports alias to shortcode (+synthetic) and alias to emoji entries, by converting the latter to a known shortcode (or synthetic if unknown)


docs/configuration/buffer.md
```diff
@@ -566,6 +570,24 @@ E.g. `:D` will not show the emoji picker unless `characters_to_trigger_picker` i
...
+Shortcodes follow the [GitHub Emoji](https://emojipedia.org/github) standard. For example, `:slightly_smiling_face:` is the shortcode for 🙂.
+
+```toml
+# Type: map
+# Values: map with string key/value pairs
+# Default: {}
+
+[buffer.emojis.aliases]
+aliases = {
+    ":)" = ":slightly_smiling_face:",
```
4e554c4c, line 584: i don't see why this shouldn't just be the emoji, so that people don't need to look up the shortcode

luca020400: This seemed the easiest solution to avoid implementing a generic "search and replace" completion kind, although that can be done regardless.

I'm partial to using shortcodes because I like them, but happy to change it if wanted.

BKVad1m:
so will it include a separated configs that prevent lot of 10k alias lines for nerd font classes?

PR #2322, opened 2026-08-27
Show tooltip with emoji shortcode

luca020400:
This is rather horrible, but we can get away by chunking
the plain text whenever we find an emoji.

It works™

### DIFF
--- a/src/emoji.rs
+++ b/src/emoji.rs
@@ -45,7 +45,7 @@ fn synthetic_shortcode(emoji: &emojis::Emoji) -> String {
     emoji.name().to_lowercase().replace(' ', "_")
 }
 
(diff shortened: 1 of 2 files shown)


luca020400:
Also on top of that, I've noticed that the emoji crate doesn't strip Unicode Variant Selectors (and rightfully so) so some emojis entered from what I assume OS provided IME aren't considered emojis and rendered _as is_

This would also be broken for the old "only emoji" font size since it wouldn't properly recognize them, now it's more noticeable because of the missing tooltip.

Note: It's broken for Fragments of content, and it can't be easily extended there as it expects all fragments to be a `span`, and tooltip clearly isn't.

PR #2329, opened 2026-08-29
Harden DCC filename sanitization against reserved names and illegal characters

gigioneggiando:
`sanitize_filename` already strips path components and control characters, but a
peer-supplied DCC filename can still resolve to a Windows reserved device name
(`CON`, `NUL`, `LPT1`, …) or contain characters that are illegal in a filename on
Windows (`< > : " | ? *`). Trailing dots/spaces are silently stripped by Windows
and could smuggle a reserved name past an extension check.

This tightens `sanitize_filename`:
- replace the Windows-illegal characters (in addition to control characters);
- strip trailing dots and spaces before further checks;
- prefix a leading `_` when the base name matches a reserved DOS device name
  (case-insensitive, with or without an extension);
- keep the existing fallback to `dcc_transfer` for empty results.

The reserved-name and illegal-character handling is applied on all platforms so a
transferred file behaves the same regardless of the receiving OS. New unit tests
cover reserved names, trailing dot/space, and illegal characters; the existing
traversal/empty/control-char tests are unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,8 @@ Fixed:
 - Scroll-to-reply when clicking a reply preview
 - Show registration-required join errors in the channel buffer instead of the server buffer
 - Include/exclude conditions are properly accounted for when highlighting nicknames
+- Sanitize incoming DCC filenames against Windows reserved device names, illegal
+  characters, and trailing dots/spaces
 
 Changed:
 
--- a/data/src/file_transfer.rs
+++ b/data/src/file_transfer.rs
@@ -12,6 +12,16 @@ pub mod task;
 
 const FALLBACK_FILENAME: &str = "dcc_transfer";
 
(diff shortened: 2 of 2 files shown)


luca020400:
tbh I'd rather use https://crates.io/crates/sanitize-filename instead of hand rolling sanification by ourselves, beforehand I checked the linux path and it was sane but forgot windows is a thing.

Up to maintainers

gigioneggiando:
Makes sense — `sanitize-filename` covers the reserved-name and illegal-char cases and lets us drop the hand-rolled logic. Happy to switch this PR over to it if you're fine taking the dependency; otherwise I've left the current approach in place (and fixed the formatting that was failing CI). Your call.

andymandias:
Closing in favor of #2366, which includes this commit.  Thank you for bringing this to our attention!

PR #2330, opened 2026-08-29
Guard server-icon decoding against oversized (decompression-bomb) images

gigioneggiando:
Server icons are fetched from a URL the server advertises. `fetch` already caps
the encoded download at 5 MiB (`MAX_ICON_SIZE`), but a small compressed payload
can still decode to enormous pixel dimensions, so the byte cap does not bound the
memory needed to decode/upload the image.

The link-preview pipeline already guards against this (`preview.rs`) by rejecting
images whose padded dimensions exceed the maximum GPU buffer size. This mirrors
that check for server icons: after the body is read and the format detected (and
skipping SVG, which has no raster dimensions), read the image dimensions from the
in-memory bytes and reject the icon with `LoadError::TooLarge` if the padded size
would exceed `wgpu::Limits::downlevel_defaults().max_buffer_size`. The check runs
before the blob is written to the on-disk cache.

### DIFF
--- a/data/src/server_icon.rs
+++ b/data/src/server_icon.rs
@@ -3,7 +3,6 @@ use std::io;
 use std::sync::Arc;
 
 use iced::Task;
-use iced_wgpu::wgpu;
 use reqwest::header;
 use sha2::{Digest, Sha256};
 use tokio::fs;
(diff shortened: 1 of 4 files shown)


data/src/server_icon.rs
```diff
@@ -256,6 +257,34 @@ async fn fetch(
         bytes.extend_from_slice(&chunk);
     }
 
+    // Guard against decompression bombs: a small compressed payload can still
```
luca020400, line 259: Since this is the same code as preview.rs, maybe create a "common" util to check bytes of images?

gigioneggiando:
Good call — pulled the padded-size/buffer-limit math into `image::gpu_buffer_overflow`, now shared by both the link-preview and server-icon guards instead of duplicated.

classabbyamp:
I would recommend reading the halloy [contributing policy](https://halloy.chat/contributing#ai) before continuing to paste llm output in comments

gigioneggiando:
I just saw it and I get it, sorry (I thought looking and verifying the output was enough, but it's not). Now everything is gonna be my own text, if you want I can re-write this PR!

casperstorm (approved):

PR #2331, opened 2026-08-29
Bound the number of concurrently open IRCv3 batches from a server

gigioneggiando:
A `BATCH +<ref>` message opens a batch that lives in `self.batches` until the
matching `BATCH -<ref>` closes it. Nothing bounds how many batches a server may
open without ever closing them, so a malicious or buggy server can grow this map
without limit from a single connection.

This caps concurrently open batches at `MAX_OPEN_BATCHES` (1024 — far above any
legitimate use, including nested multiline/chathistory batches). Beyond the cap
the batch open is dropped with a warning; messages tagged with the dropped
reference are simply handled without batch context, so no existing batch state is
disturbed. The wire is otherwise unchanged.

This addresses the clearest server-driven unbounded-collection vector; other
server-populated maps are more naturally bounded by the user's own subscriptions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,8 @@ Fixed:
 - Scroll-to-reply when clicking a reply preview
 - Show registration-required join errors in the channel buffer instead of the server buffer
 - Include/exclude conditions are properly accounted for when highlighting nicknames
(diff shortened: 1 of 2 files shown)


casperstorm (approved):
LGTM Thanks.

PR #2332, opened 2026-08-29
Redact credentials from the IRC protocol log

gigioneggiando:
When IRC protocol logging is enabled, the codec writes every sent and received
line verbatim to `irc_protocol_logs/<server>/<date>.log`. That includes lines
carrying secrets in cleartext: `PASS`, `OPER`, the SASL `AUTHENTICATE` credential
payload, and NickServ-style `PRIVMSG NickServ :IDENTIFY …`. A user who shares a
protocol log to report a bug can leak their server/NickServ password.

This adds a `redact_log_line` pass applied to both sent and received lines before
they are written to the log (the bytes actually sent on the wire are untouched):
- `PASS` — the credential is replaced with `<redacted>`;
- `OPER` — the name is kept, the password redacted;
- `AUTHENTICATE` — the payload is redacted unless it is a SASL control token
  (`+`/`*`) or an uppercase mechanism name (`PLAIN`, `EXTERNAL`, `SCRAM-…`), so
  the base64 credential blob is never written;
- `PRIVMSG` to a nick (never a channel) whose body begins with a services auth
  keyword (`IDENTIFY`, `REGISTER`, `SETPASS`, `GHOST`, `RELEASE`) — the keyword is
  kept and the credential that follows is redacted.

Message tags and source prefixes are skipped when locating the command. New unit
tests cover each case and confirm non-sensitive lines pass through unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,8 @@ Fixed:
 - Scroll-to-reply when clicking a reply preview
 - Show registration-required join errors in the channel buffer instead of the server buffer
 - Include/exclude conditions are properly accounted for when highlighting nicknames
+- Redact credentials (`PASS`, `OPER`, SASL `AUTHENTICATE`, NickServ `IDENTIFY`)
+  from the on-disk IRC protocol log
 
 Changed:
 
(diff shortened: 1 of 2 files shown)


casperstorm:
The logs which we have in Halloy are meant for debugging purposes so i don't see any reason to redact from them. Thank you for the time.

Issue #2129, opened 2026-06-23
CPU usage and UI lock


Kompootor:
On W11H 25H2 26200.8655 , Halloy newly installed via winget yesterday:

I've been letting it idle overnight on 2 servers and 3 channels with very little chat activity, just to monitor somebody's script. I find just now that the UI is semi-locked -- the window is active and the process appears live on Windows, but I cannot enter text in chat, and switching channels does not cause any response in the UI.  Also, the process is using nearly all CPU resources of one core (memory appears steady, 50 MB, and no disk or GPU).

Left-clicking on channels in the left pane of the UI switches the text in the top title bar, but the main panels do not change at all. There is no cursor. Right-clicking does nothing, and the hamburger menu is not accessible. 

I'd really like to keep uptime in this channel if possible, so I'd really not like to have to restart the program. If there's any recommendations to try sooner rather than later that will not interrupt my current channel connections, I'd appreciate any help.

casperstorm:
If you go to the user menu down to the left, and press about Halloy what GPU backend do you see?

Kompootor:
Hamburger menu doesn't respond. If you're looking for version info I thought the winget current date was enough but I can grab it from winget quick.

Halloy 2026.7.2 squidowl.halloy

I tried switching the process to efficiency mode to maybe lower CPU usage. The memory usage went to 2 GB and CPU remained the same, fwiw.

Kompootor:
Also let me provide procmon:

<img width="3840" height="2400" alt="Image" src="https://github.com/user-attachments/assets/f9368704-3f8e-405f-af25-0b73d66e1387" />

Kompootor:
My Nvidia panel is showing halloy as the only application it is running on the machine. (I'm not sure why, if there's GPU acceleration anywhere, it should be on a text client like halloy, and to the exclusion of my web browser or my desktop environment. I know Nvidia software has problems, but this seems like a weird problem to have separate from the Windows GUI.) I tried a few things, like disabling the Nvidia GPU in device manager, which got halloy off of said list and removed its GPU memory use, but it did not remove its CPU use or unlock its GUI. 

It's weird that the GUI appears to respond to keyboard commands to some extent, like I can tell it opens new panels like when I type ctrl+L for example and click on a space where the panel should be. But the window itself does not change. I don't know if the developers have a command that resets their GUI or for whatever tiling library they are using or something. (I still don't get why this should even be a thing that causes a problem in windows.)

[...]

PR #2354, opened 2026-09-08
iced, again

luca020400:
This time the Text renderer isn't tied
to the Font anymore.

Alongside that line_height is now a global
setting, thus we can avoid passing it around
if we can, and make use of the default.


luca020400:
The main difference from an UI point of view is that config editor now follows app theme
(before;after)
image

And we gain the ability to runtime configure line_height

luca020400:
Actually the last commit is probably wrong, since https://github.com/iced-rs/iced/pull/3454 just added a "default", but doesn't allow yet to change it at runtime...
Hopefully it'll in the future

luca020400:
I'm not exactly sure what fixed it, but now the metadata menu doesn't automatically select the text within it for me?

What I see ~main
[Screencast_20260912_161653.webm](https://github.com/user-attachments/assets/9eafa5b1-e27b-42fe-a44d-49feef2a81c2)

casperstorm (approved):

luca020400:
Last second fixup!

luca020400:
Last rebase may help with #2129

luca020400:
~~Sadly this seems to break message kb navigation in the weirdest way, the menu isn't shown yet all kb events are captured by it (eg can copy message)
Will investigate~~

I missed some changes related to viewport, now everything has been updated

andymandias:
I'm getting hidden bits in the context menu in my testing, such as:

Screenshot From 2026-09-20 19-24-31
Screenshot From 2026-09-20 19-27-34

But, sometimes context menus will display perfectly fine.  It seems like it might be that the bounds of the visible part of the context menu corresponds to which parts of the context menu have the source widget below it?  E.g. in the pictured context menu for `ls` the context menu items are only visible for the part of each that is over the nicklist (from which the context menu originates).

andymandias (approved):
Thank you!

PR #2356, opened 2026-09-12
Update ubuntu installation

edwardloveall:
By default, org.squidowl.halloy.desktop did not work for me because it was  looking for the `halloy` executable located in `$HOME/.local/bin/halloy` but  `$HOME/.local/bin` was not in the `$PATH`.

This changes the default `Exec` key to point to `.local/bin/halloy`. I chose  this path because if you downlod the linux distrobution `.tar.gz` and extract  it to `~/.local`, everthing (the binary, icons, .desktop file, etc) will end up  in the right place.

```sh
tar -xf halloy-xxxxxx-xxxx-linux.tar.gz -C ~/.local
```


assets/linux/org.squidowl.halloy.desktop
```diff
@@ -4,7 +4,7 @@ Comment=IRC client written in Rust
 Type=Application
 Keywords=IRC;IM;Chat;
 Categories=Network;IRCClient;
-Exec=halloy %U
+Exec=.local/bin/halloy %U
```
englut, line 7: Installs can be in any path. e.g. `/usr`, `/usr/local`, `/opt`, etc. This line will break every other install of halloy on linux.

What you want to do is to revert this line and update your `$PATH` to include `~/.local/bin` or install halloy to `/usr/local/`.

englut (changes requested):

assets/linux/org.squidowl.halloy.desktop
edwardloveall: Hmmm yes I see your point. I was coming at this from the angle of "here's how to install halloy without modifying your system". That felt better to me than "modify your system's path so that the install will work", even thought it's not a halloy-specific modification. What am I missing?

Regardless, I can change the instructions to something like: modify your `.profile` (or whatever is correct; I'll test) to add `~/.local/bin` to your `PATH`.

andymandias: I'm a little surprised that Ubuntu doesn't include .local/bin by default (it's in the [XDG base directory spec](https://specifications.freedesktop.org/basedir/latest/), not something we came up with).  But if it isn't then it isn't.  My suggestion is we open up the installation recommendation for `installing via release build` to be a directory of the user's choice that is in their `$PATH`.  Then we could suggest `~/.local` for users who are unsure where to install (or, if there's a more commonly supported path we can suggest that instead).

edwardloveall: Thanks @andymandias. I'm currently trying to figure out where that $PATH gets defined. It's not (necessarily) the user's shell. After some experimentation, mine is a very strange...

```
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/snap/bin
```

(two `/snap/bin` ???)

If I could figure out what's setting this, I'd feel better about encouraging people to put it anywhere they want and then showing them how to check that their PATH points to that location. As it is, it's hard to give guidance, which is why I came up with the suggestion of `~/.local`. I can keep (slowly) trying to figure that out, but I'm also happy to take other suggestions, or your original one.

Or maybe I'm being too guiding, here. Let me know!

andymandias: I think the instinct/motivation here is good, we would like this to be as bulletproof as possible without getting too complicated.  From a brief look at the [Exec key](https://specifications.freedesktop.org/desktop-entry/latest/exec-variables.html) it appears what PATH is used is left a little ambiguous as "$PATH environment variable used by the desktop environment" (or, at least, I'm not aware of a standard for what this is).  So it might get a bit tricky to provide guidance there, but I would be happy to be proven wrong.  If it looks like `/usr/local/` would be a more portable option I think that would be a good option (though we have to deal with permissions in that case);  `~/.local/` was chosen as the default for `install-linux.sh` somewhat arbitrarily, so we don't have a strong reason to keep it if it'll cause problems for users.

If you don't mind, could you elaborate on how the `.desktop` failed for you?  `gtk-launch` wouldn't find it?

edwardloveall: If I follow my `tar ... -C ~/.local` strategy above, Halloy does not show up in the list of apps for Ubuntu. It does after I add ~/.local/bin/` to the `Exec` path.

If I use `gtk-launch`, it does work, but it has no icon, even after I run `gtk-update-icon-cache`

andymandias: Thank you for the extra details.  Unfortunately I haven't been able to repro in Ubuntu 26.04 GNOME 50 for whatever reason, so it makes it harder for me to .  When convenient would you be able to test whether extracting to `/usr/local` (such that the relevant files go into `/usr/local/bin/`, `/usr/local/share/applications/`, etc) works on your system (without using an absolute path in the `.desktop`)?

edwardloveall: > Unfortunately I haven't been able to repro in Ubuntu 26.04 GNOME 50 for whatever reason, so it makes it harder

Oof. That's no fun. It makes me think maybe I broke my setup somehow. Could you create this `test.desktop` file and `ln -s` it to `~/.local/share/applications/test.desktop`?

```config
[Desktop Entry]
Type=Application
Name=Test
Exec=env
StartupWMClass=com.edwardloveall.test
```

You should then see it in the list of apps to run (super key then type `Test`). If you run it, it will will print the full environment to the system log which you can follow with `sudo journalctl -f`. Maybe if we compare what we have that could help?

<details><summary>Output</summary>

```sh
HOME=/home/edward
LANG=en_US.UTF-8
LOGNAME=edward
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/snap/bin
SHELL=/usr/bin/zsh
USER=edward
XDG_DATA_DIRS=/home/edward/.local/share/flatpak/exports/share:/var/lib/flatpak/exports/share:/usr/local/share/:/usr/share/:/var/lib/snapd/desktop
XDG_RUNTIME_DIR=/run/user/1000
GTK_MODULES=gail:atk-bridge
QT_ACCESSIBILITY=1
DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1000/bus
DESKTOP_SESSION=ubuntu
GDMSESSION=ubuntu
GNOME_DESKTOP_SESSION_ID=this-is-deprecated
GPG_AGENT_INFO=/run/user/1000/gnupg/S.gpg-agent:0:1
OLDPWD=/home/edward
PWD=/home/edward
QT_IM_MODULE=ibus
QT_IM_MODULES=wayland;ibus
SHLVL=0
SSH_AUTH_SOCK=/run/user/1000/gcr/ssh
USERNAME=edward
XDG_CURRENT_DESKTOP=ubuntu:GNOME
XDG_MENU_PREFIX=gnome-
XDG_SESSION_CLASS=user
XDG_SESSION_DESKTOP=ubuntu
XDG_SESSION_EXTRA_DEVICE_ACCESS=render:accel
XDG_SESSION_TYPE=wayland
XMODIFIERS=@im=ibus
_=/usr/bin/gnome-session
MANAGERPID=4936
MANAGERPIDFDID=20709
INVOCATION_ID=877941e795144ecab9354fa07af9d607
JOURNAL_STREAM=10:35400
SYSTEMD_EXEC_PID=5401
MEMORY_PRESSURE_WATCH=/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/session.slice/org.gnome.Shell@ubuntu.service/memory.pressure
MEMORY_PRESSURE_WRITE=c29tZSAyMDAwMDAgMjAwMDAwMAA=
GJS_DEBUG_OUTPUT=stderr
GJS_DEBUG_TOPICS=JS ERROR;JS LOG
GNOME_SETUP_DISPLAY=unix:/tmp/.X11-unix/X1
DISPLAY=:0
XAUTHORITY=/run/user/1000/.mutter-Xwaylandauth.JM0XV3
WAYLAND_DISPLAY=wayland-0
GIO_LAUNCHED_DESKTOP_FILE=/home/edward/.local/share/applications/test.desktop
GIO_LAUNCHED_DESKTOP_FILE_PID=91588
```

</details> 

I will also test extracting files to `/usr/local`.

andymandias: It appears to have my entire `PATH`, both the parts in my shell config and the parts inherited (which has a double `/snap/bin` like yours).

```
HOME=/home/andymandias
LANG=en_US.UTF-8
LOGNAME=andymandias
PATH=/home/andymandias/.local/share/pnpm:/usr/local/texlive/2026/bin/x86_64-linux:/home/andymandias/.bun/bin:/opt/sublime_text:/opt/sublime_merge:/opt/google/chrome:/home/andymandias/.local/bin:/home/andymandias/.local/share/pnpm:/usr/local/texlive/2026/bin/x86_64-linux:/home/andymandias/.bun/bin:/opt/sublime_text:/opt/sublime_merge:/opt/google/chrome:/home/andymandias/.local/bin:/usr/local/texlive/2026/bin/x86_64-linux:/home/andymandias/.bun/bin:/opt/sublime_text:/opt/sublime_merge:/opt/google/chrome:/home/andymandias/.local/bin:/home/andymandias/.cargo/bin:/home/andymandias/.atuin/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/snap/bin
SHELL=/usr/bin/fish
USER=andymandias
XDG_DATA_DIRS=/home/andymandias/.local/share/flatpak/exports/share:/var/lib/flatpak/exports/share:/usr/local/share/:/usr/share/:/var/lib/snapd/desktop
XDG_RUNTIME_DIR=/run/user/1000
GTK_MODULES=gail:atk-bridge
QT_ACCESSIBILITY=1
ACCENT_COLOR=blue
DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1000/bus
DESKTOP_SESSION=ubuntu
EDITOR=vi
GDMSESSION=ubuntu
GNOME_DESKTOP_SESSION_ID=this-is-deprecated
GPG_AGENT_INFO=/run/user/1000/gnupg/S.gpg-agent:0:1
PAGER=less --use-color
PNPM_HOME=/home/andymandias/.local/share/pnpm
PWD=/home/andymandias
QT_IM_MODULE=ibus
QT_IM_MODULES=wayland;ibus
SSH_AUTH_SOCK=/run/user/1000/gcr/ssh
TMUX_ACCENT_COLOR=blue
USERNAME=andymandias
XDG_CURRENT_DESKTOP=ubuntu:GNOME
XDG_MENU_PREFIX=gnome-
XDG_SESSION_CLASS=user
XDG_SESSION_DESKTOP=ubuntu
XDG_SESSION_EXTRA_DEVICE_ACCESS=render:accel
XDG_SESSION_TYPE=wayland
XMODIFIERS=@im=ibus
trash=/home/andymandias/.local/share/Trash/files/
MANAGERPID=7688
MANAGERPIDFDID=45236
INVOCATION_ID=204c413b293a4894aee3e1d94fa0e505
JOURNAL_STREAM=10:9605748
SYSTEMD_EXEC_PID=2390226
MEMORY_PRESSURE_WATCH=/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/session.slice/org.gnome.Shell@ubuntu.service/memory.pressure
MEMORY_PRESSURE_WRITE=c29tZSAyMDAwMDAgMjAwMDAwMAA=
GJS_DEBUG_OUTPUT=stderr
GJS_DEBUG_TOPICS=JS ERROR;JS LOG
GNOME_SETUP_DISPLAY=unix:/tmp/.X11-unix/X1
DISPLAY=:0
XAUTHORITY=/run/user/1000/.mutter-Xwaylandauth.Y6FDV3
WAYLAND_DISPLAY=wayland-0
GIO_LAUNCHED_DESKTOP_FILE=/home/andymandias/.local/share/applications/test.desktop
GIO_LAUNCHED_DESKTOP_FILE_PID=636288
```

edwardloveall: Fascinating. Unclear what's going on there. To confirm, you launched the `test` app from the app launcher, not using gtk-launch, right?

In any case, I think your [earlier comment](https://github.com/squidowl/halloy/pull/2356#discussion_r4049837284) about suggesting `/usr/local` is a good one. I'll update the PR.

andymandias: Yes.  From the overview search or the application grid that appears from "Show Apps" button in the dock, both produce the same (Ulauncher too, if that matters).  Very mysterious!

PR #2360, opened 2026-09-15
Exclude mode changes, kicks and topic changes from marking unread

neilalexander:
None of these actions are really attention-warranting, so let's not cause them to show the unread marker next to a room. Otherwise it can be extremely annoying that the bouncer reconnecting in the background can suddenly mark half the world as unread again.


neilalexander:
Was pointed out to me that #2293 may cover this instead, please close if so.

PR #2363, opened 2026-09-15
Highlight matches on nicknames of users in the channel

### LINKED ISSUE #2362: `highlights.match` words don't match nicknames of users in the channel
**Version:** 2026.8 (macOS). Also present on `main` at c3db026.

**Config:**

```toml
[[highlights.match]]
words = ["alice"]
case_insensitive = true
```

**Steps:**

1. Join a channel where user `alice` is present.
2. Have another user send `alice: can you look at this?`
3. Have `alice` leave the channel, then send the same message again.

**Expected:** both messages are highlighted and appear in the Highlights window.

**Actual:** only the second message is highlighted. While `alice` is in the channel, the match never fires.

Reproduced on Libera with two test users in an empty channel. The Highlights window missed the message sent while the matched nick was present, and caught the one sent after it left.

**Cause:** in `parse_fragments_with_highlights` (`data/src/message.rs`), text is split into fragments first, so channel members' nicknames become `Fragment::User`. The `highlights.matches` loop then only applies each regex to `Fragment::Text`, so user fragments are never checked. `regex` rules are affected the same way.

https://github.com/squidowl/halloy/blob/c3db026/data/src/message.rs#L2035

**Use case:** tracking mentions of teammates or bot names, e.g. relay bots that prefix each post with the author's nick.

<details>
<summary>Failing unit test (add to <code>data/src/message.rs</code> tests)</summary>

```rust
#[test]
fn highlight_match_words_on_channel_user_nickname() {
    use std::collections::HashMap;

    let server = Server {
        name: "Test Server".into(),
        network: None,
    };

    let isupport = HashMap::<isupport::Kind, isupport::Parameter>::new();
    let chantypes = isupport::get_chantypes_or_default(&isupport);
    let statusmsg = isupport::get_statusmsg_or_default(&isupport);
    let casemapping = isupport::get_casemapping_or_default(&isupport);

    let highlights = Highlights {
        nickname: Nickname {
            exclude: None,
            include: None,
            case_insensitive: true,
        },
        matches: vec![
            toml::from_str(
                "words = [\"alice\"]\ncase_insensitive = true",
            )
            .unwrap(),
        ],
    };

    let target =
        target::Target::parse("#test", chantypes, statusmsg, casemapping);
    let our_nick = Nick::from_str("steve", casemapping);
    let sender = User::from(Nick::from_str("bob", casemapping));
    let text = "alice: can you look at this?".to_string();

    let highlight_kind = |nicks: &[&str]| {
        let channel_users = nicks
            .iter()
            .map(|nick| User::from(Nick::from_str(nick, casemapping)))
            .collect::<ChannelUsers>();

        parse_fragments_with_highlights(
            text.clone(),
            Some(&sender),
            Some(&channel_users),
            &target,
            &our_nick,
            &highlights,
            &server,
            casemapping,
        )
        .1
    };

    assert!(
        highlight_kind(&["bob", "steve"]).is_some(),
        "alice not in channel"
    );
    assert!(
        highlight_kind(&["alice", "bob", "steve"]).is_some(),
        "alice in channel"
    );
}
```

`cargo test -p data --lib highlight_match_words` at c3db026: the first assert passes, the second fails with `alice in channel`.

</details>

lancewillett:
Fixes #2362

`highlights.match` rules only ran on text fragments. Nicknames of channel members are split into user fragments first, so a rule matching a present user's nickname never fired.

Each rule now also runs on each run of adjacent text and user fragments. Any user fragment a match overlaps becomes a `HighlightMatch`. The regex sees the same text it would if that user were absent, and `triggers_highlight` picks the message up.

Own-nick highlighting and fragment order are unchanged.

**Testing**

- New unit test `highlight_matches_on_channel_user_nicknames`. It fails on `main` and covers:
  - the same fragments whether the user is present or absent
  - an anchored regex not matching the nick on its own
  - own-nick precedence
- `cargo test -p data`, `cargo clippy -p data --all-targets -- -D warnings`, and `cargo +nightly fmt --check` pass.
- Live on Libera with two test users in an empty channel. Message sent while the matched nick was present: missing from Highlights on 2026.8, present with this branch.

**Known limits**

- A matched nick renders as a highlight, not a clickable nick. This matches what happens when that user is absent.
- A match spanning text and a nick (`hello alice`) highlights only the nick part. The message still counts as a highlight.
- If one rule matches part of a nick (`ali`), the whole nick becomes a highlight, so a later rule can't match the rest (`ce`).

**AI use**
🤖 Generated with [Claude Code](https://claude.com/claude-code)


andymandias:
Please see our [contributing guide](https://halloy.chat/contributing#ai).

lancewillett:
@andymandias Oof. Rough first welcome.

lancewillett:
I'm a new Halloy user and passionate about open source, but won't plan on contributing code fixes.

PR #2364, opened 2026-09-16
Prepend bouncer name to server name in irc logs if needed

luca020400:


data/src/stream.rs
```diff
@@ -749,8 +749,13 @@ impl LogWriter {
...
 
-        let log_dir =
-            data_dir.join("irc_protocol_logs").join(server.to_string());
+        let path = if let Some(network) = &server.network {
+            format!("{}/{}", server.name, network.name)
+        } else {
+            server.to_string()
+        };
+
+        let log_dir = data_dir.join("irc_protocol_logs").join(path);
```
andymandias, line 759: For our Windows users' sake, I think we want something like:

```suggestion
        let mut log_dir = data_dir.join("irc_protocol_logs");

        if let Some(network) = &server.network {
            log_dir = log_dir.join(server.name.as_ref()).join(&network.name);
        } else {
            log_dir = log_dir.join(server.name.as_ref());
        };
```

andymandias (approved):

PR #2366, opened 2026-09-16
Filename sanification

luca020400:
- **Harden DCC filename sanitization against reserved names and illegal characters**
- **Switch filename sanification to sanitize-filename**
- **Also sanitize irc protol logs**


casperstorm (approved):
LGTM

andymandias (approved):

PR #2367, opened 2026-09-16
Switch to Swatinem/rust-cache

luca020400:
It appears to be a bit smarter when it comes
to cache re-use, specifically even if Cargo.lock
changes it is able to pull old cache, build newly
changed deps, and push newly created cache.


luca020400:
I have no idea if that works, I can't run CI on my fork for reasons.

luca020400:
The main difference is that this seems to be smarter when it comes to cache re-use, eg with #2361 it was able to only compile the new rustls, while current rules forced to rebuild 2 targets.

With that said, it works fine ^^

andymandias (approved):

PR #2372, opened 2026-09-19
Add Setting to Control Anti-Flood Burst Parameter

### LINKED ISSUE #2371: "Excess Flood" from LiberaChat when connecting to already-running dircproxy session
New Halloy user, just testing it out.  When I connect it to my session running in dircproxy, Halloy sends enough messages to the server to get me booted.

Dircproxy keeps a session open even when an IRC client isn't connected to it.  When a client connects, it replays recent history for each channel, and for the server in general.  Halloy responds to this by trying to get information about each channel... but it seems to send the requests too quickly.

Anonymized, trimmed log attached.

[halloy-excess-flood.log](https://github.com/user-attachments/files/32406783/halloy-excess-flood.log)

It looks like Halloy responds to each JOIN with a WHO and a MODE... but if a bunch of joins happen all at once, it doesn't rate-limit the outgoing requests.

The joins, of course, are coming from dircproxy.  But the who/mode commands are passed to liberachat, which has no idea that the client has freshly connected.  It just sees a sudden surge of unprompted requests from a previously-established connection.

Halloy version: 2026.8 from this repo's release page.

andymandias:
Allow user control of anti-flood burst parameter (the number of messages that can be sent before strict rate-limiting kicks in), and reduce default to 6 from 10 (edit: reduced further based on feedback in #2371).

Also increases the token priority of the initial WHO poll sent when joining a channel and no-implicit-names is acknowledged.


luca020400:
~~at the moment who polls for newly joined channels are sent immediately if NoImplicitNames is ack, shouldn't we also "delay" those a bit? or the idea is to have the anti flood mechanism handle that at the end of the message queue?~~ we don't

casperstorm (approved):

